PR #3, opened 2021-03-03
Database setup changes

pboers1988:
Modified that database setup in such a way that the `db.db` object is initiliased after calling `init_database` in the OrchestratorCore class.

### DIFF
--- a/orchestrator/__init__.py
+++ b/orchestrator/__init__.py
@@ -13,7 +13,7 @@
 
 """This is the orchestrator workflow engine."""
 
-__version__ = "0.0.2-rc1"
+__version__ = "0.0.2rc2"
 
 from orchestrator.app import OrchestratorCore
 from orchestrator.settings import app_settings, oauth2_settings
--- a/test/unit_tests/conftest.py
+++ b/test/unit_tests/conftest.py
@@ -1,188 +1,30 @@
 import os
-import uuid
 from contextlib import closing
-from typing import Any, Callable, Dict, List, Optional, Tuple, cast
+from typing import Optional, cast
 
 import pytest
 import structlog
 from alembic import command
 from alembic.config import Config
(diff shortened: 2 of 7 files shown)


orchestrator/__init__.py
```diff
@@ -13,7 +13,7 @@
 
 """This is the orchestrator workflow engine."""
 
-__version__ = "0.0.2-rc1"
+__version__ = "0.0.2-rc2"
```
hmvp, line 16: When installing flit complains that it should be 0.0.2rc2

orchestrator/db/database.py
```diff
@@ -220,6 +220,10 @@ def database_scope(self, **kwargs: Any) -> Generator["Database", None, None]:
         self.scoped_session.remove()
         self.request_context.reset(token)
 
+    def update(self, *args: Any, **kwargs: Any) -> None:
```
hmvp, line 223: I dont think this should be here. Even if its a noop

test/unit_tests/conftest.py
```diff
@@ -235,8 +77,24 @@ def db_uri(worker_id):
...
+        db_uri: The database uri
+
+    Yields:
+        Nothing
+
+    """
+    db.update(Database(db_uri))
+    yield
+
+
```
hmvp, line 95: Why not move this into the database fixture itself?

Anyway it does not need autouse since its a dep from an autouse fixture

orchestrator/db/database.py
pboers1988: I added it to remove the type: ignore

hmvp: I prefer the type ignore over this.. You could also cast the db thing where you call update to Wrapped again

hmvp (approved):

PR #4, opened 2021-03-04
Database interaction commands

pboers1988:
This PR exposes cli commands to wrap alembic commands and add it to the core cli commands.


orchestrator/app.py
```diff
@@ -95,6 +99,7 @@ def _index() -> str:
 
     def instrument_app(self) -> None:
         logger.info("Activating Opentelemetry tracing to app", app=self.title)
+        self.base_settings.TRACING_ENABLED = True
```
hmvp, line 102: This feels wrong...

orchestrator/app.py
```diff
@@ -120,3 +125,33 @@ def add_sentry(
             integrations=[SqlalchemyIntegration(), RedisIntegration()],
         )
         self.add_middleware(SentryAsgiMiddleware)
+
+    @staticmethod
+    def register_subscription_model(product_to_subscription_model_mapping: Dict[str, Type[SubscriptionModel]]) -> None:
```
hmvp, line 130: `register_subscription_models`? Since you add multiple

orchestrator/cli/database.py
```diff
@@ -0,0 +1,70 @@
...
+)
+
+
+@app.command(name="init")
+def init() -> None:
+    """
+    Run the migrations.
+
+    This command will run the migrations for initialization of the database. If you have extra migrations that need to be run,
+    add this to the
```
hmvp, line 29: Finish sentence?

orchestrator/cli/database.py
```diff
@@ -0,0 +1,70 @@
...
+    logger.info("Creating directory", directory=os.path.abspath(versions))
+    os.makedirs(versions)
+    versions_schema = os.path.join(migration_dir, "versions/schema")
+    logger.info("Creating directory", directory=os.path.abspath(versions_schema))
+    os.makedirs(versions_schema)
+
+    source_env_py = os.path.join(orchestrator_module_location, f"{migration_dir}/templates/env.py.j2")
+    env_py = os.path.join(migration_dir, "env.py")
+    logger.info("Creating file", file=os.path.abspath(env_py))
+    copyfile(source_env_py, env_py)
```
hmvp, line 51: Why do we need to copy files? Is that safe?

orchestrator/app.py
pboers1988: It is wrong, good catch, this was some debug stuff.. Will change.

orchestrator/cli/database.py
hmvp: Why cant we do the same thing as we do now? With two folders? except one resides in the python lib?

thijs-creemers (approved):

orchestrator/__init__.py
```diff
@@ -13,7 +13,7 @@
 
 """This is the orchestrator workflow engine."""
 
-__version__ = "0.0.2rc2"
+__version__ = "0.0.2rc11"
```
acidjunk, line 16: we might start using a scheme here ;)

orchestrator/config/__init__.py
```diff
@@ -37,5 +37,6 @@
 NODE_SUBSCRIPTION_ID = "node_subscription_id"
 PORT_MODE = "port_mode"
 INTERNETPINNEN_PREFIX_SUBSCRIPTION_ID = "internetpinnen_prefix_subscription_id"
+PEER_GROUP_SUBSCRIPTION_ID = "peer_group_subscription_id"
```
acidjunk, line 40: Do we need these constants in a Generic release?

orchestrator/services/subscriptions.py
```diff
@@ -29,6 +29,7 @@
     IP_PREFIX_SUBSCRIPTION_ID,
     NODE_SUBSCRIPTION_ID,
     PARENT_IP_PREFIX_SUBSCRIPTION_ID,
+    PEER_GROUP_SUBSCRIPTION_ID,
```
acidjunk, line 32: We seem to leak some SURF specific constants. Maybe good to add least add a todo (in the code) to split/clean it?

PR #6, opened 2021-03-31
Improve error handling

hmvp:
This allows us to add more information to process failures by using specific Exceptions:

ProcessFailure: for normal process failures or programmer errors in validation tasks
InconsistentData: For validation task errors when data is not in an expected state (AssertionError also works for this)

### DIFF
--- a/orchestrator/utils/errors.py
+++ b/orchestrator/utils/errors.py
@@ -14,11 +14,11 @@
 
 import sys
 from http import HTTPStatus
-from typing import TYPE_CHECKING, Dict, Optional, Tuple, cast
+from typing import TYPE_CHECKING, Dict, Optional, cast
 
 import structlog
 
-from orchestrator.types import ErrorDict, ErrorState
+from orchestrator.types import JSON, ErrorDict, ErrorState
 
 # TODO Fix nwastdlib
 if TYPE_CHECKING:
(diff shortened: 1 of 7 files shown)


orchestrator/utils/errors.py
```diff
@@ -90,57 +104,29 @@ def is_api_exception(ex: Exception) -> bool:
...
-
-    if isinstance(err, FormValidationError):
+    if isinstance(err, ProcessFailure):
+        return {"class": type(err).__name__, "error": err.message, "traceback": show_ex(err), "details": err.details}
+    elif is_api_exception(err):
+        err = cast(ApiException, err)
         return {
             "class": type(err).__name__,
-            "error": str(err),
+            "error": err.reason,
```
pboers1988, line 122: It seems like the API is changing? We don't need a schema change?

pboers1988 (approved):

orchestrator/utils/errors.py
hmvp: No why would we? Btw the diff on this is not great, make sure to set "hide whitespace changes" to on

pboers1988: Ah correct ;)

hmvp: The FormValidationError and FormNotCompleteError are not handled by this function (since fastapi) and we only use this function with actual exceptions instead of anything goes (Since removal of sn7) so I just simplified it a lot and added support for ProcessFailure errors

PR #7, opened 2021-04-01
Remove surf specific code.

pboers1988:
Removed SURF Specific code from the orchestrator-core repository.


orchestrator/db/models.py
```diff
@@ -478,21 +476,6 @@ class SubscriptionTable(BaseModel):
         deferred=True,
     )
 
-    port_mode = column_property(
```
hmvp, line 481: The two above should also be moved. They are a backwards compat thing and should not be used

orchestrator/forms/validators.py
```diff
@@ -293,29 +273,6 @@ def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
         field_schema.update(format="label", type="string")
 
 
-class MigrationSummary(DisplayOnlyFieldType):
```
hmvp, line 296: Isnt this generic? There is a frontend component for it as well

orchestrator/forms/validators.py
```diff
@@ -293,29 +273,6 @@ def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
...
-    def __modify_schema__(cls, field_schema: Dict[str, Any]) -> None:
-        field_schema.update(format="summary", type="string", uniforms={"data": cls.data})
-
-
-def migration_summary(data: Optional[SummaryData] = None) -> Type[MigrationSummary]:
-    namespace = {"data": data}
-    return new_class("MigrationSummaryValue", (MigrationSummary,), {}, lambda ns: ns.update(namespace))
-
-
-class MTU(ConstrainedInt):
```
hmvp, line 309: I left the mtu and vlan stuff in since we seem to be network centric.. Or should we make a seperate lib with Network based extensions? This seems usefull for others..

orchestrator/forms/validators.py
pboers1988: Moved to `network_type_validators`

PR #10, opened 2021-04-09
fix: Product Blocks not being linked to Products with create()

howderek:
When using the `create()` helper, Products should be linked to Product Blocks when one puts a list of names for the create function (see function docstring). This was not happening correctly because I was iterating through the wrong list when trying to resolve the names to UUIDs (old: `product_blocks`, now: `products`). This PR fixes that problem.


pboers1988 (approved):
Looks good, however build is failing! I'll accept once that is resolved.

howderek:
Apologies for the delay. This ended up causing issues for us due to UUIDs being autogenerated for some rows. Once we fix this, I will update this PR.

PR #20, opened 2021-06-09
Docs

pboers1988:


docs/getting-started/development.md
```diff
@@ -0,0 +1,33 @@
...
+Install the project and its dependancies to develop on the code.
+
+### Step 1 - install flit:
+```console
+pip install flit
+```
+
+### Step 2 - install the development code:
+```console
+flit install --deps develop --symlink
```
acidjunk, line 15: This is almost never what you want. It installs all packages outside the active venv.

a `--python path/to/venv/bin/python` is needed

hmvp: @acidjunk Did you install flit in your user env or in the specific venv? I did not encounter this issue. I got it in the specific venv

pboers1988: Ill add that to the docs, that flit needs to be installed in a env not globally.

acidjunk: I installed it in an env: but it installed the packages outside the env (conda base python)

The whole point of setting up a symlinked dev version is to have 2 editable projects (IMHO)

acidjunk: @hmvp encountered the same problem, still not sure when it happens exactly. The readme has tested instructions now.

pboers1988: Added a warning in the documentation to explain what happens here.

PR #74, opened 2021-12-22
Update websocket all processes

tjeerddie:
- improve process data in all processes websocket endpoint.
- remove logic to send failed processes as the first websocket message, initial data will come from http request.
- rename create_process_step_websocket_data to create_process_websocket_data.
- rename send_process_data_to_websocket to create_process_websocket_data.
- move logic done within process show endpoint to the show_process function, so it will also be used for the all processes websocket.
- change the create_process_websocket_data to only return process data, the step data is already in the process data.
- remove unnecessary process detail websocket endpoint.
- remove engine settings from channels in send_process_data_to_websocket function, it does not need process data.
- Load new process stat in processes service before creating websocket data.


pboers1988 (approved):
LGTM

PR #87, opened 2022-01-29
Backup subscription details

acidjunk:


pboers1988 (approved):

acidjunk:
> @acidjunk If you would like this to be tested let me know and I can make a release candidate.
> 
> Otherwise LGTM

@pboers1988 : should be fine now!

orchestrator/workflows/steps.py
```diff
@@ -10,30 +10,90 @@
...
+    subscription = subscriptions.get_subscription(subscription_id)
+    subscription_backup = {subscription_id: to_serializable(subscription)}
+    Do your changes here ...
+    return {..., "subscription": to_serializable(subscription), "__old_subscriptions__": subscription_backup }
+    ```
+
+    """
+    try:
+        subscription = SubscriptionModel.from_subscription(subscription_id)
+    except:  # noqa
```
Mark90, line 73: This can at least be `except Exception:`, right?

pboers1988: As this is is at the core a Pydantic model we should get at least an instance of the Basic Pydantic exception? all other exceptions may fail, hard. IT should catch the ` PydanticTypeError` or the ` PydanticValueError`

pboers1988 (changes requested):

orchestrator/workflows/steps.py
pboers1988: but probably a ValidationError wil do.

PR #89, opened 2022-02-01
Remove missing_product_blocks check in diff_product_in_database

tjeerddie:
- Remove node_subscription_id resource type
- Add find_by_tag in ProductBlockTable model


pboers1988 (approved):
LGTM

PR #95, opened 2022-02-11
Add scheduled build

Mark90:
Add a workflow that builds and tests the project every 2 weeks and sends the outcome to Slack.


pboers1988 (approved):
LGTM

PR #101, opened 2022-02-15
feat: make list of workflows usable-while-out-of-sync extensible

nemimccarter:
Allows consumers of the package to extend the list of workflows that can be used while a subscription is out of sync using the following syntax:
```
from orchestrator.services.subscriptions import (
    WF_USABLE_WHILE_OUT_OF_SYNC,
)


WF_USABLE_WHILE_OUT_OF_SYNC.extend(
    [
        "{workflow_name}",
    ]
)
```


pboers1988 (approved):
LGTM

PR #102, opened 2022-02-15
make project editable installable without a setup.py

florisie:
Just an attempt to make project installable without a setup.py.

With this configuration we should be able to add a reference to a commit on Github or to a local directory (in case we would make it a submodule of the orchestrator project) in our requirement.txt file(s).

This will make it easier to test specific branches/commits of the orchestrator-core project when running in Docker containers.

BTW requires a very recent version of pip so please update...

You can test it by replacing:
orchestrator-core==0.3.4

with
-e git+https://github.com/workfloworchestrator/orchestrator-core.git@14f17fc8d80fed3f882ee5e6128e9ff203a0ea9d#egg=orchestrator

in base.txt of the orchestrator project.

Finally note that a version and a description are mandatory in the pyproject.toml file. Maybe we should update these before merging (if we want this PR to be merged at all)


acidjunk (approved):

pyproject.toml
```diff
@@ -1,10 +1,10 @@
...
-requires = ["flit"]
-build-backend = "flit.buildapi"
-
-[tool.flit.metadata]
+[project]
 module = "orchestrator"
+name = "orchestrator"
 dist-name = "orchestrator-core"
+# Version is mandatory
+version = "0.3.4"
```
Mark90, line 6: `pyproject.toml` should be added to https://github.com/workfloworchestrator/orchestrator-core/blob/main/.bumpversion.cfg to keep this version in sync with the other files

pboers1988:
@florisie I take it you confirmed that the project is not `-e` installable with the current setup.py ?

pyproject.toml
```diff
@@ -1,10 +1,7 @@
...
-requires = ["flit"]
-build-backend = "flit.buildapi"
-
-[tool.flit.metadata]
-module = "orchestrator"
-dist-name = "orchestrator-core"
+[project]
+name = "orchestrator-core"
+version = "0.3.5"
+description = "Open source orchestration softwre for NREN's"
```
pboers1988, line 4: Typo: "Opensource orchestration software"

florisie:
> @florisie I take it you confirmed that the project is not `-e` installable with the current setup.py ?

BTW, I just tried it again. But with a "RUN pip install --upgrade pip" in the Dockerfile before installing the requirements. In that case the installation completes wothout any errors...

pyproject.toml
florisie: Ah yes, good spot. I corrected "software" but when I google "opensource" then I see it's mostly written with a space in between. So I left the space there for now.

pyproject.toml
```diff
@@ -1,10 +1,7 @@
-[build-system]
-requires = ["flit"]
-build-backend = "flit.buildapi"
-
-[tool.flit.metadata]
-module = "orchestrator"
-dist-name = "orchestrator-core"
+[project]
+name = "orchestrator-core"
+version = "0.3.5"
```
Mark90, line 3: Somehow this version still says 0.3.5 🤔

pboers1988:
Closing due to staleness. Can be reopend when necessary.

PR #104, opened 2022-02-16
Change create product blocks migration helper

tjeerddie:
- add option to add child relations.


pboers1988 (commented):
Why do we need this?

PR #109, opened 2022-02-22
Fix self.parents error by checking if it exists in self

tjeerddie:


thijs-creemers:
Issue is solved in https://github.com/workfloworchestrator/orchestrator-core/pull/111, close this PR.

PR #116, opened 2022-03-03
Change other subscriptions endpoint with filtering statuses

tjeerddie:


orchestrator/api/api_v1/endpoints/subscriptions.py
```diff
@@ -166,16 +166,20 @@ def subscription_workflows_by_id(subscription_id: UUID) -> Dict[str, List[Dict[s
 
 
 @router.get("/instance/other_subscriptions/{subscription_instance_id}", response_model=List[UUID])
-def subscription_instance_parents(subscription_instance_id: UUID) -> List[UUID]:
-    subscription_instance = SubscriptionInstanceTable.query.get(subscription_instance_id)
+def subscription_instance_parents(subscription_instance_id: UUID, filter_statuses: Optional[str] = None) -> List[UUID]:
```
Mark90, line 169: Does this work? `filter_statuses: Optional[SubscriptionLifecycle] = None`

tjeerddie: then you can only filter by a single status and not multiple

Mark90: Oh, you want to send multiple statuses in the string so you can filter by multiple statuses. That should also be possible, but not sure if this is usable in the frontend https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#query-parameter-list-multiple-values

Mark90 (approved):
👍

PR #135, opened 2022-05-02
1388: Change subscription_details_by_id_with_domain_model to support lists

tjeerddie:
supporting lists is necessary for one to many relations

- change `getattr_in` to support lists.
- change `product_block_paths` to support lists.
- filter out all `in_use_by` ids which point to product block within it's own domain model instance (or subscription).

### DIFF
--- a/orchestrator/api/api_v1/endpoints/subscriptions.py
+++ b/orchestrator/api/api_v1/endpoints/subscriptions.py
@@ -83,14 +83,19 @@ def subscription_details_by_id_with_domain_model(subscription_id: UUID) -> Dict[
 
     subs_obj = SubscriptionModel.from_subscription(subscription_id)
     subscription = subs_obj.dict()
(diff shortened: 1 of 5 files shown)


pboers1988 (approved):
LGTM, can you add a RC for this?

orchestrator/api/api_v1/endpoints/subscriptions.py
```diff
@@ -83,14 +83,19 @@ def subscription_details_by_id_with_domain_model(subscription_id: UUID) -> Dict[
 
     subs_obj = SubscriptionModel.from_subscription(subscription_id)
     subscription = subs_obj.dict()
+    sub_instance_ids = [subs_obj.subscription_id]
```
Mark90, line 86: I wonder why the old code checked if in_use_by_id was equal to subscription_id. The relation is between SubscriptionInstances and not Subscriptions, so I think the subscription_id can be omitted (in a future change)

PR #136, opened 2022-05-02
Fix created_at field always null in subscriptions/domain-model endpoint

tjeerddie:


orchestrator/domain/base.py
```diff
@@ -1180,6 +1184,8 @@ def _to_product_model(cls: Type[S], product: ProductTable) -> ProductModel:
             product_type=product.product_type,
             tag=product.tag,
             status=product.status,
+            created_at=product.created_at if product.created_at else None,
+            end_date=product.end_date if product.end_date else None,
```
mrijk, line 1188: is

end_date = product.end_date 

not sufficient?

pboers1988 (approved):
LGTM

acidjunk (commented):
LGTM

PR #142, opened 2022-05-17
Fix optional list error when class or value is none

tjeerddie:
Changed so the `ValueError` that shows `'NoneType' object has no attribute 'name'` when you have a list that does not have `None` to a `ValidationError`.

- add tests for direct relations in `api/test_subsctiptions`

### DIFF
--- a/orchestrator/domain/base.py
+++ b/orchestrator/domain/base.py
@@ -387,13 +387,15 @@ class DomainModel(BaseModel):
                         possible_product_block_types[item.name]._from_other_lifecycle(item, status, subscription_id)
                     )
             else:
-                if is_optional_type(field_type):
+                if is_optional_type(field_type) and not is_union_type(field_type):
                     data[field_name] = None
                     field_type = first(get_args(field_type))
                 elif is_union_type(field_type):
+                    if is_optional_type(field_type):
+                        data[field_name] = None
                     field_types = get_args(field_type)
                     for f_type in field_types:
-                        if f_type.name == value.name:
+                        if value and not isinstance(None, f_type) and f_type.name == value.name:
                             field_type = f_type
                 if value:
                     data[field_name] = field_type._from_other_lifecycle(value, status, subscription_id)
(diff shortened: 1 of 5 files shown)


orchestrator/domain/base.py
```diff
@@ -387,13 +387,15 @@ def _data_from_lifecycle(cls, other: "DomainModel", status: SubscriptionLifecycl
                         possible_product_block_types[item.name]._from_other_lifecycle(item, status, subscription_id)
                     )
             else:
-                if is_optional_type(field_type):
+                if is_optional_type(field_type) and not is_union_type(field_type):
```
tjeerddie, line 390: This might be unnecessary

pboers1988: No Looks good I think,

orchestrator/domain/base.py
```diff
@@ -387,13 +387,15 @@ def _data_from_lifecycle(cls, other: "DomainModel", status: SubscriptionLifecycl
...
                     data[field_name] = None
                     field_type = first(get_args(field_type))
                 elif is_union_type(field_type):
+                    if is_optional_type(field_type):
+                        data[field_name] = None
                     field_types = get_args(field_type)
                     for f_type in field_types:
-                        if f_type.name == value.name:
+                        if value and not isinstance(None, f_type) and f_type.name == value.name:
```
Mark90, line 398: `not isinstance(None, f_type)` can None be subclassed?

tjeerddie: `f_type` can be a `NoneType`, don't know a better way to validate that

Mark90: And `if value` is not ideal for falsy values

Mark90: > `f_type` can be a `NoneType` class, don't know a better way to validate that

Ah, I see that `types.NoneType` was added in 3.10. Yeah, makes sense then :)

Mark90 (approved):

PR #144, opened 2022-05-31
Added filter options to in use by and depends on queries, made the pa…

pboers1988:
…rameter injectable through a fastapi dependency

### DIFF
--- a/orchestrator/services/subscriptions.py
+++ b/orchestrator/services/subscriptions.py
@@ -36,7 +36,7 @@ from orchestrator.db import (
 )
 from orchestrator.db.models import SubscriptionInstanceRelationTable
 from orchestrator.targets import Target
-from orchestrator.types import UUIDstr
+from orchestrator.types import SubscriptionLifecycle, UUIDstr
 from orchestrator.utils.datetime import nowtz
 
 logger = structlog.get_logger(__name__)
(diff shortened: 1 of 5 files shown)


orchestrator/services/subscriptions.py
```diff
@@ -409,12 +418,21 @@ def query_depends_on_subscriptions(subscription_id: UUID) -> Query:
         .with_entities(SubscriptionTable.subscription_id)
     )
 
-    return SubscriptionTable.query.filter(
-        or_(
-            SubscriptionTable.subscription_id.in_(resource_type_relations.scalar_subquery()),
-            SubscriptionTable.subscription_id.in_(relation_relations.scalar_subquery()),
+    if filter_statuses:
```
tjeerddie, line 421: adding all filter_statuses to the variable would probably look a bit better then two queries almost exactly the same with an if statement.

orchestrator/services/subscriptions.py
```diff
@@ -368,20 +368,29 @@ def query_in_use_by_subscriptions(subscription_id: UUID) -> Query:
...
+        return SubscriptionTable.query.filter(
```
tjeerddie, line 372: same here

tjeerddie (approved):

PR #151, opened 2022-06-24
Make it possible to push all domain-models to cache

pboers1988:
Proof of concept to see if it is possible to push all domain models to a cache, to increase read performance. This code must be executed after every workflow.

### DIFF
--- a/orchestrator/api/api_v1/endpoints/subscriptions.py
+++ b/orchestrator/api/api_v1/endpoints/subscriptions.py
@@ -12,7 +12,6 @@
 # limitations under the License.
 
 """Module that implements subscription related API endpoints."""
-
 from http import HTTPStatus
 from typing import Any, Dict, List, Optional, Union
 from uuid import UUID
@@ -25,6 +24,7 @@ from oauth2_lib.fastapi import OIDCUserModel
 from sqlalchemy import select
 from sqlalchemy.orm import contains_eager, defer, joinedload
 from starlette.exceptions import HTTPException
+from starlette.requests import Request
 from starlette.responses import Response
 
 from orchestrator.api.error_handling import raise_status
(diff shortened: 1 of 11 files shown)


orchestrator/api/api_v1/endpoints/subscriptions.py
```diff
@@ -103,41 +105,34 @@ def subscriptions_all() -> List[SubscriptionTable]:
...
+        response.headers["ETag"] = _generate_etag(domain_model)
+        return domain_model
+
     try:
-        subs_obj = SubscriptionModel.from_subscription(subscription_id)
+        subscription_model = SubscriptionModel.from_subscription(subscription_id)
+        extended_model = build_extendend_domain_model(subscription_model)
+        etag = _generate_etag(extended_model)
+
+        if etag == if_none_match:
```
mrijk, line 125: You could probably reshuffle this somewhat (use an else for the non-cached response) and have the etag setting outside the if/else to avoid some code duplication

orchestrator/utils/redis.py
```diff
@@ -0,0 +1,50 @@
...
+cache = Redis(host=app_settings.CACHE_HOST, port=app_settings.CACHE_PORT)
+
+
+def to_redis(subscription: Dict[str, Any]) -> None:
+    if getenv("AIOCACHE_DISABLE", 0) == 0 and app_settings.CACHE_DOMAIN_MODELS:
+        logger.info("Setting cache for subscription.", subscription=subscription["subscription_id"])
+        etag = _generate_etag(subscription)
+        cache.set(
+            f"domain:{subscription['subscription_id']}", dumps((subscription, etag)), ex=3600 * 24 * 7
+        )  # one week
```
mrijk, line 36: or define and use a constant called ONE_WEEK to avoid this comment :)

orchestrator/utils/redis.py
```diff
@@ -0,0 +1,50 @@
...
+    else:
+        logger.warning("Caching disabled, not caching subscription", subscription=subscription["subscription_id"])
+
+
+def from_redis(subscription_id: UUID) -> Optional[Tuple[Dict[str, Any], str]]:
+    if getenv("AIOCACHE_DISABLE", 0) == 0 and app_settings.CACHE_DOMAIN_MODELS:
```
mrijk, line 42: I would define a small helper function if_cache_enabled to avoid this duplication with to_redis

mrijk (approved):
LGTM, some minor comments

Issue #153, opened 2022-06-27
Choice with str label and dict value


gstewart86:
Hey gang,

It appears I'm unable to create a Choice with a `str` label and `dict` value. Given the following:
```
EsdbCircuitEnum = Choice("EsdbCircuitChoice", zip(esdb_circuit_map.values(), esdb_circuit_map.keys()))
```

I see the error:
```
ns
    EsdbCircuitEnum = Choice("EsdbCircuitChoice", zip(esdb_circuit_map.values(), esdb_circuit_map.keys()))
  File "/usr/local/lib/python3.9/enum.py", line 386, in __call__
    return cls._create_(
  File "/usr/local/lib/python3.9/enum.py", line 509, in _create_
    classdict[member_name] = member_value
  File "/usr/local/lib/python3.9/enum.py", line 97, in __setitem__
    if _is_private(self._cls_name, key):
  File "/usr/local/lib/python3.9/enum.py", line 49, in _is_private
    and name.startswith(pattern)
AttributeError: 'dict' object has no attribute 'startswith'
```

The pattern we've adopted in the meantime involves creating a map and translating the user choice like so:
```
    select_map: Dict = {}
    for edge in available_edges:
        select_map[edge.get('name')] = edge

    ServiceEdgeChoice = Choice("ServiceEdgeChoice", zip(select_map.keys(), select_map.keys()))

    class L2VPNForm(FormPage):
        class Config:
            title = product_name

        service_edge: ServiceEdgeChoice

    form_data = yield L2VPNForm
    user_input = form_data.dict()

    esdb_service_edge = get_resource_by_id(
        "service_edge", select_map[user_input["service_edge"]]["id"], params={"detail": 1}
    )
```

How does SURFnet's implementation typically handle these cases? Do you have any suggestions for making Choice capable of carrying a `dict`? It would be extremely nice for example to have a class like `APIChoice` that accepts parameters to generate that Choice from an external API endpoint, and return the related data automatically when chosen like so
```
    class L2VPNForm(FormPage):
        class Config:
            title = product_name

        service_edge: APIChoice('service_edge')
```

Mark90:
Hi Stewart,

[...]

PR #157, opened 2022-07-01
Add a new form field ChoiceValue: that lets the user submit an Enum v…

acidjunk:
…alue instead of label

### DIFF
--- a/orchestrator/forms/validators.py
+++ b/orchestrator/forms/validators.py
@@ -152,6 +152,8 @@ class Accept(str):
 
 
 class Choice(strEnum):
+    """Let the user choose from an enum and submit the label."""
+
     label: ClassVar[str]
 
     def __new__(cls, value: str, label: Optional[str] = None) -> "Choice":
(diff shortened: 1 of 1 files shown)


gstewart86:
found myself here as a result of a [related issue](https://github.com/workfloworchestrator/orchestrator-core/issues/153), but I just thought I'd note that `ChoiceValue` seemingly conflicts with existing validation mechanisms. Using that `ChoiceValue` code, I tried to define the following:
```
class NSOBackboneLinkIGPPreference(ChoiceValue):
    CORE = ("core", "test")
    WORST = "worst"
    TAIL = "tail"
    SEA_CABLE = "sea-cable"
    PREFERRED = "preferred"
```

and then use it in a Form like so
```
    class L2VPNForm(FormPage):
        class Config:
            title = product_name

        service_edge: ServiceEdgeChoice
        nso_test: NSOBackboneLinkIGPPreference
        validate_service_edge = validator("service_edge", allow_reuse=True)(validate_esdb_service_edge(select_map))
```

Selecting the first option, `CORE`, resulted in this error in the GUI:
image

I speculate that it's because the GUI is submitting the value "test", as defined in the value of the Enum, which is not actually part of the enum, therefore the error.

I just thought I'd add this context, in case it was unknown. I like this approach though otherwise 👍

pboers1988:
Seems to me that we should probably document what @acidjunk found about the Tuple's as that seems to cover all use-cases that were mentioned?

acidjunk:
Yup, added some extra documentation. WIP...

PR #158, opened 2022-07-01
Add ProcessDataBroadcastThread

Mark90:
Starting many workflows in bulk causes asyncio RuntimeErrors when 2 workflows try to broadcast a process step update at the same time. The root problem lies in the WebsocketManager being shared between threads, indirectly also sharing an `asyncio.Queue` object from `asyncio_redis`'s pubsub implementation. But this object is not thread-safe and causes different eventloops to clash when they access it at the same time.

A solution proposed in this PR is to have a dedicated thread for broadcasting process data.

* Adds a background thread `ProcessDataBroadcastThread` for publishing messages to redis. It contains a `threading.Queue` through which it receives process data messages 
* The thread is attached to the OrchestratorCore application which manages its lifecycle
* Processes related API endpoints submit process data to the thread though the shared `threading.Queue` object

### DIFF
--- a/docs/architecture/application/websockets.md
+++ b/docs/architecture/application/websockets.md
@@ -1,0 +1,44 @@
+# Websockets
+
+Orchestrator provides a websocket interface through which the frontend can receive real-time updates. This includes:
+* The process overview pages
+* The process detail page
+* Engine status
+
+
+## Implementation
+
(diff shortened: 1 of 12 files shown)


pboers1988 (approved):
Looks good would you mind adding some documentation?

pboers1988 (approved):

PR #163, opened 2022-08-01
Fix ENABLE_WEBSOCKETS not actually disabling websocket endpoints

tjeerddie:
- add logger for amount of websocket connections.


orchestrator/websocket/managers/broadcast_websocket_manager.py
```diff
@@ -56,8 +57,7 @@ async def disconnect(
 
     async def disconnect_all(self) -> None:
         for websocket in self.connected:
-            websocket.close(status.WS_1000_NORMAL_CLOSURE)
-            self.remove_ws_from_connected_list(websocket)
+            self.disconnect(websocket)
```
Mark90, line 60: `await self.disconnect`

Mark90 (approved):

PR #175, opened 2022-09-22
Send process data over ws on process delete

tjeerddie:
- this is done in order to update failed task banner in the frontend when tasks are deleted.

### DIFF
--- a/orchestrator/api/api_v1/endpoints/processes.py
+++ b/orchestrator/api/api_v1/endpoints/processes.py
@@ -67,7 +67,7 @@ from orchestrator.services.processes import (
 from orchestrator.settings import app_settings
 from orchestrator.types import JSON
 from orchestrator.utils.show_process import show_process
-from orchestrator.websocket import WS_CHANNELS, websocket_manager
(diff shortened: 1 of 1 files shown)


orchestrator/api/api_v1/endpoints/processes.py
```diff
@@ -76,14 +76,18 @@
...
-    if count > 0:
-        return None
-    else:
+def delete(request: Request, pid: UUID) -> None:
+    process = ProcessTable.query.filter_by(pid=pid).one_or_none()
+    if not process:
         raise_status(HTTPStatus.NOT_FOUND)
 
+    broadcast_func = api_broadcast_process_data(request)
+    websocket_data = {"process": {"id": process.pid, "status": ProcessStatus.ABORTED}}
```
codado-nl, line 85: Why aborted? :thinking:

codado-nl (approved):

orchestrator/api/api_v1/endpoints/processes.py
tjeerddie: In the frontend we remove processes that have `ABORTED` or `COMPLETED`. there also isn't a `DELETED` because the process doesn't exist anymore :3

PR #184, opened 2022-11-04
generate-domain-models: Only ask for default value for resources with instances

tjeerddie:
also fixes incorrect typing of subquery in generate-domain-models


pboers1988 (approved):
LGTM

PR #201, opened 2022-12-20
1552 fix extended domain model caching

Mark90:
We use `build_extendend_domain_model()` to populate the domain model cache. Currently this function also has a parameter `filter_owner_relations` to change the representation. The problem here is that this representation is cached and cannot be changed when retrieving from cache, i.e. when retrieving it through the API.

This PR solves this by splitting `build_extendend_domain_model()` into 2 functions, keeping the old function with a DeprecatedWarning.

1) `build_extended_domain_model()` (note the subtle name change) now only creates an extended subscription dict with all of the data
2) `format_extended_domain_model()` represents the subscription dict as instructed by the given filters, taking input from either `build_extended_domain_model()` or from cache

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.5.2-rc3
+current_version = 0.5.2-rc4
 commit = False
 tag = False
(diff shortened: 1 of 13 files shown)


pboers1988 (approved):
LGTM

PR #207, opened 2023-02-07
Docker build changes

Mark90:
* Add Dockerfile build-arg `VERSION` to guarantee the image is built with a specific orchestrator-core version
* Give docker images for normal core releases the tags: `latest` and `edge`
* Give docker images for core pre-releases the tag: `edge`

These changes are to facilitate automatic tests of the example-orchestrator-beginner app.


tjeerddie (approved):
LGTM

Issue #186, opened 2022-11-10
Add support to generate domain models for updating resource types in specific product blocks


tjeerddie:
The generate domain model can update/rename resource types over all product block, but we would also want to update resource types within a single product block.

### Tasks
- [x] change the `updates` argument to accept `block_resource_types` to update resource types per product block. 
       `updates = { 'block_resource_types': { 'block_name': { 'old_rt': 'new_rt' } } }`
- [X] ~~create a function to get the resources to update within product blocks with interactive shell. (could be split into a new issue).~~ moved to separate issue to build interactive shell
- [x] filter out the specific updates from the model diffs.
- [x] create a function to generate sql to update resource types within product blocks.

PR #208, opened 2023-02-07
Add option to migrate-domain-models to update Block Resource Types

tjeerddie:
- [x] change the `updates` argument to accept `block_resource_types` to update resource types per product block. 
       `updates = { 'block_resource_types': { 'block_name': { 'old_rt': 'new_rt' } } }`
- [x] filter out the specific updates from the model diffs.
- [x] create a function to generate sql to update resource types and within product blocks.
- add interactive way to select which old resource type should be updated and with which new resource type, per product block.

related: #186


pboers1988 (approved):
LGTM

Issue #180, opened 2022-10-24
Add support to generate migrations for multiple renamed fields


Mark90:
The migration generator implemented in https://github.com/workfloworchestrator/orchestrator-core/pull/179 only allows up to 1 resourcetype / fixed input to be renamed; if more fields need to be renamed then one would need to create additional migrations for each field renamed.

See conversation on https://github.com/workfloworchestrator/orchestrator-core/pull/179/files#r997822535

# Example
Suppose that the following productblock already exists in both the code and the database:

```
class MyBlock:
  field_a: int
  field_b: str
  field_c: bool
```

Now we rename `field_a` and `field_b` to respectively `field_x` and `field_y` and run the migration generator. 

The diff functions it uses will then return:

```
missing_resource_types_in_db = ['field_x', 'field_y']
missing_resource_types_in_model = ['field_a', 'field_b']
```

The rename algorithm should traverse these differences and allow the user to specify (preferably interactively by use of `input()`) actions like so:
```
rename_actions = {
  "field_a": "field_x",
  "field_b": "field_y",
}
```

The approach should be reusable for resource types, fixed inputs, product blocks

tjeerddie:
This is only an issue with interactive shell, you can update multiple with `updates` argument.

PR #214, opened 2023-02-09
Improve interactive resource type rename in migrate-domain-models

tjeerddie:
- changed map_rename_resource_types to first find all old resource types that aren't in any blocks and the possible resource types it can be renamed to, choices can't already be in the database.
- ^ after finding all possibilities, it will prompt user menu to choose resource type to rename and then prompt a menu with possible choices it can be renamed to.
- update unit tests.

Closes: #180


pboers1988 (approved):
LGTM

PR #215, opened 2023-02-13
Updated various dependancies

pboers1988:


pyproject.toml
```diff
@@ -56,28 +55,25 @@ requires = [
...
+    "uvicorn[standard]~=0.20.0",
+    "nwa-stdlib==1.4.4",
+    "oauth2-lib==1.2.3",
     "markupsafe==2.0.1",
     "bandit==1.7.2",
     "tabulate==0.9.0"
 ]
 description-file = "README.md"
-requires-python = ">=3.7,<3.12"
+requires-python = ">3.8,<3.12"
```
Mark90, line 76: This still includes 3.9

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.5.3
+current_version = 1.0.0-rc1
```
dmarjenburgh, line 2: I think this will be normalized to 1.0.0rc1. Maybe change the version identifier to match to avoid confusion

pyproject.toml
pboers1988: Spotted and fixed :)

Mark90 (approved):

orchestrator/__init__.py
```diff
@@ -13,7 +13,7 @@
 
 """This is the orchestrator workflow engine."""
 
-__version__ = "0.5.3"
+__version__ = "1.0.0-rc1"
```
dmarjenburgh, line 16: Here too

```suggestion
__version__ = "1.0.0rc1"
```

PR #217, opened 2023-02-20
rewrite pyproject.toml to latest standard

florisie:
Just for sharing/review, not final yet


pyproject.toml
```diff
@@ -1,10 +1,13 @@
+# TODO as a build system we should also be able to use either setuptools or poetry
```
Mark90, line 1: +1 for poetry

In other projects I've mainly used its competitor `pipenv` which is already much more capable than `flit`, and `poetry is even better. Nothing against `flit`, it's simple and works well, but just lacks dependency management.

Quick look at the pypi page will show the power of poetry https://pypi.org/project/poetry

PR #218, opened 2023-02-20
Available caches to flush endpoint

Mark90:
* Add redis util `delete_keys_matching_pattern` to delete keys using `"SCAN"`
* Add endpoint `get_cache_names` for gui to retrieve the possible cache flush options
* Dict `CACHE_FLUSH_OPTIONS` can be extended by orchestrator-implementer


pboers1988 (approved):
LGTM

PR #227, opened 2023-03-13
Add @serializabe_property

mrijk:


tjeerddie (changes requested):
Add tests

orchestrator/domain/base.py
```diff
@@ -63,6 +63,12 @@ class ProductNotInRegistryException(Exception):
     pass
 
 
+class serializabe_property(property):
```
dmarjenburgh, line 66:
```suggestion
class serializable_property(property):
```

orchestrator/domain/base.py
```diff
@@ -416,6 +422,47 @@ def _save_instances(
...
+    def dict(
+        self,
+        *,
+        include: Any | None = None,
+        exclude: Any | None = None,
+        by_alias: bool = False,
+        skip_defaults: bool | None = None,
+        exclude_unset: bool = False,
+        exclude_defaults: bool = False,
+        exclude_none: bool = False,
```
Mark90, line 441: Is it possible to use kwargs instead?

```python
def dict(self, *, **kwargs) -> dict[str, Any]:
    attribs = super().dict(**kwargs)
    
    if include := kwargs.get('include'):
        ...
```

mrijk: Probably. Was just following the suggestion from one of the comments in a Pydantic issue. I'm fine with both options.

tjeerddie (approved):
LGTM

orchestrator/domain/base.py
mrijk: Changed

PR #234, opened 2023-04-05
Revert "Let publish-package.yml trigger build-push-container.yml"

Mark90:
This reverts commit 006c98690e32ab0e391ba27e891d03f731bd80ab.

For some reason the triggered event has the main branch as `ref` instead of the tag name. And we need the tag name.

There's an explanation here https://github.com/orgs/community/discussions/27124 but the fix doesn't apply for us as we depend on external actions to read the `ref`. So, reverting.

### DIFF
--- a/.github/workflows/build-push-container.yml
+++ b/.github/workflows/build-push-container.yml
@@ -1,12 +1,8 @@
 name: Build and push container image
 
 on:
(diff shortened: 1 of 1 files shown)


hanstrompert (approved):
LGTM

PR #235, opened 2023-04-06
allow optional enum fields in populator

florisie:
Currently in the populator, when we see an enum field in the form, then we always select the first option. However, sometimes enum fields are optional and the list of options can be empty. In this case the populator fails.

This PR changes the logic such that if there are no options to select from then we simply leave the field empty.


Mark90 (approved):

tjeerddie (approved):
LGTM

PR #246, opened 2023-04-19
Fix #242: ensure a workflow cannot be resumed twice

### LINKED ISSUE #242: Make sure you cannot restart/resume workflows/tasks twice
### Summary

At this point in time it is possible to retry a workflow or task twice on the celery bus. Make sure it is not possible to do this by implementing a new ProcessStatus state RESUMED. Then a workflow/task will transition from CREATED to RUNNING to either DONE, SUSPENDED or FAILED, a SUSPENDED or FAILED workflow/task can be resumed or restarted and will transition to RESUMED.

### Solution

* [ ]  implement a new ProcessStatus RESUMED
* [ ]  add logic to prohibit a already restarted/resumed workflow to be restarted/resumed again

coleshaw:
OCD internal tracking issue: ORC-1443
GH issue: #242 

This PR adds a new `ProcessStatus` state of `RESUMED`, and checks for that value when a user accesses the API. If a process is in the `RESUMED` state, it cannot be re-resumed. This applies for both the individual process `/api/processes/{pid}/resume` endpoint as well as the `_async_resume_processes` call (linked to the `/api/processes/resume-all` endpoint).

This PR also sets the `RESUMED` state in the Celery methods, when a process is resumed in that code flow. I manually tested this -- hopefully I re-created the Celery environment correctly. I created a process for a workflow and then issued a `PUT` to the `/api/processes/{pid}/resume` endpoint. In the database, the `last_status` column value correctly changed from `created` -> `resumed` when it was put onto the Celery queue. I then tried to re-resume a process with `last_status == resumed`, and received the expected 409 error as a response. Processes that are in `failed` / `waiting` statuses also are set to `resumed` when issuing a `PUT` to `/api/processes/resume-all`. `resumed` processes are not attempted to be re-resumed by issuing a `PUT` to the `/api/processes/resume-all` endpoint.

Are there corresponding UI changes that are required, to show the user that a process might be in the `RESUMED` state, or to catch the 409 from trying to restart a resumed workflow?

Also, please provide feedback if I'm heading in the wrong direction, mis-understood the linked issue, or your team has other preferences with regards to code style / linting / etc.

### DIFF
--- a/orchestrator/api/api_v1/endpoints/processes.py
+++ b/orchestrator/api/api_v1/endpoints/processes.py
@@ -118,6 +118,9 @@ def resume_process_endpoint(
     if process.last_status == ProcessStatus.RUNNING:
         raise_status(HTTPStatus.CONFLICT, "Resuming a running workflow is not possible")
 
+    if process.last_status == ProcessStatus.RESUMED:
+        raise_status(HTTPStatus.CONFLICT, "Resuming a resumed workflow is not possible")
+
     user_name = user.user_name if user else SYSTEM_USER
 
     broadcast_func = api_broadcast_process_data(request)
--- a/orchestrator/services/celery.py
+++ b/orchestrator/services/celery.py
@@ -69,6 +69,8 @@ def _celery_resume_process(
     trigger_task = get_celery_task(task_name)
     result = trigger_task.delay(pstat.pid, user_inputs, user)
 
+    _celery_set_process_status_resumed(process)
+
     # Enables "Sync celery tasks. This will let the app wait until celery completes"
     if app_settings.TESTING:
         pid = result.get()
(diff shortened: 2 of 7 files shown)


pboers1988 (approved):
LGTM

Mark90:
When does a RESUMED process transition to RUNNING?

Mark90:
@hanstrompert @pboers1988 perhaps we could think about adding python-statemachine to the project?

Mark90 (approved):

coleshaw:
> When does a RESUMED process transition to RUNNING?
> 
> edit: I see this happens in `thread_resume_process`, which based on its name I assumed to be part of the threadpoolexecutor, but it seems to be used for both.

Yeah, I think it happens in the method you identified, the same as it did previously ... so as soon as a Celery worker picks up the process task / workflow, the `last_status` value would change from `RESUMED` -> `RUNNING`.

PR #252, opened 2023-04-24
Graphql with sqlmodel

tjeerddie:
- change create_engine and Session to `sqlmodel` version.
- add sqlmodel for process and step.
- add from_pydantic_with_extra to handle details
- - steps that haven't run need to be added with 'pending'.
- - optional form that needs to be generated.
- - optional current_state that comes from ProcessStat.state.
- move Sort type to graphql/types.
- add tests for graphql processes resource.
- remove Edge class because it's not used.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,6 +54,7 @@ repos:
           - types-redis
           - types-orjson
           - types-tabulate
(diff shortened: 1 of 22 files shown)


pboers1988:
Closing will revisit if and when SQLmodel is deem suitable.

PR #260, opened 2023-05-10
Create worker-status endpoint

dmarjenburgh:
Instead returning the worker status in the EngineSettings, there is a separate `api/settings/worker-status` endpoint.
The reason is to maintain compatibility with the current engine-settings sync-method via websockets.

### DIFF
--- a/orchestrator/api/api_v1/endpoints/settings.py
+++ b/orchestrator/api/api_v1/endpoints/settings.py
@@ -92,9 +92,10 @@ async def set_global_status(
     return status_response
 
 
+@router.get("/worker-status", response_model=WorkerStatus)
 def get_worker_status() -> WorkerStatus:
     """
-    Return information job workers and queues.
+    Return data on job workers and queues.
 
     Returns:
     - The number of queued jobs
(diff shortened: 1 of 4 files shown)


tjeerddie (approved):

pboers1988 (approved):

orchestrator/api/api_v1/endpoints/settings.py
```diff
@@ -92,9 +92,10 @@ async def set_global_status(
     return status_response
 
 
+@router.get("/worker-status", response_model=WorkerStatus)
 def get_worker_status() -> WorkerStatus:
     """
-    Return information job workers and queues.
+    Return data on job workers and queues.
```
mrijk, line 98: Or 'statistics'? Data is very generic

orchestrator/services/tasks.py
```diff
@@ -136,7 +136,7 @@ def __init__(self) -> None:
         self.number_of_workers_online = len(stats)
 
         def sum_items(d: dict) -> int:
-            return sum(len(l) for _, l in d.items())
+            return sum(len(l) for _, l in d.items()) if d else 0
```
mrijk, line 139: not sure why the 'if d' is needed now. Empty dict should be fine.

Also you could probably use:

return sum(len(l) for l in d.values())

PR #262, opened 2023-05-11
implement setting endpoint in graphql

Sparrow1029:
Cut this from the `feat/issue-237-product-endpoint-strawberry` branch with the expectation that that branch will be merged first. I will rebase this one

Looking for any input on the approach to mutations -- also think we may need to add a code path for `OATH2_ACTIVE=False` for mutations similar to query fields... something like

```python
class IsAuthenticatedForMutation(BasePermission):
    message = "User is not authenticated"

    async def has_permission(self, source: Any, info: OauthInfo, **kwargs) -> bool:  # type: ignore
        if not oath2lib_settings.OAUTH2_ACTIVE:
           return True
        mutations_active = oauth2lib_settings.OAUTH2_ACTIVE and oauth2lib_settings.MUTATIONS_ENABLED
```

**\*EDIT:** Circumvented this for local/testing by using `oauth2_lib.settings.oauth2lib_settings.ENVIRONMENT_IGNORE_MUTATION_DISABLED` attribute set to `["local", "TESTING"]`.

### DIFF
--- a/orchestrator/graphql/resolvers/settings.py
+++ b/orchestrator/graphql/resolvers/settings.py
@@ -10,9 +10,11 @@ from orchestrator.db import EngineSettingsTable
 from orchestrator.graphql.schemas.errors import Error
 from orchestrator.graphql.schemas.settings import (
     CACHE_FLUSH_OPTIONS,
+    CacheClearResponse,
     CacheClearSuccess,
     EngineSettingsType,
     StatusType,
+    StatusUpdateResponse,
     WorkerStatusType,
 )
 from orchestrator.graphql.types import CustomInfo
@@ -86,9 +88,9 @@ def set_status(info: CustomInfo, global_lock: bool) -> Union[Error, EngineSettin
 
 @strawberry.type(description="Settings endpoint mutations")
 class SettingsMutation:
-    clear_cache: Union[CacheClearSuccess, Error] = authenticated_mutation_field(
+    clear_cache: CacheClearResponse = authenticated_mutation_field(
         resolver=clear_cache, description="Clear a redis cache by name"
     )
-    update_status: Union[EngineSettingsType, Error] = authenticated_mutation_field(
+    update_status: StatusUpdateResponse = authenticated_mutation_field(
         resolver=set_status, description="Update global status of the engine"
     )
--- a/orchestrator/graphql/resolvers/settings.py
+++ b/orchestrator/graphql/resolvers/settings.py
@@ -1,0 +1,99 @@
+import os
+
+import strawberry
+import structlog
+from oauth2_lib.graphql_authentication import authenticated_mutation_field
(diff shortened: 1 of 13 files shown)


test/unit_tests/graphql/test_processes.py
```diff
@@ -37,7 +37,7 @@ def get_processes_query(
     after: int = 0,
     filter_by: Union[list[str], None] = None,
     sort_by: Union[list[dict[str, str]], None] = None,
-) -> str:
+) -> bytes:
```
tjeerddie, line 40: Wondering why you changed to this?

Sparrow1029: I noticed there were a bunch of deprecation warnings running the tests from httpx/starlette/test client(?) like `DeprecationWarning: Use 'content=<...>' to upload raw bytes/text content.` Because of the use of the `data` keyword argument. Figured I would change some of these to use the suggested `content` keyword -- though the `test_client` fixture is used everywhere in the tests and it may be better to just suppress the warning

tjeerddie: ohh, didn't notice the deprecation warnings, do find it better to fix it then suppress, so this is fine 👍

orchestrator/graphql/resolvers/settings.py
```diff
@@ -0,0 +1,99 @@
...
+from orchestrator.settings import ExecutorType, app_settings
+from orchestrator.utils.redis import delete_keys_matching_pattern
+
+logger = structlog.get_logger(__name__)
+
+
+# Queries
+def resolve_settings(info: CustomInfo) -> StatusType:
+    selected_fields = get_selected_fields(info)
+    logger.debug(OAUTH2_ENABLED=os.getenv("OATH2_ENABLED"), MUTATIONS_ENABLED=os.getenv("MUTATIONS_ENABLED"))
```
tjeerddie, line 34: This logging can be removed

tjeerddie (approved):
Nice, only a log that can be removed and typing fix.

orchestrator/graphql/resolvers/settings.py
```diff
@@ -0,0 +1,99 @@
...
+    if name not in CACHE_FLUSH_OPTIONS:
+        return Error(message="Invalid cache name")
+
+    key_name = "orchestrator:*" if name == "all" else f"orchestrator:{name}:*"
+    deleted = await delete_keys_matching_pattern(cache, key_name)
+    return CacheClearSuccess(deleted=deleted)
+
+
+# TODO(alex): Why is mypy complaining here? I thought the strawberry plugin handled this...
+def set_status(info: CustomInfo, global_lock: bool) -> StatusUpdateResponse:  # type: ignore
```
tjeerddie, line 75: It doesn't want the strawberry return type here, you want the normal return of the function.

PR #275, opened 2023-06-03
Add subscriptions graphql schema

tjeerddie:
- use processes_resolver in subscription schema to resolve processes of a subscription.
- use subscriptions_resolver in processes schema to resolve subscriptions of a process.
- add subscriptions db filters and sorting.
- add generic filters to remove duplicated code.
- use subscriptions_resolver in subscriptions schema to resolve depends_on and in_use_by subscriptions.
- make date filter better so you don't need an exact isoformat date.
- test subscriptions schema.

### DIFF
--- a/orchestrator/db/filters/generic_filters/bool_filter.py
+++ b/orchestrator/db/filters/generic_filters/bool_filter.py
@@ -1,0 +1,26 @@
+# Copyright 2019-2023 SURF.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Callable
+
+from sqlalchemy import Column
+
+from orchestrator.db.database import SearchQuery
+
+
+def generic_bool_filter(field: Column) -> Callable[[SearchQuery, str], SearchQuery]:
+    def bool_filter(query: SearchQuery, value: str) -> SearchQuery:
+        value_as_bool = value.lower() in ("yes", "y", "ye", "true", "1", "ja", "insync")
+        return query.filter(field.is_(value_as_bool))
+
+    return bool_filter
--- a/orchestrator/db/filters/generic_filters/bool_filter.py
+++ b/orchestrator/db/filters/generic_filters/bool_filter.py
@@ -20,7 +20,7 @@ from orchestrator.db.database import SearchQuery
 
 def generic_bool_filter(field: Column) -> Callable[[SearchQuery, str], SearchQuery]:
     def bool_filter(query: SearchQuery, value: str) -> SearchQuery:
-        value_as_bool = value.lower() in ("yes", "y", "ye", "true", "1", "ja", "insync")
+        value_as_bool = value.lower() in ("yes", "y", "true", "1", "ja", "j")
         return query.filter(field.is_(value_as_bool))
 
     return bool_filter
--- a/orchestrator/db/filters/generic_filters/range_filter.py
+++ b/orchestrator/db/filters/generic_filters/range_filter.py
@@ -13,6 +13,8 @@
 from datetime import datetime
 from typing import Callable, Optional
 
+import pytz
+from dateutil.parser import parse
 from sqlalchemy import Column
 
 from orchestrator.db.database import SearchQuery
@@ -29,7 +31,7 @@ def convert_to_date(value: str) -> datetime:
     Example date: "2022-07-21T03:40:48+00:00"
     """
     try:
-        return datetime.fromisoformat(value)
+        return parse(value).replace(tzinfo=pytz.UTC)
     except ValueError:
         raise ValueError(f"{value} is not a valid date")
 
--- a/orchestrator/db/filters/generic_filters/range_filter.py
+++ b/orchestrator/db/filters/generic_filters/range_filter.py
@@ -1,0 +1,87 @@
+# Copyright 2019-2023 SURF.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from datetime import datetime
+from typing import Callable, Optional
+
+from sqlalchemy import Column
+
+from orchestrator.db.database import SearchQuery
+
+# from sqlalchemy.sql.expression import ColumnOperators
+from orchestrator.utils.helpers import to_camel
+
+range_operator_list = ["gt", "gte", "lt", "lte", "ne"]
+
+
+def convert_to_date(value: str) -> datetime:
+    """Parse iso 8601 date from string to datetime.
+
+    Example date: "2022-07-21T03:40:48+00:00"
+    """
+    try:
+        return datetime.fromisoformat(value)
+    except ValueError:
+        raise ValueError(f"{value} is not a valid date")
+
+
+def convert_to_int(value: str) -> int:
+    try:
+        return int(value)
+    except ValueError:
+        raise ValueError(f"{value} is not a valid integer")
+
+
+def get_filter_value_convert_function(field: Column) -> Callable:
+    if field.type.python_type == datetime:
+        return convert_to_date
+    if field.type.python_type == int:
+        return convert_to_int
+    return lambda x: x
+
+
+def get_range_filter(range_type: str, field: Column) -> Callable:
+    if range_type == "gt":
+        return field.__gt__
+    if range_type == "gte":
+        return field.__ge__
+    if range_type == "lt":
+        return field.__lt__
+    if range_type == "lte":
+        return field.__le__
+    if range_type == "ne":
+        return field.__ne__
+    raise ValueError(f"{range_type} not a valid range type ({range_operator_list})")
+
+
+def generic_range_filter(range_type: str, field: Column) -> Callable[[SearchQuery, str], SearchQuery]:
+    filter_operator = get_range_filter(range_type, field)
+    convert_filter_value = get_filter_value_convert_function(field)
+
+    def use_filter(query: SearchQuery, value: str) -> SearchQuery:
+        if val := convert_filter_value(value):
+            value = val
+        return query.filter(filter_operator(value))
+
+    return use_filter
+
+
+def generic_range_filters(
+    column: Column, column_alias: Optional[str] = None
+) -> dict[str, Callable[[SearchQuery, str], SearchQuery]]:
+    dict_filters = {}
+    for operator in range_operator_list:
+        dict_filters[f"{to_camel(column_alias or column.name)}{operator.capitalize()}"] = generic_range_filter(
+            operator, column
+        )
+
+    return dict_filters
--- a/orchestrator/db/filters/generic_filters/range_filter.py
+++ b/orchestrator/db/filters/generic_filters/range_filter.py
@@ -11,6 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from datetime import datetime
+from functools import partial
 from typing import Callable, Optional
 
 import pytz
@@ -22,7 +23,13 @@ from orchestrator.db.database import SearchQuery
 # from sqlalchemy.sql.expression import ColumnOperators
 from orchestrator.utils.helpers import to_camel
 
-range_operator_list = ["gt", "gte", "lt", "lte", "ne"]
+RANGE_TYPES = {
+    "gt": Column.__gt__,
+    "gte": Column.__ge__,
+    "lt": Column.__lt__,
+    "lte": Column.__le__,
+    "ne": Column.__ne__,
+}
 
 
 def convert_to_date(value: str) -> datetime:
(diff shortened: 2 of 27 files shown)


orchestrator/db/filters/generic_filters/range_filter.py
```diff
@@ -0,0 +1,89 @@
...
+def get_filter_value_convert_function(field: Column) -> Callable:
+    if field.type.python_type == datetime:
+        return convert_to_date
+    if field.type.python_type == int:
+        return convert_to_int
+    return lambda x: x
+
+
+def get_range_filter(range_type: str, field: Column) -> Callable:
+    if range_type == "gt":
```
mrijk, line 55: We can't use  match/case you because of older Python version support, but just putting this in a dict and a simple lookup.get() is cleaner

orchestrator/db/filters/generic_filters/range_filter.py
```diff
@@ -0,0 +1,89 @@
...
+    raise ValueError(f"{range_type} not a valid range type ({range_operator_list})")
+
+
+def generic_range_filter(range_type: str, field: Column) -> Callable[[SearchQuery, str], SearchQuery]:
+    filter_operator = get_range_filter(range_type, field)
+    convert_filter_value = get_filter_value_convert_function(field)
+
+    def use_filter(query: SearchQuery, value: str) -> SearchQuery:
+        if val := convert_filter_value(value):
+            value = val
```
mrijk, line 74: Overwriting function parameters quickly becomes confusing. If convert_filter_value just returns the value unmodified if there is no filter, than you don't need the if statement.

orchestrator/db/filters/generic_filters/range_filter.py
```diff
@@ -0,0 +1,89 @@
...
+        return query.filter(filter_operator(value))
+
+    return use_filter
+
+
+def generic_range_filters(
+    column: Column, column_alias: Optional[str] = None
+) -> dict[str, Callable[[SearchQuery, str], SearchQuery]]:
+    dict_filters = {}
+    for operator in range_operator_list:
```
mrijk, line 84: This can probably become a one-liner with list/dict comprehension.

orchestrator/db/filters/process.py
```diff
@@ -112,16 +96,17 @@ def target_filter(query: SearchQuery, value: str) -> SearchQuery:
 
 
 VALID_FILTER_FUNCTIONS_BY_COLUMN: dict[str, Callable[[SearchQuery, str], SearchQuery]] = {
```
mrijk, line 98: Just curious on how this is going to work out when extending queries. Will we be able to easily add additional filter functions, etc.?

orchestrator/db/sorting/subscription.py
```diff
@@ -0,0 +1,20 @@
+from orchestrator.db import SubscriptionTable
+from orchestrator.db.sorting.sorting import generic_sort
+
+VALID_SORT_KEY_LIST = [
+    "subscription_id",
+    "product_id",
+    "customer_id",
```
mrijk, line 7: customer_id?

Also, is this list different from just all the fields in a subscription?

orchestrator/db/sorting/subscription.py
```diff
@@ -0,0 +1,20 @@
...
+    "name",
+    "description",
+    "customer_descriptions",
```
mrijk, line 10: Not sure how sorting this will work out with multiple descriptions.

orchestrator/graphql/schemas/subscription.py
```diff
@@ -0,0 +1,159 @@
...
+        from orchestrator.services.subscriptions import query_in_use_by_subscriptions
+
+        in_use_by_query = query_in_use_by_subscriptions(self.subscription_id)
+        query_results = in_use_by_query.with_entities(SubscriptionTable.subscription_id).all()
+        subscription_ids = ",".join([str(s.subscription_id) for s in query_results])
+        filter_by = (filter_by or []) + [GraphqlFilter(field="subscriptionIds", value=subscription_ids)]
+        return await resolve_subscription(info, filter_by, sort_by, first, after)
+
+    @authenticated_field(description="Returns list of subscriptions that this subscription depends on")  # type: ignore
+    async def depends_on_subscriptions(
```
mrijk, line 144: Lots of similar code to in_use_by_subscriptions

orchestrator/graphql/schemas/process.py
```diff
@@ -76,3 +83,19 @@ class ProcessType:
...
+        filter_by: Union[list[GraphqlFilter], None] = None,
+        sort_by: Union[list[GraphqlSort], None] = None,
+        first: int = 10,
+        after: int = 0,
+    ) -> Connection[Annotated["SubscriptionType", strawberry.lazy(".subscription")]]:
+        from orchestrator.graphql.resolvers.subscription import resolve_subscription
+
+        process: ProcessTable = ProcessTable.query.options(load_only(ProcessTable.process_subscriptions)).get(self.id)
+        subscription_ids = [str(s.subscription_id) for s in process.process_subscriptions]
+        filter_by = (filter_by or []) + [GraphqlFilter(field="subscriptionIds", value=",".join(subscription_ids))]
```
mrijk, line 100: I'm not a big fan of overwriting variables

orchestrator/db/filters/generic_filters/bool_filter.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import Callable
+
+from sqlalchemy import Column
+
+from orchestrator.db.database import SearchQuery
+
+
+def generic_bool_filter(field: Column) -> Callable[[SearchQuery, str], SearchQuery]:
+    def bool_filter(query: SearchQuery, value: str) -> SearchQuery:
+        value_as_bool = value.lower() in ("yes", "y", "ye", "true", "1", "ja", "insync")
```
mrijk, line 23: "ye" ?

tjeerddie: I kept boolean filter the same as how it was done in the endpoint filter.

do we want to change the versions of "yes" that we want to support and if so, which ones?

orchestrator/db/filters/process.py
tjeerddie: that would be the idea yea.

orchestrator/db/sorting/subscription.py
tjeerddie: did a test, added:
```
from orchestrator.db.filters.subscription import VALID_FILTER_FUNCTIONS_BY_COLUMN
from orchestrator.db.filters.generic_filters import generic_values_in_column_filter

VALID_FILTER_FUNCTIONS_BY_COLUMN.update({
    "test_filter": generic_values_in_column_filter(SubscriptionTable.status)
})
```
to our specific version, which gives the `test_filter` as new filterable in subscriptions query.

tjeerddie: does make the `VALID_FILTER_FUNCTIONS_BY_COLUMN` to generic, so will update the naming to be specific for the resource.

tjeerddie: commented on the wrong review comment

acidjunk: is customer_descriptions SURF specific?

test/unit_tests/graphql/test_processes.py
```diff
@@ -81,6 +81,70 @@ def get_processes_query(
...
+query ProcessQuery($first: Int!, $after: Int!, $sortBy: [GraphqlSort!], $filterBy: [GraphqlFilter!]) {
+  processes(first: $first, after: $after, sortBy: $sortBy, filterBy: $filterBy) {
+    page {
+      assignee
+      createdBy
+      failedReason
+      isTask
+      lastStep
+      traceback
+      customer
```
acidjunk, line 100: I might be wrong about this: when customer will be removed from subscription in the GPL version. This should also be removed?

test/unit_tests/graphql/test_processes.py
```diff
@@ -254,10 +318,11 @@ def test_processes_filtering_with_invalid_filter(
                     "status",
                     "workflow",
                     "creator",
-                    "organisation",
+                    "customerId",
```
acidjunk, line 321: Another comment about customer ;)

Feel free to ignore; there might be other tickets that scope the removal of customer from core.

tjeerddie: yea, removed it 👍

orchestrator/db/sorting/subscription.py
tjeerddie: yea, like customer id, removed it.

orchestrator/db/sorting/subscription.py
tjeerddie: pretty much all subscription fields, tried just taking all the fields from the table, but then you will have to filter all the extra functions and vars, so thought explicit would be more appropriate.

tjeerddie: does need to change to something similar to filtering, since you want to be able to sort on other tables

tjeerddie: will be done in a new PR

orchestrator/db/filters/generic_filters/bool_filter.py
tjeerddie: changing it to `("yes", "y", "true", "1", "ja", "j")`

orchestrator/graphql/schemas/subscription.py
tjeerddie: I can't see a improvement of splitting stuff without making it even less readable.

orchestrator/db/filters/generic_filters/bool_filter.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import Callable
+
+from sqlalchemy import Column
+
+from orchestrator.db.database import SearchQuery
+
+
+def generic_bool_filter(field: Column) -> Callable[[SearchQuery, str], SearchQuery]:
+    def bool_filter(query: SearchQuery, value: str) -> SearchQuery:
+        value_as_bool = value.lower() in ("yes", "y", "true", "1", "ja", "j")
```
Mark90, line 23: 🙈  "ja"

orchestrator/db/filters/generic_filters/range_filter.py
```diff
@@ -0,0 +1,86 @@
...
+def get_range_filter(range_type: str, field: Column) -> Callable:
+    range_types = {
+        "gt": field.__gt__,
+        "gte": field.__ge__,
+        "lt": field.__lt__,
+        "lte": field.__le__,
+        "ne": field.__ne__,
+    }
+    if range_type in range_types:
+        return range_types[range_type]
```
Mark90, line 63: Perhaps it's possible to define this at the module level like so
```
RANGE_TYPES = {
        "gt": Column.__gt__,
...
```

Then add the "self" parameter by returning `partial(range_type, field)`. What do you think? It would eliminate the need for the `range_operator_list`

tjeerddie: That is a good one, it also makes this entire function useless 👍

orchestrator/graphql/resolvers/subscription.py
```diff
@@ -0,0 +1,81 @@
...
+from orchestrator.graphql.schemas.subscription import SubscriptionType
+from orchestrator.graphql.types import CustomInfo, GraphqlFilter, GraphqlSort
+
+logger = structlog.get_logger(__name__)
+
+
+def handle_subscription_error(info: CustomInfo) -> CallableErrorHander:
+    def _handle_subscription_error(message: str, **kwargs) -> None:  # type: ignore
+        logger.debug(message, **kwargs)
+        extra_values = dict(kwargs.items()) if kwargs else {}
```
Mark90, line 35: This line is unnecessary, `kwargs` is already a dictionary

orchestrator/graphql/resolvers/subscription.py
```diff
@@ -0,0 +1,81 @@
...
+async def resolve_subscription(
+    info: CustomInfo,
+    filter_by: Union[list[GraphqlFilter], None] = None,
+    sort_by: Union[list[GraphqlSort], None] = None,
+    first: int = 10,
+    after: int = 0,
+) -> Connection[SubscriptionType]:
+    _error_handler = handle_subscription_error(info)
+
+    _range: Union[list[int], None] = [after, after + first] if after is not None and first else None
```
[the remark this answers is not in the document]
tjeerddie: used for the logging but indeed, the if statements are useless since `after` and `first` cant be `None` anymore.

orchestrator/graphql/resolvers/subscription.py
```diff
@@ -0,0 +1,81 @@
...
+def handle_subscription_error(info: CustomInfo) -> CallableErrorHander:
+    def _handle_subscription_error(message: str, **kwargs) -> None:  # type: ignore
+        logger.debug(message, **kwargs)
+        extra_values = dict(kwargs.items()) if kwargs else {}
+        info.context.errors.append(GraphQLError(message=message, path=info.path, extensions=extra_values))
+
+    return _handle_subscription_error
+
+
+async def resolve_subscription(
```
Mark90, line 41:
```suggestion
async def resolve_subscriptions(
```

For consistency with the other resolvers :)

orchestrator/graphql/schemas/subscription.py
```diff
@@ -0,0 +1,172 @@
...
+            return is_resource_type(value) and requested_resource_type(key) and key not in ("id", "parent")
+
+        return SubscriptionProductBlock(
+            id=product_block["id"],
+            parent=product_block.get("parent"),
+            owner_subscription_id=product_block["owner_subscription_id"],
+            resource_types={to_camel(k): v for k, v in product_block.items() if included(k, v)},
+        )
+
+    product_blocks = (to_product_block(product_block) for product_block in get_all_product_blocks(subscription, tags))
```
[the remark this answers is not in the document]
Mark90: Edit: I see that we're just returning a list of product blocks, so it will be even easier

Something like;

```
for path in product_block_paths(subscription):
    product_block_dict = getattr_in(subscription, path)
    to_product_block(product_block_dict)
```

Mark90 (approved):
Few small remarks but looks good, nice refactoring @ the filters!

orchestrator/graphql/schemas/subscription.py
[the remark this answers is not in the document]
tjeerddie: not just the product blocks, it does have a aditional id and parent ids

orchestrator/db/filters/generic_filters/bool_filter.py
tjeerddie: won't leave the MR hanging for this, can be updated later.

mrijk (approved):

Issue #266, opened 2023-05-22
Support for adding generic metadata to subscriptions


dmarjenburgh:
## Feature description 

This feature allows workfloworchestrator developers to add various kinds of metadata to subscriptions in their workflows. The goal is to support metadata for subscriptions in a way that doesn't require custom database schema changes. It can be viewed as an extension of the *note* field that is present on subscriptions currently. The supported metadata should be semi-static (slow-changing) and flexible.

### Rationale

There have been a number of use-cases where more data needs to be attached to subscriptions than is currently supported out-of-the-box. The type of data varies per domain and use-case, but some examples could be:

- External references (like 3rd-part ids)
- Free text descriptions
- Location / contact information

By supporting flexible metadata objects to subscriptions, less custom work is required by workfloworchestrator developers to support their use-case.

### Design and implementation

We propose to add a `subscription_metadata` table with the following schema:

| column_name | type |
|--------|--------|
| subscription_id | UUID (references subscription)|
| metadata | JSON | 

The table contents can be managed by custom workflows with a JSON schema that match the required use-case. An basic endpoint will be available at `GET api/subscriptions/{subscription_id}/metadata` that returns the requested metadata in a JSON response.
**Optional:** The endpoint can be extended to accept a URI-encoded JSONPath filter in a query parameter.

Tasks:

- [ ] Add db migration to create the subscription_metadata table
- [ ] Define SQLAlchemy objects for interacting with the metadata
- [ ] Implement api endpoint for querying the metadata

Optional or future tasks:

- [ ] Migrate existing subscription note to the generic metadata object
- [ ] Add filter options to subscription metadata endpoint (e.g. a `filter` query-param with a URI-encoded JSONPath string)
- [ ] Migrate subscription_customer_descriptions to the generic metadata object

acidjunk:
We discussed using this in the frontend, this afternoon: it would be nice to know the types. Maybe good to consider using JSON scheme

https://json-schema.org/specification.html

hanstrompert:
Good suggestion. We will revisit this issue after an initial implementation of the metadata endpoint. For now the GUI will just display all metadata as strings without any type specific processing.

PR #276, opened 2023-06-08
Subscription metadata support

dmarjenburgh:
Features for #266 

- Add subscription_metadata table
- Add SubscriptionMetadataTable db model
- Add subscription metadata API endpoint


orchestrator/api/api_v1/endpoints/subscriptions.py
```diff
@@ -296,3 +296,11 @@ def failed_processes() -> List[str]:
             logger.info("Subscription already in sync")
     except ValueError as e:
         raise_status(HTTPStatus.NOT_FOUND, str(e))
+
+
+@router.get("/{subscription_id}/metadata", response_model=None, status_code=HTTPStatus.OK)
+def subscription_metadata(subscription_id: UUID, response: Response) -> Optional[dict]:
+    md = get_subscription_metadata(subscription_id)
+    if md is None:
+        response.status_code = HTTPStatus.NOT_FOUND
```
mrijk, line 305: Of course the question is always if you want to return nothing (an empty dict) or a 404.  I'm slightly in favour of an empty dictionary since optionals always introduce additional checking.

mrijk (approved):
LGTM

orchestrator/api/api_v1/endpoints/subscriptions.py
Mark90: What I think makes most sense is:
* If the subscription exists -> return 200 with body `metadata or {}`
* If the subscription does not exist -> return 404

(I recall similar discussions on the customer descriptions endpoint, last year or so :) )

Mark90 (approved):

PR #295, opened 2023-06-30
Generate graphql domain models

tjeerddie:


pyproject.toml
```diff
@@ -65,7 +65,7 @@ dependencies = [
     "nwa-stdlib~=1.4.8",
     "oauth2-lib~=1.3.1",
     "tabulate==0.9.0",
-    "strawberry-graphql==0.171.1"
+    "strawberry-graphql@git+https://github.com/tjeerddie/strawberry#egg=pydantic-convert-specialized-list-class"
```
tjeerddie, line 68: using my forked version until it gets merged

https://github.com/strawberry-graphql/strawberry/pull/2909

orchestrator/graphql/add_graphql.py
```diff
@@ -0,0 +1,114 @@
...
+def add_graphql(app: OrchestratorCore) -> OrchestratorCore:
+    strawberry_models = GRAPHQL_MODELS
+    strawberry_enums: EnumList = {}
+    products = {
+        product_type.__base_type__.__name__: product_type.__base_type__
+        for product_type in SUBSCRIPTION_MODEL_REGISTRY.values()
+        if product_type.__base_type__
+    }
+    for key, product_type in products.items():
+        add_class_to_strawberry(graphql_name(key), product_type, strawberry_models, strawberry_enums, True)
```
mrijk, line 102: Probably can be rewritten so that you don't have to pass strawberry_models but instead use a generator (with yield) and create it here.

orchestrator/graphql/resolvers/subscription.py
```diff
@@ -11,21 +11,74 @@
...
+
+    def get_selections(selected_field: Selection) -> list[Selection]:
+        def has_field_name(selection: Selection, field_name: str) -> bool:
+            return isinstance(selection, SelectedField) and selection.name == field_name
+
+        page_field = [selection for selection in selected_field.selections if has_field_name(selection, "page")]
+
+        if not page_field:
+            return selected_field.selections
+        return one(page_field).selections if page_field else []
```
mrijk, line 59: I would expect that page_field always has a value (given line 57)

orchestrator/graphql/add_graphql.py
tjeerddie: `strawberry_models` needs to be given so we can check if a strawberry type is already created for a model.

.github/workflows/run-unit-tests.yml
```diff
@@ -50,7 +50,39 @@ jobs:
...
+      - name: Run test_single_subscription_product_list_union_type_provisioning_subscription Unit test
+        run: CACHE_URI=redis://redis DATABASE_URI=postgresql://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_HOST/$POSTGRES_DB  pytest ./test/unit_tests/ -k test_single_subscription_product_list_union_type_provisioning_subscription
+        env:
+          POSTGRES_DB: orchestrator-core-test
+          POSTGRES_USER: nwa
+          POSTGRES_PASSWORD: nwa
+          POSTGRES_HOST: postgres
+          ENVIRONMENT: TESTING
+      - name: Run test_single_subscription_product_list_union_type_terminated_subscription Unit test
+        run: CACHE_URI=redis://redis DATABASE_URI=postgresql://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_HOST/$POSTGRES_DB  pytest ./test/unit_tests/ -k test_single_subscription_product_list_union_type_terminated_subscription
```
Mark90, line 85: This was temporary I guess? It seems like it's not running all tests from `test/unit_tests/` at the moment

orchestrator/app.py
```diff
@@ -180,6 +187,26 @@ def register_subscription_models(product_to_subscription_model_mapping: Dict[str
...
+        strawberry_models = GRAPHQL_MODELS
+        strawberry_enums: EnumList = {}
+        products = {
+            product_type.__base_type__.__name__: product_type.__base_type__
+            for product_type in SUBSCRIPTION_MODEL_REGISTRY.values()
+            if product_type.__base_type__
+        }
+        for key, product_type in products.items():
+            add_class_to_strawberry(
+                model_name=graphql_subscription_name(key),
```
Mark90, line 200: What happens if `add_class_to_strawberry` is called for different product types but with the same `model_name=graphql_subscription_name(key)`?

orchestrator/graphql/add_graphql.py
```diff
@@ -0,0 +1,105 @@
+# Copyright 2019-2023 SURF.
```
Mark90, line 1: Maybe call this module `autoregistration.py`?

orchestrator/app.py
```diff
@@ -180,6 +187,26 @@ def register_subscription_models(product_to_subscription_model_mapping: Dict[str
         """
         SUBSCRIPTION_MODEL_REGISTRY.update(product_to_subscription_model_mapping)
 
+    def register_graphql(self: "OrchestratorCore", query: Any = Query, mutation: Any = Mutation) -> None:
```
Mark90, line 190: The graphql endpoint used to be enabled by default but now it looks like it's opt-in, as this function is not called in `OrchestratorCore.__init__`. Is that intentional?

orchestrator/graphql/add_graphql.py
```diff
@@ -0,0 +1,105 @@
...
+    with_interface: bool = False,
+) -> None:
+    strawberry_enums = create_strawberry_enums(model, strawberry_enums)
+
+    product_blocks_types_in_model = get_depends_on_product_block_type_list(model._get_depends_on_product_block_types())
+    for field in product_blocks_types_in_model:
+        if is_not_strawberry_type(field.__name__, strawberry_models) and field.__name__ != model_name:
+            add_class_to_strawberry(field.__name__, field, strawberry_models, strawberry_enums)
+
+    strawberry_name = graphql_name(model_name)
```
Mark90, line 101: The way that the name of a model is determined is now a bit scattered among the `register_graphql()` function (for SubscriptionModel) and here in this function (for ProductBlockModel). 

I think it would be better to only pass `model` to this function, and then let this function (or another helper function) decide the `strawberry_name` with something like a 

```py
if isinstance(model, SubscriptionModel):
    strawberry_name = graphql_subscription_name(model.__base_type__.__name__)
elif isinstance(model, ProductBlockModel):
    strawberry_name = graphql_name(model.__name__)
else:
    raise TypeError("some helpful error message")
```

What do you think?

Mark90: Also it might be helpful to add an extra check and debug logging (I'll admit I'm a bit of a logging addict..)

```py
if strawberry_name in strawberry_models:
    logger.debug("Skip already registered strawberry model", model=repr(model), strawberry_name=strawberry_name)
    return None

logger.debug("Registering strawberry model", model=repr(model), strawberry_name=strawberry_name)
```

orchestrator/graphql/add_graphql.py
```diff
@@ -0,0 +1,105 @@
...
+
+import strawberry
+from strawberry.experimental.pydantic.conversion_types import StrawberryTypeFromPydantic
+
+from orchestrator.domain.base import DomainModel, get_depends_on_product_block_type_list
+from orchestrator.graphql.schema import StrawberryModelType
+from orchestrator.graphql.schemas.subscription import SubscriptionInterface
+from orchestrator.utils.helpers import to_camel
+
+EnumList = dict[str, EnumMeta]
```
Mark90, line 26: It's not a list :p

orchestrator/graphql/resolvers/subscription.py
```diff
@@ -14,18 +14,67 @@
...
+# needs subscription details or not, and use that information here. Left as an exercise for the reader.
+base_sub_props = (
+    [to_lower_camel(key) for key in SubscriptionInterface.__annotations__]
+    + [to_lower_camel(key) for key in ProductModelGraphql.__annotations__]
+    + ["__typename"]
+)
+
+
+def _has_subscription_details(info: CustomInfo) -> bool:
+    """Check if the query asks for subscription details (surf specific properties)."""
```
Mark90, line 45: "surf specific" should probably become "product specific"

orchestrator/graphql/schema.py
```diff
@@ -0,0 +1,146 @@
...
+                StrawberryLogger.logger.debug(message)
+            else:
+                StrawberryLogger.error(error, execution_context)
+
+
+def custom_context_dependency(
+    get_current_user: Callable = Depends(get_oidc_user),  # noqa: B008
+    get_opa_decision: Callable = Depends(get_opa_security_graphql),  # noqa: B008
+) -> CustomContext:
+    return CustomContext(get_current_user=get_current_user, get_opa_decision=get_opa_decision)
```
Mark90, line 126: I realize I forgot to change this in the my oauth2lib PR (#298); but `CustomContext` has been moved to `oauth2_lib.strawberry.OauthContext`. Could you change that in this PR?

orchestrator/graphql/schemas/subscription.py
```diff
@@ -166,3 +93,16 @@ async def depends_on_subscriptions(
...
         return await resolve_subscriptions(info, filter_by_with_related_subscriptions, sort_by, first, after)
+
+
+@strawberry.experimental.pydantic.type(model=SubscriptionModel, all_fields=True)
+class SubscriptionPydantic(SubscriptionInterface):
+    pass
+
+
+@strawberry.federation.type(description="Virtual base class for subscriptions", keys=["subscriptionId"])
+class Subscription(SubscriptionPydantic):
```
Mark90, line 104: Should this be `class SubscriptionGraphql`? For consistency with the other types

orchestrator/graphql/schemas/product.py
```diff
@@ -36,8 +37,22 @@ async def subscriptions(
...
+    @staticmethod
+    def from_pydantic(model: ProductSchema) -> "ProductType":  # type: ignore
+        graphql_model = ProductPydantic.from_pydantic(model)
+        graphql_model.__class__ = ProductType
+        return graphql_model  # type: ignore
+
+
+@strawberry.experimental.pydantic.type(model=ProductModel, all_fields=True)
+class ProductModelGraphql:
+    pass
```
Mark90, line 58: What does this provide next to the `ProductPydantic` class? Do we need both?

test/unit_tests/api/test_processes_ws.py
```diff
@@ -139,10 +139,12 @@ def test_websocket_process_detail_workflow(test_client, long_running_workflow):
...
-            data = websocket.receive_text()
-            process = json_loads(data)["process"]
-            assert process["workflow_name"] == "long_running_workflow_py"
-            assert process["status"] == ProcessStatus.RUNNING
+
+            # WITH AN UPDATE IT DOES NOT SEND THIS ANYMORE
+            # data = websocket.receive_text()
+            # process = json_loads(data)["process"]
+            # assert process["workflow_name"] == "long_running_workflow_py"
+            # assert process["status"] == ProcessStatus.RUNNING
```
Mark90, line 147: Worth looking into later in a separate ticket?

orchestrator/graphql/schemas/product.py
tjeerddie: `ProductModelGraphql` is specific for `SubscriptionInterface` and only has product data, it does not have things like `productBlocks`

orchestrator/graphql/schema.py
tjeerddie: yea, also good to change the `CustomContext` to `OrchestratorContext`

orchestrator/app.py
tjeerddie: it will create the strawberry class twice and break it, adding a check to see if it exists in the `strawberry_models` dict

.github/workflows/run-unit-tests.yml
tjeerddie: they fail when running with all unit tests.

orchestrator/app.py
tjeerddie: yes, because the domain models can only be auto registered after the products are imported, otherwise the `REGISTERED_DOMAIN_MODELS` will be empty.

.github/workflows/run-unit-tests.yml
Mark90: Ah ok. But we should fix the other tests then eventually :)

orchestrator/app.py
Mark90: Ah right, that makes sense.

Makes me think we should come up with a better way for users to register domain models.. something that would automatically include the graphql model generation

orchestrator/graphql/add_graphql.py
Mark90: Alright! And what do you think about my first suggestion of only passing the DomainModel derivative into this function, and letting the name generation take place here?

test/unit_tests/api/test_processes_ws.py
tjeerddie: Created an issue for this: #309

orchestrator/app.py
tjeerddie: I believe that this is the best way currently, to only do it when registering the graphql model. You can't register anything after `register_graphql`, because it won't be recognized anymore once the graphql is initialized.

.github/workflows/run-unit-tests.yml
tjeerddie: ticket for fixing it: #310

PR #296, opened 2023-07-03
Optimize subscription search re-indexing

dmarjenburgh:
- Put a last refresh timestamp on the subscriptions_search_view metadata
- Update trigger function to check for the last_refresh timestamp
- Don't refresh if last refresh was less than specified interval (currently 120 seconds)

### DIFF
--- a/orchestrator/migrations/versions/schema/2023-06-28_a09ac125ea73_add_throttling_to_refresh_subscriptions.sql
+++ b/orchestrator/migrations/versions/schema/2023-06-28_a09ac125ea73_add_throttling_to_refresh_subscriptions.sql
@@ -1,0 +1,25 @@
+CREATE OR REPLACE FUNCTION refresh_subscriptions_search_view()
+    RETURNS TRIGGER
+    LANGUAGE plpgsql
+AS
+$$
+DECLARE
+    should_refresh     bool;
+    current_epoch      int;
+    last_refresh_epoch int;
+    comment_sql        text;
+BEGIN
+    SELECT extract(epoch from now())::int INTO current_epoch;
+    SELECT coalesce(pg_catalog.obj_description('subscriptions_search'::regclass)::int, 0) INTO last_refresh_epoch;
+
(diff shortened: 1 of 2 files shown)


tjeerddie (approved):
LGTM

orchestrator/migrations/versions/schema/2023-06-28_a09ac125ea73_add_throttling_to_refresh_subscriptions.sql
```diff
@@ -0,0 +1,25 @@
...
+AS
+$$
+DECLARE
+    should_refresh     bool;
+    current_epoch      int;
+    last_refresh_epoch int;
+    comment_sql        text;
+BEGIN
+    SELECT extract(epoch from now())::int INTO current_epoch;
+    SELECT coalesce(pg_catalog.obj_description('subscriptions_search'::regclass)::int, 0) INTO last_refresh_epoch;
```
Mark90, line 13: How does this work?

Mark90: Ok, after reading the psql docs I understand.

Though I wonder if this is the right function to call as the docs say it's deprecated
image

Also, what does this return on the very first invocation (i.e. when there is no comment yet)?

Mark90 (approved):

orchestrator/migrations/versions/schema/2023-06-28_a09ac125ea73_add_throttling_to_refresh_subscriptions.sql
dmarjenburgh: `pg_catalog.obj_description('subscriptions_search'::regclass)` reads the comment (metadata string) on the table. It's converted to an int. `coalesce` ensures that, if there is no data, the timestamp `0` will be used.

pboers1988 (approved):

PR #300, opened 2023-07-05
#299 load resource types together with productblocks

### LINKED ISSUE #299: Optimize endpoint to fetch all products
The /api/products endpoint [consists](https://github.com/workfloworchestrator/orchestrator-core/blob/38124b385fc4cf591951ff03d5c7ab064743bd09/orchestrator/api/api_v1/endpoints/products.py#L34) of 1 SQLAlchemy statement which appears to generate a lot of SQL queries.

Screenshot from tracing a slow call on our development environment (with low load):
image

There must be a better way.

Mark90:
Small improvement to the "get all products" endpoint: resource_types were previously lazy loaded, now they're retrieved in 1 query.

Local tests don't show a big impact on the query time however. This endpoint now performs a few queries to retrieve products, product blocks, resource types, workflows and fixed inputs. The query plans shows each of them performs a sequential scan, which seemed wrong given that there are indexes on all of the ID columns, but this is apparently [a good thing](https://stackoverflow.com/a/5203827/1284043) when all of the table's rows are selected.

Closes #299

### DIFF
--- a/orchestrator/api/api_v1/endpoints/products.py
+++ b/orchestrator/api/api_v1/endpoints/products.py
@@ -21,7 +21,7 @@ from sqlalchemy.orm import joinedload, selectinload
 
 from orchestrator.api.error_handling import raise_status
 from orchestrator.api.models import delete, save, update
-from orchestrator.db import ProductTable, SubscriptionTable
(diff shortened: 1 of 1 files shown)


mrijk (approved):
LGTM

tjeerddie (approved):

Issue #278, opened 2023-06-12
bool and int lists in product blocks don't save false values


tjeerddie:
## current behaviour

If you have a list with zero values like `[10, 0, 20, 30, 40, 0, 0]` in a product block, the next time you get it out from the database, the list is without the `0` values: `[10, 20, 30, 40]`.

The same happens with boolean lists like `[True, False, True, False]` which turns into `[True, True]`. 

test created to check the behaviour.
```
def test_subscription_saves_zero_values(
    test_product_type_one, test_product_sub_block_one, product_one_subscription_1
):
    _, _, ProductTypeOneForTest = test_product_type_one

    subscription = ProductTypeOneForTest.from_subscription(product_one_subscription_1)
    subscription.block.list_field = [10, 0, 20, 30, 40, 0, 0]
    subscription.save()
    subscription = ProductTypeOneForTest.from_subscription(product_one_subscription_1)
    assert subscription.block.list_field == [10, 0, 20, 30, 40, 0, 0]
```
gives error: `AssertionError: assert [10, 20, 30, 40] == [10, 0, 20, 30, 40, 0, 0]`

## wanted behaviour

a list `[10, 0, 20, 30, 40, 0, 0]` should return the values as they were saved: `[10, 0, 20, 30, 40, 0, 0]`.

PR #301, opened 2023-07-06
#278: Fix lists with false values not saved in subscription

tjeerddie:
- remove sorting on _load_instances_values.
- add tests.

Closes: #278


mrijk (approved):
LGTM, but this whole base.py could use a good refactoring (for example using list comprehension and smaller functions) to make this a bit more readable.

Mark90 (approved):

PR #307, opened 2023-07-18
238 remove customer id from subscription

### LINKED ISSUE #238: Remove Customer ID from subscription model

Sparrow1029:
This PR changes `subscriptions.customer_id` and `subscription_customer_description.customer_id` fields from `UUID` type to `VARCHAR`, allowing more flexibility for downstream users of orchestrator core to:
- define one "customer" for all their workflows (themselves, likely)
- have `customer_id` be a foreign key reference into an external system to fetch data about a customer
- implement a customer table with a migration to the orchestrator-core db itself which can contain additional data about customers.

Acceptance Criteria:
- [x] Loosen the restriction on the customer_id column to be a varchar.
- [ ] Furthermore we need to make an example implementation where the following can be set as default:
  - [ ] Fullname
  - [ ] Shortcode
  - [ ] Identifier

- [x] Add configuration option DEFAULT_CUSTOMER_ONLY that will enable GraphQL endpoint /customers that serves the one and only default customer as specified above.

closes #238

### DIFF
--- a/orchestrator/graphql/resolvers/default_customer.py
+++ b/orchestrator/graphql/resolvers/default_customer.py
@@ -2,7 +2,6 @@ from typing import Union
 
 from orchestrator.graphql.pagination import Connection
 from orchestrator.graphql.schemas.default_customer import DefaultCustomerType
-from orchestrator.graphql.schemas.errors import Error
 from orchestrator.graphql.types import GraphqlFilter, GraphqlSort, OrchestratorInfo
 from orchestrator.graphql.utils.to_graphql_result_page import to_graphql_result_page
 from orchestrator.settings import app_settings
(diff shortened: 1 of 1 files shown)


orchestrator/graphql/__init__.py
```diff
@@ -66,6 +68,9 @@ class Query:
         resolver=resolve_settings,
         description="Returns information about cache, workers, and global engine settings",
     )
+    customers: DefaultCustomerType = authenticated_field(
```
tjeerddie, line 71: I think we do want the customers as a `Connection` like the others, even with just 1 default customer

Sparrow1029: Made some changes regarding this comment—let me know if what I did was in line with what you meant

orchestrator/graphql/resolvers/default_customer.py
```diff
@@ -0,0 +1,29 @@
...
+            DefaultCustomerType(
+                fullname=app_settings.DEFAULT_CUSTOMER_FULLNAME,
+                shortcode=app_settings.DEFAULT_CUSTOMER_SHORTCODE,
+                identifier=app_settings.DEFAULT_CUSTOMER_IDENTIFIER,
+            )
+        ]
+        total = len(default_customer_list)
+        return to_graphql_result_page(default_customer_list, first, after, str(total))
+    except Exception as e:
+        return Error(message=f"Error returning default customer: {e}")
```
tjeerddie, line 29: I think its better to log the error and return an empty list when it goes wrong

Mark90: I'm not sure how returning the default customer could go so terribly wrong. No need to catch `Exception` here at least :)

orchestrator/migrations/versions/schema/2023-07-17_165303a20fb1_customer_id_to_varchar.sql
```diff
@@ -0,0 +1,125 @@
...
+    FROM subscriptions s
+        JOIN products p ON s.product_id = p.product_id
+),
+fi_info AS (
+    SELECT s.subscription_id,
+        string_agg(
+            fi.name || ':' || fi.value,
+            ', '
+            ORDER BY fi.name
+        ) AS fi_info
```
Mark90, line 59: I don't think there is any change to the view?

Mark90 (approved):

orchestrator/migrations/versions/schema/2023-07-17_165303a20fb1_customer_id_to_varchar.sql
Sparrow1029: There is no change to the view, it just has to be recreated since there is a change to the `customer_id` column in two of the referenced tables.

tjeerddie (approved):

PR #308, opened 2023-07-20
280 move to pydantic forms

Sparrow1029:
This PR adds [`pydantic_forms`](https://github.com/workfloworchestrator/pydantic-forms/tree/main) library as a dependency, and uses the external library for form processing.


Sparrow1029:
The `ProductId` form validators probably can't move out of the orch-core because they depend on `orchestrator.services` module.

Moving the `network_type_validators` to the `pydantic_forms` library is totally possible, if they seem like they will be universally useful to other consumers of `pydantic_forms` -- I can create a separate PR for that...

tjeerddie (approved):
LGTM

pyproject.toml
```diff
@@ -65,7 +65,8 @@ dependencies = [
     "nwa-stdlib~=1.4.8",
     "oauth2-lib==1.3.2",
     "tabulate==0.9.0",
-    "strawberry-graphql==0.171.1"
+    "strawberry-graphql==0.171.1",
+    "pydantic-forms @ git+https://github.com/workfloworchestrator/pydantic-forms.git@0.2.0",
```
Mark90, line 69: Pydantic forms is [published on pypi ](https://pypi.org/project/pydantic-forms/#description) so `pydantic-forms==0.2.0` should work :)

Sparrow1029: Ah yep! Changed

orchestrator/forms/__init__.py
```diff
@@ -9,153 +9,3 @@
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
-# limitations under the License.
```
Mark90, line 12: Deleted one line too many here

orchestrator/forms/__init__.py
```diff
@@ -9,153 +9,3 @@
...
-class DisplayOnlyFieldType:
-    @classmethod
-    def __get_validators__(cls) -> Generator:
-        yield cls.nothing
-
-    def nothing(cls, v: Any, field: ModelField) -> Any:
-        return field.default
-
-
-class FormPage(BaseModel):
```
Mark90, line 132: I've tried running mypy on our surf orchestrator when using this branch of the orchestrator-core, and this was the result:

image

I didn't really consider this upfront. :) 

We can rewrite all of these, but this effort will also be required from other users of the orchestrator. We'd at least need to bump the package version to `1.3.x` to signal these breaking changes.

For the sake of backwards compatibility and making it easier to migrate, we could consider to import the pydantic_forms types/functions in orchestrator so that the import paths don't change.


For example:
```
# orchestrator/forms/__init__.py

from pydantic_forms.core import FormPage  # noqa: F401
```
(F401 prevents linters from removing the unused import)

This would allow users of the orchestrator-core package to keep doing `from orchestrator.forms import FormPage`, and postpone having to change the imports. (perhaps we can create our own code-migration tool for it someday, similar to https://github.com/asottile/pyupgrade and https://github.com/pydantic/bump-pydantic)

What do you think?

Mark90: Apart of having backwards-compatible imports (or not), we'll at least need a summary in the release notes of which symbols have been removed, and tell users where they can find them in pydantic_forms.

From what I can see in our surf orchestrator mypy results, these are the validators/types that we depend on

```
orchestrator.forms.FormPage
orchestrator.forms.ReadOnlyField
orchestrator.forms.validators.Accept
orchestrator.forms.validators.Choice
orchestrator.forms.validators.ContactPersonList
orchestrator.forms.validators.DisplaySubscription
orchestrator.forms.validators.Divider
orchestrator.forms.validators.Label
orchestrator.forms.validators.ListOfOne
orchestrator.forms.validators.ListOfTwo
orchestrator.forms.validators.LongText
orchestrator.forms.validators.MigrationSummary
orchestrator.forms.validators.OrganisationId
orchestrator.forms.validators.UniqueConstrainedList
orchestrator.forms.validators.choice_list
orchestrator.forms.validators.contact_person_list
orchestrator.forms.validators.migration_summary
orchestrator.types.InputForm
orchestrator.types.FormGenerator
orchestrator.types.SummaryData
```

Mark90 (changes requested):

orchestrator/forms/__init__.py
Sparrow1029: fixed

Mark90 (approved):
Great work! 🚀

Issue #303, opened 2023-07-12
Implement product blocks into GraphQL as resource in the root Query


tjeerddie:
This is necessary for the metadata - product block page.

PR #316, opened 2023-07-27
Add product block resource to graphql root Query

tjeerddie:
Closes: #303

### DIFF
--- a/orchestrator/db/sorting/product_block.py
+++ b/orchestrator/db/sorting/product_block.py
@@ -5,7 +5,7 @@ from orchestrator.db.sorting.sorting import generic_column_sort, generic_sort
 from orchestrator.utils.helpers import to_camel
 
 PRODUCT_BLOCK_SORT_FUNCTIONS_BY_COLUMN = {
-    to_camel(key): generic_column_sort(value) for [key, value] in inspect(ProductBlockTable).columns.items()
+    to_camel(key): generic_column_sort(value) for key, value in inspect(ProductBlockTable).columns.items()
 }
 
 sort_product_blocks = generic_sort(PRODUCT_BLOCK_SORT_FUNCTIONS_BY_COLUMN)
--- a/orchestrator/db/sorting/product_block.py
+++ b/orchestrator/db/sorting/product_block.py
@@ -1,0 +1,11 @@
+from sqlalchemy.inspection import inspect
+
+from orchestrator.db import ProductBlockTable
+from orchestrator.db.sorting.sorting import generic_column_sort, generic_sort
+from orchestrator.utils.helpers import to_camel
+
+PRODUCT_BLOCK_SORT_FUNCTIONS_BY_COLUMN = {
+    to_camel(key): generic_column_sort(value) for [key, value] in inspect(ProductBlockTable).columns.items()
+}
+
+sort_product_blocks = generic_sort(PRODUCT_BLOCK_SORT_FUNCTIONS_BY_COLUMN)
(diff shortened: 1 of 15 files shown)


orchestrator/graphql/resolvers/product_block.py
```diff
@@ -0,0 +1,57 @@
...
+async def resolve_product_blocks(
+    info: CustomInfo,
+    filter_by: Union[list[GraphqlFilter], None] = None,
+    sort_by: Union[list[GraphqlSort], None] = None,
+    first: int = 10,
+    after: int = 0,
+) -> Connection[ProductBlock]:
+    _error_handler = create_resolver_error_handler(info)
+
+    pydantic_filter_by: list[Filter] = [item.to_pydantic() for item in filter_by] if filter_by else []
```
mrijk, line 28: Just curious, do why do we prefix these variables with pydantic? Is that relevant?

mrijk (approved):
LGTM

orchestrator/graphql/resolvers/product_block.py
tjeerddie: to not overwrite `filter_by`.

orchestrator/db/sorting/product_block.py
```diff
@@ -0,0 +1,11 @@
+from sqlalchemy.inspection import inspect
+
+from orchestrator.db import ProductBlockTable
+from orchestrator.db.sorting.sorting import generic_column_sort, generic_sort
+from orchestrator.utils.helpers import to_camel
+
+PRODUCT_BLOCK_SORT_FUNCTIONS_BY_COLUMN = {
+    to_camel(key): generic_column_sort(value) for [key, value] in inspect(ProductBlockTable).columns.items()
```
Mark90, line 8: List syntax is not needed here

```suggestion
    to_camel(key): generic_column_sort(value) for key, value in inspect(ProductBlockTable).columns.items()
```

orchestrator/graphql/resolvers/product_block.py
```diff
@@ -0,0 +1,57 @@
...
+    pydantic_sort_by: list[Sort] = [item.to_pydantic() for item in sort_by] if sort_by else []
+    logger.info(
```
Mark90, line 30: decrease log level to debug?

orchestrator/graphql/resolvers/product_block.py
```diff
@@ -0,0 +1,57 @@
...
+
+    return Connection(
+        page=page_product_blocks,
+        page_info=PageInfo(
+            has_previous_page=bool(after),
+            has_next_page=has_next_page,
+            start_cursor=start_cursor,
+            end_cursor=end_cursor,
+            total_items=total if total else None,
+        ),
```
Mark90, line 56: I think we'll see this pattern (requesting an extra item to determine has_next_page, calculating the cursors etc) in a few more places (if not already).

We could probably make a generic function for line 36-56? Fine with me to do this in the next PR though!

orchestrator/graphql/resolvers/product_block.py
tjeerddie: The logging was kept from the old filtering, I don't mind changing it

Mark90 (approved):

orchestrator/graphql/resolvers/product_block.py
Mark90: With info it's going to be really noisy in production :)

PR #318, opened 2023-07-31
Pin mkdocs lower then 1.5 since it fails

tjeerddie:
mkdocs fails in 1.5 and 1.5.1, there is already an issue: https://github.com/squidfunk/mkdocs-material/issues/5772

hopefully this PR fixes it: https://github.com/squidfunk/mkdocs-material/issues/5778

When the above is merged we can test to remove the pin.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,7 +110,7 @@ test = [
     "types-PyYAML"
 ]
 doc = [
-    "mkdocs",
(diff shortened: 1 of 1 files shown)


Mark90 (approved):

ricardovdheijden (approved):

crankynetman:
@tjeerddie, FYI, I've removed this pin in the branch `feat/partner-meeting-doc-updates` as it seems to be building docs just fine locally on 1.5.3.

PR #326, opened 2023-08-07
Skip sorting on unknown fields

### LINKED ISSUE #322: Passing invalid sort field results in a KeyError
https://github.com/workfloworchestrator/orchestrator-core/blob/3d4c2ffbcc9ed1dea7b277cc6b88e58da5e8e7a2/orchestrator/api/helpers.py#L134

The field name from the sort parameter is assumed to be in the SubscriptionTable. So if somebody provides a wrong fieldname in the URL this triggers a 500 response.

Perform the `SubscriptionTable.__dict__[item[0]` lookup before this if/else and in case of KeyError  call `raise_status(HTTPStatus.BAD_REQUEST, f"Invalid sort field: '{item[0]}'")`

dmarjenburgh:
Fixes #322

### DIFF
--- a/orchestrator/api/helpers.py
+++ b/orchestrator/api/helpers.py
@@ -52,6 +52,26 @@ def _process_text_query(q: str) -> str:
         return q
 
 
+def _add_sort_to_query(query: Query, sort: Optional[List[str]]) -> Query:
+    if sort is None or len(sort) < 2:
+        return query
+
+    for item in chunked(sort, 2):
+        if item and len(item) == 2:
+            order_by_expr = expression.desc if item[1].upper() == "DESC" else expression.asc
+            if item[0] in ["product", "tag"]:
+                field = "name" if item[0] == "product" else "tag"
+                col = ProductTable.__dict__.get(field)
+            else:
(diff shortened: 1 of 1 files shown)


orchestrator/api/helpers.py
```diff
@@ -52,6 +52,26 @@ def _process_text_query(q: str) -> str:
...
+            if item[0] in ["product", "tag"]:
+                field = "name" if item[0] == "product" else "tag"
+                col = ProductTable.__dict__.get(field)
+            else:
+                col = SubscriptionTable.__dict__.get(item[0])
+
+            if col:
+                query = query.order_by(order_by_expr(col))
+            else:
+                logger.warn(f"Unable to add sort. Unknown field: {item[0]}")
```
Mark90, line 71: I don't think we have any use in logging this warning, it's the user (or at least client) who made an error. And imo we should return a 400 error instead of 200

Mark90 (approved):

orchestrator/api/helpers.py
dmarjenburgh: Done

tjeerddie (approved):

PR #337, opened 2023-08-23
Basic websearch_to_tsquery support for subscription resolver

### LINKED ISSUE #288: Support Elastic search query syntax subset on paginated endpoints
Make sure we start supporting the Elastic query syntax on the subscriptions and processes endpoint.

Apply to all endpoints, e.g. subscriptions, products.

Required features of the Elastic Query syntax:
- Field/value filter, single words do not need quotes `insync:true` `description:SURF`
- Field/value filter, multiple words require double quotes: `description:"SURFdomain test"`
- For subscriptions endpoint only -- Value filter, single word without fieldname: `SURF` assumes using the "tsv" field. The given example is the equivalent to `tsv:SURF`

Endpoint should respond with an error when the query is invalid.

Sparrow1029:
This PR adds basic support for querying the `subscriptions_search` (tsv) table via graphql & use of the `websearch_to_tsquery` function builtin to postgresql.

### Proposal from Issue:

The GraphQL endpoint for subscriptions will have the following arguments schema:

    first: Int! (Already there)
    after: Int! (Already there)
    sortBy: [GraphqlSort!] (Already there)
    query: String (Replacement for filterBy and tsv)

Clients will only have to specify pagination (`first`/`after`) options, sorting (`sortBy`) options and a single text query query.

### Not currently implemented (yet) in this PR:

The backend can decide to use the full-text capabilities or parse out specific field filters like `insync` and `status` for better performance (these fields can be filtered on the `subscription` table instead of the `subscriptions_search` view). This keeps the API simple to use.

Closes #288


Sparrow1029:
This needs some cleanup on the work I've done so far (comments to remove, move "REFRESH MATERIALIZED VIEW" functionality for tests to `conftest.py` as a fixture or helper.

Some problems I've encountered:

- Had to skip some tests as the `start_date` and `end_date` fields are not included in the `subscriptions_search` materialized view. These fields may be candidates for parsing out of the `query` string and used as direct filters against the subscription table (instead of performing a schema migration to update the view).
- certain fields like `product_name:<name>` used in the product resolver didn't work as expected—returning multiple subscription results in the `test_single_product_with_subscriptions` test after modifying the product resolver to use the websearch query `f"product_name:{self.name}"`
- the key:value capabilities of `websearch_to_tsquery` seem (in general) dubious, depending on the field being searched.

More tests should probably be written to exercise the text search capabilities.

**Additional Note:**
The `subscriptions_search` materialized view is (for performance reasons) on a 120 second trigger to push updates from the `subscriptions` table. I think this means that if a user creates/modifies a subscription, then it won't be query-able via graphql for 2 minutes, as its data will not have been converted into a tsv in the view yet.

Hoping for good conversation/feedback regarding these changes!

orchestrator/graphql/schemas/process.py
```diff
@@ -108,7 +108,8 @@ async def subscriptions(
...
         if not subscription_ids:
             return EMPTY_PAGE
-        filter_by_with_related_subscriptions = (filter_by or []) + [
-            GraphqlFilter(field="subscriptionIds", value=",".join(subscription_ids))
-        ]
-        return await resolve_subscriptions(info, filter_by_with_related_subscriptions, sort_by, first, after)
+        # filter_by_with_related_subscriptions = (filter_by or []) + [
+        #     GraphqlFilter(field="subscriptionIds", value=",".join(subscription_ids))
+        # ]
+        query_related_subscriptions = " OR ".join(f"subscription_id:{sub_id}" for sub_id in subscription_ids)
```
tjeerddie, line 114: this should become something that should add the user input ontop of the subscription check, so you never get unrelated subscriptions but still able to filter through those.
`(subscription_id in (id_1, id_2)) AND user input`

same for the other relation resolvers.

pboers1988: Never a fan of raw SQL in the code, better to use the SQLalchemy functions.

Sparrow1029: In the proposal section of the #288 issue, it's stated that the `filter_by` parameter should be removed:

> The GraphQL endpoint for subscriptions will have the following arguments schema:
> - `first: Int!` (Already there)
> - `after: Int!` (Already there)
> - `sortBy: [GraphqlSort!]` (Already there)
> - `query: String` (Replacement for `filterBy` and `tsv`)

Also the `OR` is not raw SQL, as it is getting passed into the [`websearch_to_tsquery`](https://www.postgresql.org/docs/current/textsearch-controls.html#id-1.5.11.6.4.11) function, which converts it to `|`—though I agree that `" OR ".join(...)` is not the ideal solution.

Maybe in the case of these sub-queries, we should use async sqlalchemy calls instead of the subscription resolver, if the query string syntax is fully replacing tsv/filterBy?

Sparrow1029: I thought according to the Issue/proposal that the `filterBy` parameter was supposed to not be used in the graphQL query endpoint, and that the `websearch_to_tsquery` is supposed to be replacing it.

The `"OR"`s here admittedly aren't a very clean solution, but they were intended to be passed through the `websearch_to_tsquery` translation function. Maybe here though should just be SQLAlchemy orm queries (as @pboers1988 suggests) instead to get the related subscriptions...

tjeerddie: The point I'm trying to make is that no filtering is passed to the `resolve_subscriptions` underneath this line and it only adds filtering for the subscriptions that are inside the process.

Using sqlalchemy to get the subscriptions instead of `resolve_subscriptions` is an option but I don't have a strong opinion on that.

dmarjenburgh: I think using sqlalchemy instead of `resolve_subscriptions` is the better solution here since the query parameter is really for full-text searching which we don't support for processes. (and have no intention to IIRC)

Sparrow1029:
Closing in favor of #412

PR #339, opened 2023-08-29
Sql alchemy 20

mrijk:
- Udate to SqlAlchemy 2.0

Note: users of orchestrator-core need to be notified that this change is potentially not backward compatible. This impacts:

- migration scripts
- any queries directly on the database


orchestrator/db/models.py
```diff
@@ -77,25 +77,25 @@ def process_bind_param(self, value: Optional[datetime], dialect: Dialect) -> Opt
...
-    assignee = Column(String(50), server_default=Assignee.SYSTEM, nullable=False)
-    last_status = Column(String(50), nullable=False)
-    last_step = Column(String(255), nullable=True)
-    started_at = Column(UtcTimestamp, server_default=text("current_timestamp()"), nullable=False)
-    last_modified_at = Column(UtcTimestamp, server_default=text("current_timestamp()"), onupdate=nowtz, nullable=False)
-    failed_reason = Column(Text())
-    traceback = Column(Text())
-    created_by = Column(String(255), nullable=True)
-    is_task = Column(Boolean, nullable=False, server_default=text("false"), index=True)
+    pid = mapped_column(UUIDType, server_default=text("uuidmapped_cgenerate_v4()"), primary_key=True, index=True)
```
Mark90, line 86: I guess `uuidmapped_cgenerate_v4()` = `uuid_generate_v4()` + accidental paste of `mapped_c`

orchestrator/db/models.py
```diff
@@ -169,15 +169,16 @@ class ProcessSubscriptionTable(BaseModel):
...
-
-    product_id = Column(UUIDType, server_default=text("uuid_generate_v4()"), primary_key=True)
-    name = Column(String(), nullable=False, unique=True)
-    description = Column(Text(), nullable=False)
-    product_type = Column(String(255), nullable=False)
-    tag = Column(String(TAG_LENGTH), nullable=False, index=True)
-    status = Column(String(STATUS_LENGTH), nullable=False)
-    created_at = Column(UtcTimestamp, nullable=False, server_default=text("current_timestamp()"))
-    end_date = Column(UtcTimestamp)
+    __allow_unmapped__ = True
```
Mark90, line 172: What happens without `__allow_unmapped__`?

From what I'm [reading in the guide](__allow_unmapped__) this should only be needed for models that don't use mapped columns, but all columns here are mapped (I think?)

orchestrator/db/models.py
```diff
@@ -200,18 +201,20 @@ class ProductTable(BaseModel):
     __table_args__ = {"extend_existing": True}
 
     def find_block_by_name(self, name: str) -> ProductBlockTable:
-        return (
-            object_session(self).query(ProductBlockTable).with_parent(self).filter(ProductBlockTable.name == name).one()
-        )
+        if session := object_session(self):
+            return session.query(ProductBlockTable).with_parent(self).filter(ProductBlockTable.name == name).one()
+        raise AssertionError("Session should not be None")
```
Mark90, line 206: Seeing as this occurs in 3 places, maybe add a function to the BaseModel class for this?

```
class BaseModel(_Base):


    ...

    def get_session(self) -> Session:
        if session := object_session(self):
            return session
        raise AssertionError("Session should not be None")
```

And use it like suggested
```suggestion
        return self.get_session().query(ProductBlockTable).with_parent(self).filter(ProductBlockTable.name == name).one()
```

orchestrator/db/models.py
```diff
@@ -234,29 +237,27 @@ def workflow_by_key(self, name: str) -> Optional[WorkflowTable]:
...
     __tablename__ = "product_blocks"
-    product_block_id = Column(UUIDType, server_default=text("uuid_generate_v4()"), primary_key=True)
-    name = Column(String(), nullable=False, unique=True)
-    description = Column(Text(), nullable=False)
-    tag = Column(String(TAG_LENGTH))
-    status = Column(String(STATUS_LENGTH))
-    created_at = Column(UtcTimestamp, nullable=False, server_default=text("current_timestamp()"))
-    end_date = Column(UtcTimestamp)
+
+    __allow_unmapped__ = True
```
Mark90, line 252: Same question here as for the other `__allow_unmapped__` occurrence

orchestrator/db/models.py
Mark90: Is it because of the functions perhaps?

orchestrator/db/models.py
```diff
@@ -284,13 +285,15 @@ class ProductBlockTable(BaseModel):
...
-    in_use_by: list[ProductBlockTable] = association_proxy(
+    # in_use_by: List[ProductBlockTable] = association_proxy(
+    in_use_by = association_proxy(
         "in_use_by_block_relations",
         "in_use_by",
         creator=lambda in_use_by: ProductBlockRelationTable(in_use_by=in_use_by),
     )
 
-    depends_on: list[ProductBlockTable] = association_proxy(
+    # depends_on: List[ProductBlockTable] = association_proxy(
```
Mark90, line 295: Contains old commented code

orchestrator/graphql/utils/to_graphql_result_page.py
```diff
@@ -18,6 +18,6 @@ def to_graphql_result_page(items: list[Any], first: int, after: int, total: str)
             has_next_page=has_next_page,
             start_cursor=start_cursor,
             end_cursor=end_cursor,
-            total_items=int(total) if total else None,
+            total_items=total if total else None,
```
Mark90, line 21:
```suggestion
            total_items=total,
```

Otherwise `0` is returned as `None`

orchestrator/migrations/versions/schema/2020-10-19_3323bcb934e7_fix_tsv_triggers.py
```diff
@@ -17,7 +18,8 @@
 def upgrade() -> None:
     conn = op.get_bind()
     conn.execute(
-        """
+        text(
+            """
```
Mark90, line 22: Side note, should we also update the migration generator to output migrations with the queries wrapped in `text()` ?

Mark90 (changes requested):

orchestrator/graphql/utils/to_graphql_result_page.py
mrijk: Yep, so that was wrong in the first place

orchestrator/db/database.py
```diff
@@ -187,7 +186,9 @@ class Database:
     def __init__(self, db_url: str) -> None:
         self.request_context: ContextVar[str] = ContextVar("request_context", default="")
         self.engine = create_engine(db_url, **ENGINE_ARGUMENTS)
-        self.session_factory = sessionmaker(bind=self.engine, **SESSION_ARGUMENTS)
+        self.session_factory = sessionmaker(
+            bind=self.engine, class_=WrappedSession, autocommit=False, autoflush=True, query_cls=SearchQuery
```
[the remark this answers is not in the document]
mrijk: Yes, mypy complains because it is now enforced that these are keyword arguments. The solution that still uses the SESSION_ARGUMENTS looks a bit bulky. I prefer the direct approach so I can see what the values are.

orchestrator/migrations/versions/schema/2020-10-19_3323bcb934e7_fix_tsv_triggers.py
mrijk: Yes.  (Although I really really don't like the output that the migration generator is spitting out.)

orchestrator/db/models.py
mrijk: Oops.

orchestrator/db/models.py
mrijk: I'll double check. I think it's only needed now in SubscriptionInstanceTable because of a 'free' variable.

orchestrator/db/database.py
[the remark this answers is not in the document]
Mark90: Yeah, but somehow the tests should use the same configuration as the real application. Maybe a separate function that calls `sessionmaker()`? Then the test code can call that and only override the `bind` param

orchestrator/graphql/utils/to_graphql_result_page.py
Mark90: Well the old code had a string coming in, so `"0"` would not be converted to None there. Not sure what changed on the calling side but either way it's a bit weird to have `total_items=None`

orchestrator/db/database.py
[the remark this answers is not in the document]
mrijk: I'll create a cleanup/followup ticket that addresses the issues mentioned in this ticket, as well as the remaining SqlAlchemy 2.0 deprecations (that will be removed in 3.0)

Issue #297, opened 2023-07-04
workflow step decorator to handle asynchronous communication with subsystem


hanstrompert:
# Problem

A common way to implement asynchronous communication with a substem in a workflow today uses two workflow steps, one step to prepare and submit the payload to the subsystem, and a following step with an input form that will wait for the subsystem to post back the results of the submitted job. This works just fine when the submitted job succeeded and the workflow can continue to execute the next steps in the workflow. In case the submitted job fails you are stuck, after the problem in the subsystem is fixed it should be possible to resubmit the job, but that is not possible because you are already in a step after the succeeded job submit step. And continuing the workflow is also not an option because the job in the subsystem has failed. This violates the WFO philosophy that states that it should be possible to retry every step until it succeeds and can continue with the next step until the workflow finishes.

# Solution

Design a workflow step decorator that:

* combines the job submit, wait for the result, and validation of the result into one step
* allows the registration of a callback URL to be used by the subsystem to post the result back
* add a new process state to indicate that the process is waiting on a callback

pboers1988:
Must Haves:

- [ ] Use the inputform generator to define the callback protocol
- [ ] Implement a configurable timer on the callback step
- [ ] Make it usable in the form of a `@callback_step` decorator

PR #343, opened 2023-09-05
Feature/step group

dmarjenburgh:
Partial feature implementation for #297

Addition of a `step_group` step.

- A step group is like a mini workflow that acts as a single step.
- A workflow can be suspended and resumed on a sub step

### DIFF
--- a/orchestrator/services/processes.py
+++ b/orchestrator/services/processes.py
@@ -46,6 +46,7 @@ from orchestrator.workflow import Failed, ProcessStat, ProcessStatus, Step, Step
 from orchestrator.workflow import Process as WFProcess
 from orchestrator.workflows import get_workflow
 from orchestrator.workflows.removed_workflow import removed_workflow
+from pydantic_forms.core import post_form
 from pydantic_forms.exceptions import FormValidationError
 
 logger = structlog.get_logger(__name__)
@@ -150,8 +151,31 @@ def _get_current_step_to_update(
     """Checks if last error state is identical to determine if we add a new step or update the last one."""
     step_state: State = process_state.unwrap()
     current_step = None
+    last_db_step = p.steps[-1] if len(p.steps) else None
+    current_sub_step, current_step_group = step_state.get("__sub_step"), step_state.get("__step_group")
+    db_sub_step, db_step_group = (
+        (last_db_step.state.get("__sub_step"), last_db_step.state.get("__step_group")) if last_db_step else (None, None)
+    )
+
+    if current_sub_step is None and db_sub_step is None:
+        pass  # Skip step group logic
+    elif (db_sub_step is None and current_sub_step) or (db_sub_step and current_sub_step is None):
+        # New entry
+        current_step = ProcessStepTable(
+            process_id=stat.process_id,
+            name=current_step_group or db_step_group,
+            status=process_state.status,
+            state=step_state,
+            created_by=stat.current_user,
+        )
+    else:
+        # Update last step entry in db
+        logger.info("Will be merged")
+        last_db_step.status = process_state.status
+        last_db_step.state = step_state
+        current_step = last_db_step
+
     if process_state.isfailed() or process_state.iswaiting():
-        last_db_step = p.steps[-1] if len(p.steps) else None
         if (
             last_db_step is not None
             and last_db_step.status == process_state.status
@@ -174,9 +198,8 @@ def _get_current_step_to_update(
 
     if current_step is None:
         # add a new entry to the process stat
-        logger.info("Adding a new process step with state info")
         current_step = ProcessStepTable(
-            pid=stat.process_id,
+            process_id=stat.process_id,
             name=step.name,
             status=process_state.status,
             state=step_state,
@@ -187,6 +210,7 @@ def _get_current_step_to_update(
     # Test will fail if multiple steps have the same timestamp
     current_step.executed_at = nowtz()
     return current_step
+
 
 def _db_log_step(
     stat: ProcessStat,
@@ -206,7 +230,7 @@ def _db_log_step(
         WFProcess
 
     """
-    p = _update_process(stat.pid, step, process_state)
+    p = _update_process(stat.process_id, step, process_state)
     current_step = _get_current_step_to_update(stat, p, step, process_state)
 
     db.session.add(p)
--- a/orchestrator/services/processes.py
+++ b/orchestrator/services/processes.py
@@ -170,7 +170,6 @@ def _get_current_step_to_update(
         )
     else:
         # Update last step entry in db
-        logger.info("Will be merged")
         last_db_step.status = process_state.status
         last_db_step.state = step_state
         current_step = last_db_step
--- a/orchestrator/services/processes.py
+++ b/orchestrator/services/processes.py
@@ -46,7 +46,6 @@ from orchestrator.workflow import Failed, ProcessStat, ProcessStatus, Step, Step
 from orchestrator.workflow import Process as WFProcess
 from orchestrator.workflows import get_workflow
 from orchestrator.workflows.removed_workflow import removed_workflow
-from pydantic_forms.core import post_form
 from pydantic_forms.exceptions import FormValidationError
 
 logger = structlog.get_logger(__name__)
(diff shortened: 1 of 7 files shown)


test/unit_tests/test_workflow.py
```diff
@@ -453,3 +458,113 @@ def whatever():
...
+    assert log == [
+        ("Start", Success({})),
+        ("Step 1", Success({"steps": [1]})),
+        ("Step 2", Success({"steps": [1, 2], "__sub_step": "Step 2", "__step_group": "Multistep"})),
+        ("Input Name", Suspend({"steps": [1, 2], "__sub_step": "Input Name", "__step_group": "Multistep"})),
+        ("Multistep", Suspend({"steps": [1, 2], "__sub_step": "Input Name", "__step_group": "Multistep"})),
+    ]
+
+
+def test_step_group_with_inputform_resume():
```
tjeerddie, line 532: I think it is usefull to create a test where you make the input step fail

orchestrator/services/processes.py
```diff
@@ -155,54 +138,102 @@ def _db_log_step(
...
+        current_step = ProcessStepTable(
+            process_id=stat.process_id,
+            name=current_step_group or db_step_group,
+            status=process_state.status,
+            state=step_state,
+            created_by=stat.current_user,
+        )
+    else:
+        # Update last step entry in db
+        logger.info("Will be merged")
```
tjeerddie, line 173: remove or improve loging

orchestrator/services/processes.py
```diff
@@ -155,54 +138,102 @@ def _db_log_step(
...
+        last_db_step.status = process_state.status
+        last_db_step.state = step_state
+        current_step = last_db_step
```
tjeerddie, line 176: could change the first `current_step = None` to `current_step = last_db_step` and remove it here

orchestrator/workflow.py
```diff
@@ -285,6 +289,43 @@ def suspend(state: State) -> Process:
     return decorator
 
 
+# A step group is a sequence of steps that act as a single step.
```
tjeerddie, line 292: Might be good to add this as docstring to the function

tjeerddie (changes requested):
Great job, I only have some small comments

test/unit_tests/test_workflow.py
dmarjenburgh: Good point

orchestrator/services/processes.py
dmarjenburgh: I find this a bit more intuitive.

orchestrator/services/processes.py
```diff
@@ -155,54 +138,101 @@ def _db_log_step(
...
+    current_step = None
+    last_db_step = p.steps[-1] if len(p.steps) else None
+    current_sub_step, current_step_group = step_state.get("__sub_step"), step_state.get("__step_group")
+    db_sub_step, db_step_group = (
+        (last_db_step.state.get("__sub_step"), last_db_step.state.get("__step_group")) if last_db_step else (None, None)
+    )
+
+    if current_sub_step is None and db_sub_step is None:
+        pass  # Skip step group logic
+    elif (db_sub_step is None and current_sub_step) or (db_sub_step and current_sub_step is None):
```
mrijk, line 162: Looks like a xor? Maybe you can use `elif db_sub_step ^ current_sub_step:`

orchestrator/services/processes.py
```diff
@@ -155,54 +138,101 @@ def _db_log_step(
...
+            created_by=stat.current_user,
+        )
+    else:
+        # Update last step entry in db
+        last_db_step.status = process_state.status
+        last_db_step.state = step_state
+        current_step = last_db_step
+
+    if process_state.isfailed() or process_state.iswaiting():
         if (
```
mrijk, line 178: An if within an if and lots of conditions.... not very easy to ready. Maybe extract this to a small function so its clear what the if condition means.

orchestrator/utils/errors.py
```diff
@@ -108,7 +108,7 @@ def is_api_exception(ex: Exception) -> bool:
     return ex.__class__.__name__ == "ApiException"
 
 
-def error_state_to_dict(err: Exception) -> ErrorDict:
+def error_state_to_dict(err: Union[Exception, ErrorDict]) -> ErrorDict:
```
mrijk, line 111: Haven't looked at the details, but cant this Union be avoided? Maybe having two separate functions? Also the doc string is not correct/complete anymore.

test/unit_tests/test_workflow.py
```diff
@@ -453,3 +458,145 @@ def whatever():
...
+
+        result, process, step_log = run_workflow("step_group_test_workflow", init_state)
+        assert_failed(result)
+        assert extract_error(result) == "Failure Message"
+        assert 1 == len(side_effect_counter), "Side effect should be called once"
+
+        step_log = [t for t in step_log if t[0] not in [effect_step, step2, fail]]
+
+        # # Retry workflow
+        persistent = list(filter(lambda p: not (p[1].isfailed() or p[1].issuspend() or p[1].iswaiting()), step_log))
```
mrijk, line 598: In general I stop reading/trusting tests when they contain code.....

mrijk (changes requested):
LGTM, but I think code could be improved a little.

orchestrator/services/processes.py
dmarjenburgh: Oof. True but less readable I thbink

orchestrator/services/processes.py
dmarjenburgh: Yeah, I already refactored `db_log_step` into smaller functions, but I left this part the way it is. I suggest another refactor to make it more readable as a different issue.

orchestrator/utils/errors.py
dmarjenburgh: This function is used as `step_result_process.on_failed(error_state_to_dict)` and it should just work on any `Process`. This line basically transforms a `Failure[Exception]` to `Failure[ErrorDict]`.

Then we have 2 versions of `Failure` in the code, one wrapping a dictionary and one wrapping an exception.
If you called this function on a `Failure` class wrapping a dictionary, it crashes. This has never happened because to there happened to be no code path where `step_result_process.on_failed(error_state_to_dict)` would be called on a `Failure[ErrorDict]`.

This update makes that operation idempotent. But I agree the signature is ugly. Preferably we should never carry exceptions around in `Failure` classes and serialize them immediately, but that is a bigger change.

test/unit_tests/test_workflow.py
dmarjenburgh: Woops, leftover debug code

orchestrator/utils/errors.py
dmarjenburgh: I could make it a singledispatch function to split up the separate behaviours based on the error type

orchestrator/services/processes.py
dmarjenburgh: Ideally, we don't overwrite the value of local variables ever. But I think this is better done as part of a larger refactor of this module. The current_step is again used further down the line and checked against None.
Note that I copied this part from the previous implementation and added an if branch for sub_steps. I'd like to refactor the whole function, but again, better for a separate issue.

tjeerddie: Thought it was such a small change that it didn't need to wait but if you want to refactor the whole function it can wait.

tjeerddie (approved):

mrijk (approved):

orchestrator/services/processes.py
Mark90, line 162: I wonder what are the possible scenarios to reach this code.

Does `(db_sub_step is None and current_sub_step)` mean we've just entered a sub-step in the step-group and nothing was yet logged to the database?

And then does `(db_sub_step and current_sub_step is None)` mean we've resumed/continued a sub-step that was already in the database?

orchestrator/workflow.py
```diff
@@ -1157,11 +1363,18 @@ def _logstep(*x: Any) -> Process:
...
+        else:
+            step = steps.pop(0)
         return _logstep(step, process)
 
     next_state = pstat.state.resume(resume_suspend)
-
+    # Set the step_log_fn in the contextvar.
+    # This enables recursive step execution of sub steps with the same StepLogFunc.
+    # Should probably be refactored at some point as contextvars is a kind of global state.
+    step_log_fn_var.set(_logstep)
```
Mark90, line 1377: It is global state but also completely isolated, so it's okay-ish.

I do think it would be wise to clear the contextvar after the stepgroup is done; I don't think we do that at the moment?

orchestrator/services/processes.py
dmarjenburgh: Yes and no. The `current_sub_step` is the sub step as recorded in the state after the last step. If this is not present, it means we are not currently in a sub step. If `db_sub_step` is present, it means we just finished the last sub step of the step group.

```mermaid
stateDiagram-v2

    [*] --> exec_steps
    exec_steps --> a
    exec_steps --> b
    exec_steps --> f
    b --> exec_steps(2)
   exec_steps(2) --> c
   exec_steps(2) --> d
   exec_steps(2) --> e
```

- `(db_sub_step is None and current_sub_step)` happens when we just executed `c`
- `(db_sub_step and current_sub_step is None)` happens when we just executed `e`

I've been thinking on how to refactor this, but one change could be to just automatically add a sub_step to the start and end of the sub_steps chain to add/remove sub_step related data to the state. Then this logic could be simplified. The db_logstep does way too much at the moment.

PR #350, opened 2023-09-12
Add missing product_id in format_subscription of enrich_process

tjeerddie:


Mark90 (approved):

orchestrator/graphql/schemas/process.py
```diff
@@ -117,10 +116,9 @@ def last_modified(self) -> datetime:
 
     @authenticated_field(description="Returns the associated product")  # type: ignore
     def product(self) -> Optional[ProductType]:
-        subscription = first(self._original_model.subscriptions, None)  # type: ignore
         product = None
```
mrijk, line 119: Not sure what ruff advises nowadays, but maybe just:

```python
if self.product_id:
   return ProductType.from_pydantic(ProductTable.query.get(self.product_id))
return None
```

mrijk (approved):
LGTM

PR #354, opened 2023-09-18
Add callback_step functionality

### LINKED ISSUE #297: workflow step decorator to handle asynchronous communication with subsystem

dmarjenburgh:
Resolves #297

Implements a `callback_step` function which uses a step_group to handle workflows with asynchronous communication with subsystems.

- The `callback_step` is supplied with an `action_fn` and a `validate_fn`.
- Upon execution, an api endpoint is generated with a random token and added to the process state.
- The `action_fn` can then invoke the subsystem with any data (including the callback endpoint)
- The process enters  `Awaiting Callback` status
- When the endpoint is invoked, the token is matched. If it passes, the workflow continues with the `validate_fn` step. 
- If an error occurs, a retry will start from the beginning of the callback_step: creating a new endpoint and invoking the action_fn.

### DIFF
--- a/orchestrator/api/api_v1/endpoints/processes.py
+++ b/orchestrator/api/api_v1/endpoints/processes.py
@@ -154,7 +154,7 @@ def resume_process_endpoint(
 
 
 @router.post(
-    "/{process_id}/continue/{token}",
+    "/{process_id}/callback/{token}",
     response_model=None,
     status_code=HTTPStatus.OK,
     dependencies=[Depends(check_global_lock, use_cache=False)],
--- a/orchestrator/api/api_v1/endpoints/processes.py
+++ b/orchestrator/api/api_v1/endpoints/processes.py
@@ -13,6 +13,7 @@
 
 """Module that implements process related API endpoints."""
 
+import json
 import struct
 import zlib
 from http import HTTPStatus
@@ -56,6 +57,7 @@ from orchestrator.services.processes import (
     _get_process,
     abort_process,
     api_broadcast_process_data,
+    continue_awaiting_process,
     load_process,
     resume_process,
     start_process,
@@ -149,6 +151,31 @@ def resume_process_endpoint(
 
     broadcast_func = api_broadcast_process_data(request)
     resume_process(process, user=user, user_inputs=json_data, broadcast_func=broadcast_func)
+
+
+@router.post(
+    "/{process_id}/continue/{token}",
+    response_model=None,
+    status_code=HTTPStatus.OK,
+    dependencies=[Depends(check_global_lock, use_cache=False)],
+)
+def continue_awaiting_process_endpoint(
+    process_id: UUID,
+    token: str,
+    request: Request,
+    json_data: JSON = Body(...),
+) -> None:
+    check_global_lock()
+
+    process = _get_process(process_id)
+
+    if process.last_status != ProcessStatus.AWAITING_CALLBACK:
+        raise_status(HTTPStatus.CONFLICT, "This process is not in an awaiting state.")
+
+    try:
+        continue_awaiting_process(process, token=token, input_data=json.loads(json_data))
+    except AssertionError as e:
+        raise_status(HTTPStatus.NOT_FOUND, str(e))
 
 
 @router.put(
--- a/orchestrator/services/processes.py
+++ b/orchestrator/services/processes.py
@@ -563,11 +563,9 @@ def continue_awaiting_process(
 
     # We need to pass the callback data to the worker executor. Currently, this is not supported.
     # Therefore, we update the step state in the db and kick-off resume_workflow
-    # TODO: Allow passing additional data to be merged to the state upon resume_workflow
-    if "__callback_result_key" in state:
-        state = {**state, state["__callback_result_key"]: input_data}
-    else:
-        state = {**state, **input_data}
+    # Possible improvement: Allow passing additional data to be merged to the state upon resume_workflow
+    result_key = state.get("__callback_result_key", "callback_result")
+    state = {**state, result_key: input_data}
 
     current_step = process.steps[-1]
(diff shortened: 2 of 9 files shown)


orchestrator/services/processes.py
```diff
@@ -526,6 +535,50 @@ def resume_process(
...
+    state = pstat.state.unwrap()
+
+    # Check if the token matches
+    token_from_state = state.get("__callback_token")
+    if token != token_from_state:
+        raise AssertionError("Invalid token")
+
+    # We need to pass the callback data to the worker executor. Currently, this is not supported.
+    # Therefore, we update the step state in the db and kick-off resume_workflow
+    # TODO: Allow passing additional data to be merged to the state upon resume_workflow
```
tjeerddie, line 566: Remove todo, looks done.
```suggestion
```

tjeerddie (approved):

orchestrator/services/processes.py
dmarjenburgh: It's not really done. The current way is an ugly workaround. We should pass the callback data to the resume_workflow task like it is done with user_inputs. Currently, resume_workflow only supports user_inputs and no additional data.

tjeerddie: Then fix it here or create ticket and add the ticket number to the todo. so we don't keep random todo's in the project

orchestrator/api/api_v1/endpoints/processes.py
```diff
@@ -122,6 +124,26 @@ def resume_process_endpoint(
     resume_process(process, user=user_name, user_inputs=json_data, broadcast_func=broadcast_func)
 
 
+@router.post("/{process_id}/continue/{token}", response_model=None, status_code=HTTPStatus.OK)
```
Mark90, line 127: What about just calling it `/{process_id}/callback/{token}`?

Mark90: But I don't have a problem with the endpoint name as it is. "Resume suspended" versus "Continue awaiting" is distinct enough

dmarjenburgh: callback in the url sounds good

orchestrator/workflow.py
```diff
@@ -323,10 +331,55 @@ def dblogstep(step_: Step, p: Process) -> Process:
...
     # Make sure we return a form is a sub step has a form
-    form = next((sub_step.form for sub_step in steps if sub_step.form), None)
+    form = next((sub_step.form for sub_step in steps if sub_step.form), None) if extract_form else None
     return make_step_function(func, name, form)
 
 
+def _create_endpoint_step(key: str = DEFAULT_CALLBACK_ROUTE_KEY) -> StepFunc:
+    def stepfunc(process_id: UUID) -> State:
+        token = secrets.token_urlsafe()
+        route = f"/processes/{process_id}/continue/{token}"
```
Mark90, line 341: What's the route used for?

test/unit_tests/workflows/test_async_workflow.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @workflow("Test wf", target="Target.CREATE")
+    def testwf():
+        return begin >> step1 >> call_ext_system >> step2 >> done
+
+    with WorkflowInstanceForTests(testwf, "testwf"):
+        result, process, step_log = run_workflow("testwf", {})
+        assert_awaiting_callback(result)
+        state = result.unwrap()
+        assert "callback_route" in state
```
Mark90, line 43: You could use the DEFAULT_CALLBACK_ROUTE constant here

orchestrator/services/processes.py
```diff
@@ -526,6 +535,50 @@ def resume_process(
...
+    if "__callback_result_key" in state:
+        state = {**state, state["__callback_result_key"]: input_data}
+    else:
+        state = {**state, **input_data}
```
Mark90, line 570: What happens when we have a callback step without a result key, and the other system (accidentally) sends back an array instead of a dict?

Mark90 (approved):

orchestrator/services/processes.py
Mark90: I think the callback result should always be stored in a key to avoid TypeErrors (or verify that the json data is an object/dict)

orchestrator/workflow.py
dmarjenburgh: That's the part of the endpoint used for the callback.
It's a route (or path) and not a full endpoint url, because we are missing the BASE_API_URL, but that can be added in the call to the external system.
It is added to the state so you can access it in the action step. You need to communicate this endpoint to the other system.

orchestrator/services/processes.py
dmarjenburgh: :boom: 

In that case, what should be the preferred action? Maybe always add it under a default result key.

Mark90: Yeah, agree

Issue #360, opened 2023-09-25
unnecessary json.loads in continue_awaiting_process_endpoint


torkashvand:
Hi considering this code

```python 
@router.post(
    "/{process_id}/callback/{token}",
    response_model=None,
    status_code=HTTPStatus.OK,
    dependencies=[Depends(check_global_lock, use_cache=False)],
)
def continue_awaiting_process_endpoint(
    process_id: UUID,
    token: str,
    request: Request,
    json_data: JSON = Body(...),
) -> None:
    check_global_lock()

    process = _get_process(process_id)

    if process.last_status != ProcessStatus.AWAITING_CALLBACK:
        raise_status(HTTPStatus.CONFLICT, "This process is not in an awaiting state.")

    try:
        continue_awaiting_process(process, token=token, input_data=json.loads(json_data))
    except AssertionError as e:
        raise_status(HTTPStatus.NOT_FOUND, str(e))
```
        

The `json_data: JSON = Body(...)`  line in this function argument already instructs FastAPI to take the JSON content from the request body and convert it into the appropriate Python type. Therefore, you can use json_data directly in your function as a Python object without the need for `json.loads.`
     
I wanted to create a PR but I couldn't. I might do something wrong!

tjeerddie:
good spot and you should be able to create a PR from a fork of the project

torkashvand:
Created a [PR](https://github.com/workfloworchestrator/orchestrator-core/pull/361)

pboers1988:
PR merged

PR #361, opened 2023-09-25
remove unnecessary json loads in continue_awaiting_process_endpoint

torkashvand:
Copied from [this issue](https://github.com/workfloworchestrator/orchestrator-core/issues/360)

Hi considering this code

```python 
@router.post(
    "/{process_id}/callback/{token}",
    response_model=None,
    status_code=HTTPStatus.OK,
    dependencies=[Depends(check_global_lock, use_cache=False)],
)
def continue_awaiting_process_endpoint(
    process_id: UUID,
    token: str,
    request: Request,
    json_data: JSON = Body(...),
) -> None:
    check_global_lock()

    process = _get_process(process_id)

    if process.last_status != ProcessStatus.AWAITING_CALLBACK:
        raise_status(HTTPStatus.CONFLICT, "This process is not in an awaiting state.")

    try:
        continue_awaiting_process(process, token=token, input_data=json.loads(json_data))
    except AssertionError as e:
        raise_status(HTTPStatus.NOT_FOUND, str(e))
```
        

The `json_data: JSON = Body(...)`  line in this function argument already instructs FastAPI to take the JSON content from the request body and convert it into the appropriate Python type. Therefore, you can use json_data directly in your function as a Python object without the need for `json.loads.`

### DIFF
--- a/orchestrator/api/api_v1/endpoints/processes.py
+++ b/orchestrator/api/api_v1/endpoints/processes.py
@@ -13,11 +13,10 @@
 
 """Module that implements process related API endpoints."""
 
-import json
 import struct
 import zlib
 from http import HTTPStatus
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union, cast
 from uuid import UUID
 
 import structlog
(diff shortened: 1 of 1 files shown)


tjeerddie (approved):
LGTM, need to check if someone can start the pipelines

pboers1988 (approved):

dmarjenburgh (approved):
Right, FastAPI converts it if you supply the `Content-Type: application.json` which I omitted when testing.

It should work without the cast, but this is fine as well

PR #363, opened 2023-09-27
Feature/searchable metadata

dmarjenburgh:
- Add migration to include subscription_metadata to subscription search view
- Return subscription metadata in subscriptions search endpoint
- Refactor code to use  `sqlalchemy.select`

### DIFF
--- a/orchestrator/api/api_v1/endpoints/subscriptions.py
+++ b/orchestrator/api/api_v1/endpoints/subscriptions.py
@@ -215,10 +215,10 @@ def subscriptions_filterable(
         List of subscriptions
 
     """
-    _range: Optional[list[int]] = list(map(int, range.split(","))) if range else None
-    _sort: Optional[list[str]] = sort.split(",") if sort else None
-    _filter: Optional[list[str]] = filter.split(",") if filter else None
-    logger.info("subscriptions_filterable() called", range=_range, sort=_sort, filter=_filter)
+    range_ = list(map(int, range.split(","))) if range else None
+    sort_ = sort.split(",") if sort else None
+    filter_ = filter.split(",") if filter else None
+    logger.info("subscriptions_filterable() called", range=range_, sort=sort_, filter=filter_)
     stmt = select(SubscriptionTable, SubscriptionMetadataTable.metadata_).join_from(
         SubscriptionTable, SubscriptionMetadataTable, isouter=True
     )
@@ -226,8 +226,8 @@ def subscriptions_filterable(
     stmt = stmt.join(SubscriptionTable.product).options(
         contains_eager(SubscriptionTable.product), defer(SubscriptionTable.product_id)
     )
-    stmt = query_with_filters(stmt, _sort, _filter)
-    stmt = add_response_range(stmt, _range, response)
+    stmt = query_with_filters(stmt, sort_, filter_)
+    stmt = add_response_range(stmt, range_, response)
 
     sequence = db.session.execute(stmt).all()
     return [{**s.__dict__, "metadata": md} for s, md in sequence]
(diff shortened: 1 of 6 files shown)


orchestrator/api/api_v1/endpoints/subscriptions.py
```diff
@@ -197,10 +199,29 @@ def depends_on_subscriptions(
     return query_depends_on_subscriptions(subscription_id, filter_statuses).all()
 
 
-@router.get("/", response_model=List[SubscriptionSchema])
+def _add_response_range(stmt: Select, range_: Optional[list[int]], response: Response) -> Select:
```
tjeerddie, line 202: there is a `apply_range_to_query` in `orchestrator.db.range`

tjeerddie (approved):

orchestrator/api/api_v1/endpoints/subscriptions.py
```diff
@@ -213,14 +215,22 @@ def subscriptions_filterable(
         List of subscriptions
 
     """
-    _range: Union[List[int], None] = list(map(int, range.split(","))) if range else None
-    _sort: Union[List[str], None] = sort.split(",") if sort else None
-    _filter: Union[List[str], None] = filter.split(",") if filter else None
+    _range: Optional[list[int]] = list(map(int, range.split(","))) if range else None
+    _sort: Optional[list[str]] = sort.split(",") if sort else None
```
mrijk, line 219: Just curious if you need the type declarations here.

orchestrator/api/api_v1/endpoints/subscriptions.py
```diff
@@ -213,14 +215,22 @@ def subscriptions_filterable(
...
+    _filter: Optional[list[str]] = filter.split(",") if filter else None
```
mrijk, line 220: Maybe use filter_ here instead, since _filter suggests that the var isn't used

orchestrator/api/helpers.py
```diff
@@ -73,86 +73,82 @@ def _add_sort_to_query(query: Query, sort: Optional[List[str]]) -> Query:
...
-    response: Response,
-    query: Query,
-    range: Optional[List[int]] = None,
+def query_with_filters(  # noqa: C901
+    stmt: Select,
     sort: Optional[List[str]] = None,
     filters: Optional[List[str]] = None,
-) -> List:
+) -> Select:
     if filters is not None:
```
mrijk, line 81: In the future we may want to decomplect this code a bit. Huge for loop with nested ifs...

orchestrator/api/helpers.py
```diff
@@ -73,86 +73,82 @@ def _add_sort_to_query(query: Query, sort: Optional[List[str]]) -> Query:
...
 
-    query = _add_sort_to_query(query, sort)
+    return _add_sort_to_query(stmt, sort)
 
-    if range is not None and len(range) == 2:
+
+def add_response_range(stmt: Selectable, range_: Optional[list[int]], response: Response) -> Selectable:
+    if range_ is not None and len(range_) == 2:
+        total = db.session.scalar(select(func.count()).select_from(stmt.subquery()))
+        range_start = int(range_[0])
```
mrijk, line 142: Is this cast to int needed, since it's a list of int?

If not, you could do:

range_start, range_end = range_

orchestrator/api/api_v1/endpoints/subscriptions.py
dmarjenburgh: Nope. Wasn't necessary

orchestrator/api/api_v1/endpoints/subscriptions.py
dmarjenburgh: Check

orchestrator/api/helpers.py
dmarjenburgh: Yeah, it's already better than a month ago, but can definitely be simplified further

mrijk (approved):
LGTM

PR #385, opened 2023-10-10
Update SQLAlchemy queries to use v2 instead of v1 for graphql resolvers

tjeerddie:
`orchestrator.db.sorting` and `orchestrator.db.filters` are also updated.

related: #386

### DIFF
--- a/orchestrator/graphql/resolvers/process.py
+++ b/orchestrator/graphql/resolvers/process.py
@@ -15,12 +15,13 @@ from typing import Union
 
 import structlog
 from pydantic.utils import to_lower_camel
-from sqlalchemy.orm import defer, joinedload
+from sqlalchemy import func, select
+from sqlalchemy.orm import defer, selectinload
 
-from orchestrator.db import ProcessSubscriptionTable, ProcessTable, SubscriptionTable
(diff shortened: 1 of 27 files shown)


mrijk (approved):
LGTM

orchestrator/graphql/resolvers/process.py
```diff
@@ -62,19 +63,19 @@ async def resolve_processes(
...
     )
 
-    query = filter_processes(query, pydantic_filter_by, _error_handler)
-    query = sort_processes(query, pydantic_sort_by, _error_handler)
-    total = query.count()
-    query = apply_range_to_query(query, after, first)
+    stmt = filter_processes(stmt, pydantic_filter_by, _error_handler)
+    stmt = sort_processes(stmt, pydantic_sort_by, _error_handler)
+    total = db.session.scalar(select(func.count()).select_from(stmt.subquery()))
+    stmt = apply_range_to_statement(stmt, after, after + first + 1)
```
Mark90, line 76: Why did the `limit` parameter change from `first` to `after + first + 1`?

Mark90: Nevermind, I see it's a different function as well.

Mark90 (approved):

PR #387, opened 2023-10-10
Various generator CLI improvements

hanstrompert:
- replace typer.echo() with proper log messsages
- exit when product configuration cannot be read
- change type of all path variables from str to Path
- add 'variable' (snake-cased product name) to config
- make --config-file option required
- honour DATABASE_URI and PYTHONPATH environment variables while generating migration
- correct location of database migration files
- fix the generation of translations
- fix the generation of lazy workflow definitions
- remove superflous whitespace from templates
- fix migration template
- import missing Optional in product template
- add variable to product blocks config
- add option to specify custom templates folder for generate workflow commando
- extend Jinja filesystem loader search path with custom folder
- create missing dunder init files
- add product_types_module to workflow renderers
- refactor create and modify workflow templates to leverage new custom templating capabilities
- while failing to read product config raise exception instead of exit()
- create migration head 'data' if it does not exist
- add products for all fixed input combinations to migration
- add all new products to subscription model registry
- close workflow file before calling writer()
- fix typos in subscription descriptions
- use begin instead of StepList()
- make input fields Optional if not required for that lifecycle state
- make input fields ReadOnlyField's if not modifiable
- fix modify workflow summary page
- refactor terminate workflow template to include custom templates


orchestrator/cli/generate.py
```diff
@@ -11,56 +11,67 @@
...
             except yaml.YAMLError:
-                typer.echo("Failed to parse configuration file.")
+                logger.error("failed to parse configuration file", config_file=str(config_file))
     except FileNotFoundError:
-        typer.echo(f'File "{config_file}" not found')
+        logger.error("configuration file not found", config_file=str(config_file))
 
-    return None
+    # when we cannot get the product configuration somehow, it does not make sense to continue ...
+    sys.exit(1)
```
Mark90, line 47: True!

It would be better to raise a (perhaps custom) exception though, as sys.exit will attempt to exit the current interpreter, which may not always be what you want.

orchestrator/cli/generator/generator/migration.py
```diff
@@ -12,21 +12,28 @@
...
+        environ.update({"PYTHONPATH": "."})
+    logger.info(
+        "creating new db revision", database_uri=environ.get("DATABASE_URI"), pythonpath=environ.get("PYTHONPATH")
+    )
+    create_migration_command = f'python main.py db revision --message "Add {config["name"]}" --head=data@head'
     result = subprocess.check_output(create_migration_command, shell=True, text=True)  # noqa: S602
 
     for line in result.splitlines():
-        if m := re.search("migrations/versions/general/([^ ]+)", line):
+        if m := re.search("migrations/versions/schema/([^ ]+)", line):
```
Mark90, line 36: Hmm, in surf orchestrator the path is `migrations/versions/general/`, in core it is indeed `migrations/versions/schema/` .. I think we want the former ? 🤔 

I think we've had a discussion about making this configurable, not sure though. Maybe @tjeerddie remembers?

orchestrator/cli/generator/generator/workflow.py
```diff
@@ -55,10 +55,18 @@ def add_optional_nso_config(config: dict) -> dict:
...
 
     path = settings.FOLDER_PREFIX / settings.WORKFLOWS_PATH / Path("__init__.py")
-    writer(path, content, append=True)
+    if not path.exists():
+        writer(path, "from orchestrator.workflows import LazyWorkflowInstance\n\n")
+    with open(path, "r") as fp:
+        if f"workflows.{variable}.create_{variable}" in fp.read():
+            logger.warning("not re-adding lazy workflows", product=variable)
+        else:
+            writer(path, content, append=True)
```
Mark90, line 69: This will probably work but it's a bit cleaner to close the read handle before writing

Mark90 (commented):
Nice work :) I see this is still a draft, just leaving a few early review comments

orchestrator/cli/generate.py
hanstrompert: Good point. I removed the exit(1) and now just re-raise the exception after logging the problem.

orchestrator/cli/generator/generator/migration.py
hanstrompert: Not sure why different version locations are being used, @tjeerddie do you know why this is?

I can add a field to the product yaml to indicate the version location. I would prefer to keep the default  migrations/versions/schema/ as this is what most user use.

orchestrator/cli/generator/generator/workflow.py
hanstrompert: It doesn't hurt to have the file open for read and pointing to end of file, but I agree that it would be better to close the file when we do not want to read from it anymore, if only to release an open file slot. Added a close() before calling the writer().

PR #389, opened 2023-10-12
Callback wf improvements

dmarjenburgh:
A refactor to simplify some workings of dblogstep and accompanying changes to step_group

- `_db_log_step` which handles persisting `process` and `process_steps` in the db now looks at certain internal keys in the state to customize this handling. It is now step_group agnostic.
- The Step group `step_list` is now extended with and `enter` and `exit` step which handles initialization and cleanup.
- Added a multistep unit_test for an async workflow with persistence.

### DIFF
--- a/orchestrator/services/processes.py
+++ b/orchestrator/services/processes.py
@@ -16,7 +16,7 @@ import threading
 from concurrent.futures.thread import ThreadPoolExecutor
 from functools import partial
 from http import HTTPStatus
-from typing import Any, Callable, Dict, Iterable, List, Literal, Optional, Tuple
+from typing import Any, Callable, Dict, List, Literal, Optional, Tuple
 from uuid import UUID, uuid4
 
 import structlog
--- a/orchestrator/services/processes.py
+++ b/orchestrator/services/processes.py
@@ -16,7 +16,7 @@ import threading
 from concurrent.futures.thread import ThreadPoolExecutor
 from functools import partial
 from http import HTTPStatus
-from typing import Any, Callable, Dict, List, Literal, Optional, Tuple
+from typing import Any, Callable, Dict, Iterable, List, Literal, Optional, Tuple
 from uuid import UUID, uuid4
 
 import structlog
@@ -43,6 +43,7 @@ from orchestrator.websocket import (
 )
 from orchestrator.websocket.websocket_manager import WebSocketManager
 from orchestrator.workflow import (
+    CALLBACK_TOKEN_KEY,
     Failed,
     ProcessStat,
     ProcessStatus,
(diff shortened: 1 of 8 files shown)


orchestrator/services/processes.py
```diff
@@ -158,31 +159,30 @@ def _update_process(process_id: UUID, step: Step, process_state: WFProcess) -> P
...
     stat: ProcessStat, p: ProcessTable, step: Step, process_state: WFProcess
 ) -> ProcessStepTable:
-    """Checks if last error state is identical to determine if we add a new step or update the last one."""
+    """Checks if last error state is identical to determine if we add a new step or update the last one.
+
+    There are some special internal keys checked to customise how the step is updated in the database.
+    These keys are only used for this purpose and not stored in the database.
+    - '__step_name_override' - When this key is present, the value will be used as the ProcessStepTable.name
+    - '__replace_last_state' - When the value for this key is truthy, the previous state in the db will be overridden
+    - '__remove_keys' - Removes the keys in this given list from the state
```
Mark90, line 168: Maybe it's an idea to group all of these keys in the state, in a key called '__runtime' or whatever, and then shape it through a pydantic model?

```python
class RuntimeVariables(BaseModel):
    step_name_override: str | None = Field(None, description="When this key is present, the value will be used as ProcessStepTable.name")
    replace_last_state: bool | None = Field(None, description="...")
    remove_keys: list[str] | None = Field(None, description="...")


# serialize
variables = RuntimeVariables(step_name_override="my_step", remove_keys=["__step_group"])
step_state['__runtime'] = variables.dict() 


# deserialize
runtime = RuntimeVariables(**step_state.get("__runtime", {}))
```

I think it can save a bit of type checking, validation and dict manipulation spreading through the codebase

orchestrator/workflow.py
```diff
@@ -292,6 +292,19 @@ def suspend(state: State) -> Process:
...
+    def add_sub_group_info_to_state() -> State:
+        return {"__step_name_override": name, "__step_group": name}
+
+    def remove_sub_group_info_from_state() -> State:
+        return {"__remove_keys": ["__step_group", "__sub_step"]}
+
+    enter_step = begin >> step(f"{name} - Enter")(add_sub_group_info_to_state)
+    exit_step = step(f"{name} - Exit")(remove_sub_group_info_from_state)
+
+    return begin >> enter_step >> steps >> exit_step
```
Mark90, line 305: `begin` is done twice?

test/unit_tests/workflows/test_async_workflow.py
```diff
@@ -83,3 +87,95 @@ def testwf():
...
+        response = test_client.get(f"api/processes/{process_id}")
+        response_data = response.json()
+
+        assert response_data["status"] == "completed"
+
+        state = response_data["current_state"]
+        assert state["ext_data"] == "56789"
+
+        assert state["phase"] is None
+
```
Mark90, line 180: Nice test!

Do you think it's worth it to add a similar one for the sad flow? (unless this is already present somewhere)

Mark90 (approved):

test/unit_tests/workflows/test_async_workflow.py
dmarjenburgh: After running into a nasty issue with using database scoping in tests I've decided to postpone this.

There are fortunately unit_tests in `test_workflow.py` for step_groups which do test failure modes. They use the `run_workflow` test utility and bypass the database scoping problem.

orchestrator/services/processes.py
```diff
@@ -158,31 +159,30 @@ def _update_process(process_id: UUID, step: Step, process_state: WFProcess) -> P
...
         current_step = last_db_step
+        current_step.status = process_state.status
+        current_step.state = step_state
+
+    if "__remove_keys" in step_state:
+        keys_to_remove = step_state.get("__remove_keys")
+        keys_to_remove = keys_to_remove if isinstance(keys_to_remove, Iterable) else []
+        for k in keys_to_remove:
+            step_state.pop(k, None)
+        step_state.pop("__remove_keys", None)
```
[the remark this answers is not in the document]
dmarjenburgh: Done

orchestrator/services/processes.py
```diff
@@ -158,31 +159,37 @@ def _update_process(process_id: UUID, step: Step, process_state: WFProcess) -> P
...
         current_step = last_db_step
+        current_step.status = process_state.status
+        current_step.state = step_state
+
+    # Core internal: __remove_keys
+    try:
+        keys_to_remove = step_state.get("__remove_keys", [])
+        keys_to_remove = keys_to_remove if isinstance(keys_to_remove, Iterable) else []
+        for k in keys_to_remove:
+            step_state.pop(k, None)
```
mrijk, line 188: Personally I would prefer creating a new state (with the keys omitted) instead of .pop, but maybe thats to much overhead?

orchestrator/services/processes.py
```diff
@@ -158,31 +159,37 @@ def _update_process(process_id: UUID, step: Step, process_state: WFProcess) -> P
...
+    except TypeError:
+        logger.error("Value for '__keys_to_remove' is not iterable.")
```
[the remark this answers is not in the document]
dmarjenburgh: Indeed, needed to remove the instance check

orchestrator/workflow.py
```diff
@@ -307,28 +320,29 @@ def step_group(name: str, steps: StepList, extract_form: bool = True) -> Step:
...
-            return p
-
         # If sub_step information is present in the state. Resume from the next sub step
         if "__sub_step" in initial_state:
             step_list = StepList(dropwhile(lambda s: s.name != initial_state.get("__sub_step"), steps))[1:]
         else:
             step_list = steps
 
+        def dblogstep(step_: Step, p: Process) -> Process:
+            p = p.map(lambda s: s | {"__sub_step": step_.name, "__step_name_override": name})
```
mrijk, line 335: Can we have different names instead of reusing/overwriting p? Makes debugging easier if something goes wrong here.

test/unit_tests/test_workflow.py
```diff
@@ -419,7 +420,14 @@ def inc_n(n=0):
 
 def store(log) -> StepLogFunc:
     def _store(_: ProcessStat, step_: Step, process: Process):
-        log.append((step_.name, process))
+        state = process.unwrap()
+        step_name = state.pop("__step_name_override", step_.name)
+        for k in state.get("__remove_keys", []) + ["__remove_keys"]:
```
mrijk, line 425: Maybe assign `state.get("__remove_keys", []) + ["__remove_keys"]` to a variable

test/unit_tests/workflows/__init__.py
```diff
@@ -156,13 +156,21 @@ def __repr__(self) -> str:
...
+
+        state = process.unwrap()
+        state.pop("__step_name_override", None)
+        for k in state.get("__remove_keys", []) + ["__remove_keys"]:
+            state.pop(k, None)
+        if state.pop("__replace_last_state", None):
+            step_log[-1] = (step, process)
+        else:
+            step_log.append((step, process))
+        return process
```
mrijk, line 173: This looks very similar to the actual code, and I assume we have to keep it carefully in sync? Would it make sense to extract this as a function, so it can be reused here?

mrijk (approved):
LGTM, couple of minor remarks.

PR #390, opened 2023-10-13
Replace update_ticket_step with additional_steps

hanstrompert:
The optional update_ticket_step argument of the create_workflow, modify_workflow and terminate_workflow decorators is implemented in a non-optional way. This is resolved by replacing it with an additional_steps argument that takes a StepList instead of a Step. Besides now being able to add more than one additional step, as will be used by the Generate Workflows CLI command, the step list can also be empty.

Note that this will be a breaking change for those who use update_ticket_step in their custom orchestrator.


orchestrator/workflows/utils.py
```diff
@@ -186,7 +186,7 @@ def create_workflow(
     description: str,
     initial_input_form: Optional[InputStepFunc] = None,
     status: SubscriptionLifecycle = SubscriptionLifecycle.ACTIVE,
-    update_ticket_step: Optional[Step] = None,
+    additional_steps: StepList = begin,
```
Mark90, line 189: Looks a bit odd for `begin` to be inserted in the middle of the workflow. Perhaps just default to `StepList()`?

Mark90 (approved):
We probably only use this ourselves for conditionally running a Jira Ticket update step if there's `ticket_id` in the state.
So in this case it's ok to skip the deprecation warning.

We do need to adapt our Surf code for this.

orchestrator/workflows/utils.py
hanstrompert: That is what I initially had 😃, but then ruff started to complain: `B008 Do not perform function call 'StepList' in argument defaults`, but I can off course ignore that warning.

Mark90: Yeah, ruff is right, and I'm wrong. There shouldn't be a list in the argument defaults.

Better to default to `None` and evaluate in the function body with `additional_steps = additional_steps or StepList()`

hanstrompert:
> We probably only use this ourselves for conditionally running a Jira Ticket update step if there's `ticket_id` in the state. So in this case it's ok to skip the deprecation warning.

It can also be used this to add an send_email() step for example.
 
> We do need to adapt our Surf code for this.

There already is a merge request open to adapt the SURF code.

orchestrator/workflows/utils.py
hanstrompert: Applied suggestion.

pboers1988 (approved):
LGTM, after conflict fix

hanstrompert:
Conflicts were caused by premature merge of part of this pull request. Conflicts are solved.

PR #392, opened 2023-10-17
Generator improvements

hanstrompert:
- replace typer.echo() with proper log messsages
- exit when product configuration cannot be read
- change type of all path variables from str to Path
- add 'variable' (snake-cased product name) to config
- make --config-file option required
- honour DATABASE_URI and PYTHONPATH environment variables while generating migration
- correct location of database migration files
- fix the generation of translations
- fix the generation of lazy workflow definitions
- remove superflous whitespace from templates
- fix migration template
- import missing Optional in product template
- add variable to product blocks config
- add option to specify custom templates folder for generate workflow commando
- extent Jinja filesystem loader search path with custom folder
- create missing dunder init files
- add product_types_module to workflow renderers
- refactor create and modify workflow templates to leverage new custom
  templating capabilities
- replace update_ticket_step with addtional_steps
- while failing to read product config raise exception instead of exit()
- create migration head 'data' if it does not exist
- add products for all fixed input combinations to migration
- add all new products to subscription model registry
- close workflow file before calling writer()
- fix typos in subscription descriptions
- use begin instead of StepList()
- make input fields Optional if not required for that lifecycle state
- make input fields ReadOnlyField's if not modifiable
- fix modify workflow summary page
- refactor terminate workflow template to include custom templates


orchestrator/cli/generator/generator/migration.py
```diff
@@ -10,28 +10,67 @@
...
     else:
         return None
 
 
+def get_heads() -> dict:
+    get_heads_command = "python main.py db heads"
+    result = subprocess.check_output(get_heads_command, shell=True, text=True)  # noqa: S602
+
+    heads = {}
+    for line in result.splitlines():
```
mrijk, line 54: Can probably be rewritten a bit nicer with set comprehension:

Semi working code:

```python
def is_revision(line):
   if match := re.search("^([^ ]+) \\(([^ ]+)\\)", line):
       return match.groups()
   return None

revisions = (revision for line in result.splitlines() if revision := is_revision(line))
return dict(revisions)
```

orchestrator/cli/generator/generator/migration.py
```diff
@@ -45,26 +84,69 @@ def process() -> Generator:
...
 
+    if migration_file := create_migration_file(message=f"add {config['name']}", head="data"):
+        if revision_info := get_revision_info(migration_file):
+            if "fixed_inputs" in config and config["fixed_inputs"]:
+                fixed_input_values = [
+                    [(fi["name"], str(value)) for value in fi["values"]] for fi in config["fixed_inputs"]
+                ]
+                fixed_input_combinations = list(itertools.product(*fixed_input_values))
+                product_variants = [
+                    (" ".join([config["name"]] + [field[1] for field in combination]), combination)
```
mrijk, line 141: Maybe just extract calculating this value as a small function, to avoid looking at nested for loops.

mrijk (approved):
LGTM

Issue #375, opened 2023-10-03
Add stateDelta to the process grapgh endpoint in orchestrator core


wouter1975:
(SURF shadow issue https://git.ia.surfsara.nl/netdev/automation/projects/orchestrator/-/issues/1799)

## Summary
For the detail process page each step shows the state delta with the previous step. In current V1 and V2 GUI this delta is calculated in the front end. This requires some code logic, which could well be done in the backend. The expectation is that adding delta calculation in the backend will also improve the UX, as the page will (most) probably load quicker.
How to calculate the delta:

* perform deep diff on state (json)
* return only added state content or changed values
* don't return removed content from state
* don't filter any of hidden keys, like objects starting with LABEL, DIVIDER, __OLD_SUBSCRIPTIONS __

stateDelta for no changes
```
{}
```

stateDelta with changes added to state of certain step
```
{
   "vc": {
      "speed_policer": false
   }
}
```

### Testing

compare the BE implementation with the current V1 implementation.

PR #401, opened 2023-10-24
Add stateDelta to process step in graphql

tjeerddie:
`deepdiff` is good to see the differences but not for getting the new and updated values and you would need to parse that data. There is a [dictdiffer](https://pypi.org/project/dictdiffer/) package that gives it in a better format which seems much easier to parse into new and updated values.

I kept this simple to get the same step state delta as frontend V1 creates but it's much faster.

The delta of one of our surf processes took frontend 30sec and backend 0,248 seconds

Closes: #375

### DIFF
--- a/orchestrator/api/api_v1/endpoints/workflows.py
+++ b/orchestrator/api/api_v1/endpoints/workflows.py
@@ -18,17 +18,40 @@ from typing import List, Optional
 from fastapi.routing import APIRouter
 
 from orchestrator.db import ProductTable, WorkflowTable
-from orchestrator.schemas import WorkflowSchema, WorkflowWithProductTagsSchema
+from orchestrator.schemas import StepSchema, WorkflowSchema, WorkflowWithProductTagsSchema
+from orchestrator.workflows import get_workflow
 
 router = APIRouter()
 
 
+def _add_steps_to_workflow(workflow: WorkflowTable) -> WorkflowSchema:
+    def get_steps() -> list[StepSchema]:
+        if registered_workflow := get_workflow(workflow.name):
+            return [StepSchema(name=step.name) for step in registered_workflow.steps]
+        raise AssertionError(f"Workflow {workflow.name} should be registered")
+
+    return WorkflowSchema(
+        workflow_id=workflow.workflow_id,
+        name=workflow.name,
+        target=workflow.target,
+        description=workflow.description,
+        created_at=workflow.created_at,
+        steps=get_steps(),
+    )
+
+
 @router.get("/", response_model=List[WorkflowSchema])
-def get_all(target: Optional[str] = None) -> List[WorkflowTable]:
+def get_all(target: Optional[str] = None, include_steps: bool = False) -> List[WorkflowSchema]:
     query = WorkflowTable.query
     if target:
         query = query.filter(WorkflowTable.__dict__["target"] == target)
-    return query.all()
+
+    workflows = query.all()
+
+    if include_steps:
+        return [_add_steps_to_workflow(workflow) for workflow in workflows]
+
+    return workflows
 
 
 @router.get("/with_product_tags", response_model=List[WorkflowWithProductTagsSchema])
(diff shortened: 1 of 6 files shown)


orchestrator/utils/get_updated_fields.py
```diff
@@ -0,0 +1,37 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any
+
+
+def get_dict_updates(obj1: dict, obj2: dict, delta: bool = False) -> dict:
```
mrijk, line 17: Mixed use of dict and Dict in the orchestrator-core codebase

orchestrator/utils/get_updated_fields.py
```diff
@@ -0,0 +1,37 @@
...
+            elif obj1[key] != obj2[key]:
+                updated[key] = obj2[key]
+        else:
+            updated[key] = obj2[key]
+    return updated
+
+
+def get_updated_fields(obj1: Any, obj2: Any, delta: bool = False) -> dict | Any | None:
+    if isinstance(obj1, dict) and isinstance(obj2, dict):
+        return get_dict_updates(obj1, obj2, delta)
```
mrijk, line 34: Looks like mutual recursion. I didn't check, but wasn't it possible to write in a single function?

orchestrator/utils/get_updated_fields.py
tjeerddie: yea, 3.9 -_-

orchestrator/utils/get_updated_fields.py
tjeerddie: I think it is easier to go through an entire dict with recursion for the updated props.

orchestrator/utils/enrich_process.py
```diff
@@ -11,14 +11,17 @@
...
+from more_itertools import first
+
+from orchestrator.db import ProcessStepTable, ProcessTable, SubscriptionTable
+from orchestrator.utils.get_updated_properties import get_dict_updates
+from orchestrator.workflow import ProcessStat, Step, StepStatus
 from pydantic_forms.core import generate_form
 
 
-def format_subscription(subscription: SubscriptionTable) -> dict:
+def format_subscription(subscription: SubscriptionTable) -> Dict:
```
dmarjenburgh, line 24: We require python >= 3.9. `dict` is equally powerful as `Dict` and preferred

orchestrator/utils/get_updated_properties.py
```diff
@@ -0,0 +1,37 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, Dict, Union
+
+
+def get_dict_updates(obj1: Dict, obj2: Dict) -> Dict:
```
dmarjenburgh, line 17: Use plain `dict` instead of `Dict`

orchestrator/utils/get_updated_properties.py
```diff
@@ -0,0 +1,37 @@
...
+                if nested_updates:
+                    updated[key] = nested_updates
+            elif obj1[key] != obj2[key]:
+                updated[key] = obj2[key]
+        else:
+            updated[key] = obj2[key]
+    return updated
+
+
+def get_updated_properties(obj1: Any, obj2: Any) -> Union[Dict, Any, None]:
```
dmarjenburgh, line 32: `Union` with `Any` is just `Any`

orchestrator/utils/enrich_process.py
tjeerddie: ye, thought 3.9 didn't like it....

Issue #383, opened 2023-10-10
Add in_use_by_ids to the product block instances in subscription graphql type


tjeerddie:
necessary to be able to map the inUseBySubscriptions to the product block that they are connected to. In orchestrator client V1 its used to show the in use by subscriptions in a specific product block.

`dependsOn` works fine because you can use `ownerSubscriptionId` to see that its a different subscription.

PR #404, opened 2023-10-26
Add inUseByRelations to subscription.productBlockInstances

tjeerddie:
it returns a list of `{ "subscription_id": id, "subscription_instance_id": instance_id } `. so the relations can be matched with `subscription_id` with the ids from `subscriptions.inUseBySubscriptions`.

Closes: #383

### DIFF
--- a/orchestrator/services/subscriptions.py
+++ b/orchestrator/services/subscriptions.py
@@ -580,6 +580,10 @@ def _generate_etag(model: dict) -> str:
     return md5(encoded).hexdigest()  # noqa: S303, S324
 
 
+def convert_subscription_instance(obj: Any) -> dict[str, str]:
+    return {"subscription_instance_id": str(obj.subscription_instance_id), "subscription_id": str(obj.subscription_id)}
+
+
 def build_extended_domain_model(subscription_model: SubscriptionModel) -> dict:
     """Create a subscription dict from the SubscriptionModel with additional keys."""
     customer_descriptions = SubscriptionCustomerDescriptionTable.query.filter(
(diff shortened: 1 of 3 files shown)


pboers1988 (approved):
LGTM

orchestrator/services/subscriptions.py
```diff
@@ -580,6 +580,10 @@ def _generate_etag(model: dict) -> str:
     return md5(encoded).hexdigest()  # noqa: S303, S324
 
 
+def convert_subscription_instance(obj: Any) -> dict[str, str]:
```
mrijk, line 583: convert to what? The name is very generic and could mean anything.

orchestrator/services/subscriptions.py
```diff
@@ -592,8 +596,11 @@ def build_extended_domain_model(subscription_model: SubscriptionModel) -> dict:
     def inject_in_use_by_ids(path_to_block: str) -> None:
         if not (in_use_by_subs := getattr_in(subscription_model, f"{path_to_block}.in_use_by")):
             return
-        block_instance_ids = [obj.in_use_by_id for obj in in_use_by_subs.col]
-        update_in(subscription, f"{path_to_block}.in_use_by_ids", block_instance_ids)
+
+        in_use_by_ids = [obj.in_use_by_id for obj in in_use_by_subs.col]
+        in_use_by_relations = [convert_subscription_instance(instance) for instance in in_use_by_subs]
+        update_in(subscription, f"{path_to_block}.in_use_by_ids", in_use_by_ids)
+        update_in(subscription, f"{path_to_block}.in_use_by_relations", in_use_by_relations)
```
mrijk, line 603: I hope someday we can build up an (extended) domain model from the ground using immutable datastructures instead of injecting stuff as an afterthought. Not easy to fix right now.

mrijk (approved):
LGTM

Issue #406, opened 2023-10-30
[Feature]: Enforce that `TARGET` is set for all processes


wouter1975:
### What should we build?

It is possible that processes have no target set which is undesired. In most cases these processes should be set to TARGET = SYSTEM.

The allowed Targets are:
* create
* modify
* terminate
* system

### Proposal
- [ ] Enforce the Target being set at the workflow decorator.

tjeerddie:
so target default should be `SYSTEM`

PR #407, opened 2023-10-31
Force workflow to be set with default SYSTEM

tjeerddie:
Closes: #406

### DIFF
--- a/orchestrator/workflow.py
+++ b/orchestrator/workflow.py
@@ -92,7 +92,7 @@ class Workflow(Protocol):
     name: str
     description: str
     initial_input_form: Optional[InputFormGenerator] = None
-    target: Optional[Target] = None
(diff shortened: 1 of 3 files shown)


mrijk (approved):
LGTM

orchestrator/workflow.py
```diff
@@ -92,7 +92,7 @@ class Workflow(Protocol):
     name: str
```
tjeerddie, line 92: Target is still not shown correctly on task list page, checking out why

tjeerddie: This is not shown correctly because of old incorrect database structure, created a issue to fix this: https://github.com/workfloworchestrator/orchestrator-core/issues/409

Mark90 (approved):

PR #412, opened 2023-11-03
Search parser

dmarjenburgh:
Implement parser and ts_query emitter for formalized search query syntax.

Related to #288

The query syntax introduced here allows for more flexible queries than is supported by Postgres' `websearch_to_tsquery`. A new endpoint `/subscriptions/search` has also been added.

The query syntax is a basically a subset of the ElasticSearch simple query syntax (with a few exceptions).

Features:
- Conjunction: `term1 term2` (default operator = `AND')
- Disjunction: `subquery2 | subquery2`
- Negation: `-term`
- Grouping: `(term1 term2)`
- Phrases: `"word1 word2"` matches consecutive words
- Prefix matching: `gre*` matches words starting with `gre`
- Key-value searching:
  - `field:value` matches items where field = value
  - `"field:(val1 | val2)"` matches items where `field in [val1, val2]`

### DIFF
--- a/orchestrator/api/api_v1/endpoints/subscriptions.py
+++ b/orchestrator/api/api_v1/endpoints/subscriptions.py
@@ -28,7 +28,7 @@ from starlette.responses import Response
 
 from oauth2_lib.fastapi import OIDCUserModel
 from orchestrator.api.error_handling import raise_status
-from orchestrator.api.helpers import add_response_range, query_with_filters
+from orchestrator.api.helpers import add_response_range, add_subscription_search_query_filter, query_with_filters
 from orchestrator.db import (
     ProcessStepTable,
     ProcessSubscriptionTable,
@@ -229,6 +229,38 @@ def subscriptions_filterable(
     stmt = query_with_filters(stmt, sort_, filter_)
     stmt = add_response_range(stmt, range_, response)
 
+    sequence = db.session.execute(stmt).all()
+    return [{**s.__dict__, "metadata": md} for s, md in sequence]
+
+
+@router.get("/search", response_model=list[SubscriptionWithMetadata])
+def subscriptions_search(
+    response: Response, query: str, range: Optional[str] = None, sort: Optional[str] = None
+) -> list[dict]:
+    """Get subscriptions filtered based on a search query string.
+
+    Args:
+        response: Fastapi Response object
+        query: The search query
+        range: Range
+        sort: Sort
+
+    Returns:
+        List of subscriptions
+
+    """
+    range_ = list(map(int, range.split(","))) if range else None
+    sort_ = sort.split(",") if sort else None
+    logger.info("subscriptions_search() called", range=range_, sort=sort_)
+    stmt = select(SubscriptionTable, SubscriptionMetadataTable.metadata_).join_from(
+        SubscriptionTable, SubscriptionMetadataTable, isouter=True
+    )
+
+    stmt = stmt.join(SubscriptionTable.product).options(
+        contains_eager(SubscriptionTable.product), defer(SubscriptionTable.product_id)
+    )
+    stmt = add_subscription_search_query_filter(stmt, query)
+    stmt = add_response_range(stmt, range_, response)
     sequence = db.session.execute(stmt).all()
     return [{**s.__dict__, "metadata": md} for s, md in sequence]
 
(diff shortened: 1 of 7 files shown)


orchestrator/services/subscriptions.py
```diff
@@ -115,7 +115,7 @@ def update_subscription_status(subscription_id: UUIDstr, status: str) -> Subscri
 
 
 def update_subscription_description(subscription_id: UUIDstr, description: str) -> SubscriptionTable:
-    """Update a subscriptions description.
+    """Update a subscriptions' description.
```
Mark90, line 118:
```suggestion
    """Update a subscription's description.
```

If we're going to put salt on all snails, I think it should be like this :)

orchestrator/utils/search_query.py
```diff
@@ -0,0 +1,348 @@
...
+        values = [value]
+        while next_token := self.peek():
+            if next_token[0] == Token.RPAREN:
+                break
+            if next_token[0] == Token.OP_OR:
+                self.expect(Token.OP_OR)
+                if next_value := self.parse_value():
+                    values.append(next_value)
+            else:
+                raise ParseError(f"Expected a Value or ')', got {next_token}")
```
Mark90, line 154: I don't think this supports nested groups at the moment?

orchestrator/utils/search_query.py
```diff
@@ -0,0 +1,348 @@
...
+            TSQueryVisitor.visit_term(term, acc)
+            acc.append(" & ")
+        if len(node[1]):
+            acc.pop()
+
+    @staticmethod
+    def visit_query(node: Node, acc: list[str]) -> None:
+        for expression in node[1]:
+            TSQueryVisitor.visit_and_expression(expression, acc)
+            acc.append(" | ")
```
Mark90, line 333: I understand that `AND` is the default, but is it currently possible to do a query `foo OR bar`?

orchestrator/utils/search_query.py
```diff
@@ -0,0 +1,348 @@
...
+    def visit(parse_tree: Node) -> str:
+        acc: list[str] = []
+        node_type = parse_tree[0]
+        if node_type == "Query":
+            TSQueryVisitor.visit_query(parse_tree, acc)
+        return "".join(acc)
+
+
+def create_ts_query_string(search_query: str) -> str:
+    return TSQueryVisitor.visit(Parser(Lexer(search_query).lex()).parse())
```
Mark90, line 348: I think it would be good to define an upper limit for the query. Maybe based on the number of nodes?

Mark90 (approved):
Very nice!

orchestrator/utils/search_query.py
dmarjenburgh: You mean an upper limit to the size of the query?

We would have to track the nesting level. Maybe easier to limit the number of characters in the submitted query string?

Mark90: I think something more detailed than query size is needed to stop queries `" ".join("a" for i in range(10_000))` etc. On my machine this takes 2-3 seconds to complete, and the final PG query might take even longer.

With the target users of this library there might not be any concerns of deliberate abuse but we might as well prevent it.

Maybe we can add such a constraint to the accumulator, make it wrap a list and and raise an error when appending/extending it beyond a certain size?

orchestrator/api/api_v1/endpoints/subscriptions.py
```diff
@@ -233,6 +233,38 @@ def subscriptions_filterable(
...
+    sort_ = sort.split(",") if sort else None
+    logger.info("subscriptions_search() called", range=range_, sort=sort_)
+    stmt = select(SubscriptionTable, SubscriptionMetadataTable.metadata_).join_from(
+        SubscriptionTable, SubscriptionMetadataTable, isouter=True
+    )
+
+    stmt = stmt.join(SubscriptionTable.product).options(
+        contains_eager(SubscriptionTable.product), defer(SubscriptionTable.product_id)
+    )
+    stmt = add_subscription_search_query_filter(stmt, query)
```
tjeerddie, line 262: Will there be a follow up to add this to the graphql?

tjeerddie (approved):

orchestrator/api/api_v1/endpoints/subscriptions.py
dmarjenburgh: Yes. As part of #288 which we will pick up

orchestrator/utils/search_query.py
Mark90: That seems short, but sure let's start there :) If it is an ENV setting we can change it adhoc if need be.

dmarjenburgh: I made it 512 :laughing:

PR #419, opened 2023-11-30
Search resolvers

dmarjenburgh:
Add `query` parameter to the following graphql resolvers:

- products
- product_blocks
- workflows
- resource_types
- processes

The resolver uses the implemented `SQLALchemyVisitor` which dynamically generates the SQLAlchemy statement from the parsed query.


orchestrator/cli/migrate_domain_models.py
```diff
@@ -89,7 +89,11 @@ def map_product_blocks_in_class(
         model_class._get_depends_on_product_block_types()
     )
 
-    new_blocks = {block.name: block for block in product_blocks_types_in_model if block.name not in product_blocks}
+    new_blocks = {
+        block.name: block
+        for block in product_blocks_types_in_model
+        if (block.name not in product_blocks) and (block.name is not None)
```
mrijk, line 95: Shouldn't we wonder why block.name can be None?

orchestrator/db/filters/generic_filters/bool_filter.py
```diff
@@ -13,12 +13,12 @@
...
-from sqlalchemy import Column
+from sqlalchemy import ColumnClause
 
 from orchestrator.db.filters.filters import QueryType
 
 
-def generic_bool_filter(field: Column) -> Callable[[QueryType, str], QueryType]:
+def generic_bool_filter(field: ColumnClause) -> Callable[[QueryType, str], QueryType]:
     def bool_filter(query: QueryType, value: str) -> QueryType:
         value_as_bool = value.lower() in ("yes", "y", "true", "1", "ja", "j")
```
mrijk, line 23: I vaguely remember having seen a function for this str -> bool conversion

orchestrator/cli/migrate_domain_models.py
dmarjenburgh: This is a peculiarity in how we defined the ProductBlockModelMeta class. (https://github.com/workfloworchestrator/orchestrator-core/blob/2b2d17368da40027056db20b32a2618014f608eb/orchestrator/domain/base.py#L585C14-L585C14)

The type is `str`, but we also set it to `None` so it's asking for mypy trouble. 1.7.1 complained about it so this code is backwards compatible and 1.7.1 will not complain if we do upgrade to it.

But better would be finding a less hacky solution.

orchestrator/db/filters/product.py
```diff
@@ -25,5 +37,12 @@ def product_block_filter(query: QueryType, value: str) -> QueryType:
     "product_blocks": product_block_filter,
 }
 
+PRODUCT_TABLE_COLUMN_CLAUSES: dict[str, WhereCondGenerator] = {
+    k: inferred_filter(column) for key, column in inspect(ProductTable).columns.items() for k in [key, to_camel(key)]
```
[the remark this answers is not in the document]
dmarjenburgh: It's a pretty useful feature for for-comprehensions. Would you like it if it's written like this?

```clojure
(for [[key column] (-> ProductTable inspect :columns)
      k [key (to-camel key)]]
  [k (inferred-filter column)])
```

Joking aside, making a separate function sounds good.

orchestrator/db/filters/generic_filters/bool_filter.py
dmarjenburgh: YAML?

orchestrator/cli/migrate_domain_models.py
Mark90: @mrijk It's also this `name` attribute that Pydantic 2.x warns about, because it's used as both a class variable and pydantic Field. That problem may or may not be relevant here, just mentioning it.

In practice, name will never be `None` because this only happens for the abstract classes which we don't register. But as Daniel said this is impossible for mypy to derive. There's probably a clever OOP solution for this.

orchestrator/cli/migrate_domain_models.py
```diff
@@ -89,7 +89,11 @@ def map_product_blocks_in_class(
...
+    }
```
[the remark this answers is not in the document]
Mark90: Actually I did it exactly the same here. In another place where only the name value was needed I used a helper function

```py

def get_product_block_names(pbs: list[Type[ProductBlockModel]]) -> Iterable[str]:
    return filter(None, (pb.name for pb in pbs))
```

orchestrator/db/filters/generic_filters/bool_filter.py
dmarjenburgh: I opt for the first one :sweat_smile:

orchestrator/db/filters/product_block.py
```diff
@@ -42,5 +53,11 @@ def resource_types_filter(query: QueryType, value: str) -> QueryType:
     | end_date_range_filters
 )
 
+PRODUCT_BLOCK_TABLE_COLUMN_CLAUSES = default_inferred_column_clauses(ProductBlockTable) | {
+    # "product": products_clause,
```
Mark90, line 57: Remove or uncomment line?

orchestrator/db/sorting/process.py
```diff
@@ -14,7 +14,11 @@
 
 def generic_process_relation_sort(field: Column) -> Callable[[QueryType, SortOrder], QueryType]:
     def sort_function(query: QueryType, order: SortOrder) -> QueryType:
-        joined_query = query.join(ProcessSubscriptionTable).join(SubscriptionTable).join(ProductTable)
+        if isinstance(query, Select):
+            joined_query: QueryType = query.join(ProcessSubscriptionTable).join(SubscriptionTable).join(ProductTable)
+        else:
+            # Can't join on CompoundSelect
```
Mark90, line 20: Would it make sense to log a warning in this case?

test/unit_tests/graphql/test_product.py
```diff
@@ -202,11 +207,18 @@ def test_product_has_previous_page(test_client, generic_product_1, generic_produ
...
-    data = get_product_query(
-        filter_by=[{"field": "product_blocks", "value": "PB_1-PB_3"}],
-        sort_by=[{"field": "name", "order": "ASC"}],
-    )
+@pytest.mark.parametrize(
+    "query_args",
+    [
+        {"filter_by": [{"field": "product_blocks", "value": "PB_1-PB_3"}]},
+        {"query_string": "product_block:(PB_1|PB_3)"},
+        # {"query_string": f"productBlock:PB_1 | product_block:PB_3"}, Doesn't work yet :(
```
Mark90, line 215: Do we intend to fix that? :) If not let's just remove it. Otherwise you could mark this scenarios as an xfail (expected failure) so that it at least runs https://docs.pytest.org/en/6.2.x/skipping.html#skip-xfail-with-parametrize

Mark90 (approved):
Nice!

orchestrator/db/sorting/process.py
dmarjenburgh: No, it's more an explanation why we need to check for the instance type

test/unit_tests/graphql/test_product.py
dmarjenburgh: Thanks. Should have added a TODO. It was fixed

PR #420, opened 2023-12-05
Fix Phrase in ValueGroup error

dmarjenburgh:
Renamed the incorrectly named SEMICOLON token to COLON. Regular expression used for word boundary has also been compiled at its initialization to enhance regex operation's performance. Matching test cases have been updated accordingly.


pboers1988 (approved):

mrijk (approved):
LGTM

PR #423, opened 2023-12-12
Handle words with underscores as consecutive words

dmarjenburgh:
Fixes issue with searching for terms with underscores.

Postgres will split words with underscores into separate words during `ts_tsvector` and treat it differently in `to tsquery`.

Postgres will to process the query `word_with_underscores <-> value` as `( 'word' & 'with' & 'underscores' ) <-> 'value'` instead of `word <-> with <-> underscores <-> value`. This PR provides a fix for this behavior.


pboers1988 (approved):
LGTM

tjeerddie (approved):

PR #425, opened 2023-12-14
Graphql only fetch db models when necessary

tjeerddie:
- add unit tests for is_query_detailed and is_querying_page_data.


pboers1988 (approved):
LGTM

PR #433, opened 2023-12-18
Use workflow_id on processes table

dmarjenburgh:
- Add db migration that replaces workflow column with workflow_id in processes table
- Add Soft delete workflows functionality (so processes are always tied to a workflow)
- Refactor tests tooling to handle new process/workflow relation
- Configure ProductTable model to auto-filter non-deleted workflows


orchestrator/api/helpers.py
```diff
@@ -257,7 +257,7 @@ def enrich_process(p: ProcessTable) -> _ProcessListItem:
         last_status=p.last_status,
         last_step=p.last_step,
         subscriptions=subscriptions,
-        workflow=p.workflow_name,
+        workflow=str(p.workflow_name),
```
tjeerddie, line 260: Why do you need to cast it to str?

orchestrator/migrations/versions/schema/2023-12-06_048219045729_add_workflow_id_to_processes_table.py
```diff
@@ -0,0 +1,46 @@
...
+# revision identifiers, used by Alembic.
+revision = "048219045729"
+down_revision = "da5c9f4cce1c"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(
+        "DELETE FROM processes WHERE NOT EXISTS (SELECT 1 FROM workflows wf WHERE processes.workflow = wf.name);"
```
tjeerddie, line 22: not sure if we want to delete these processes...
We need atleast a confirmation so we don't let other people delete processes that have workflow names of deleted workflows without notice

tjeerddie (approved):

orchestrator/api/helpers.py
dmarjenburgh: Mypy will see it as a Column type otherwise.

orchestrator/migrations/versions/schema/2023-12-06_048219045729_add_workflow_id_to_processes_table.py
dmarjenburgh: There isn't really a way to get confirmation in a migration. It just has to work on deployment.

A different option would be to fail the migration and have the user manually deal with the dangling processes. Not sure if that is desirable either.

pboers1988: In practice there won't be many processes that will be deleted, however maybe just to be sure we don't throw away other peoples data, we should just fail the migration with a good warning and let people solve their dangling processes.

pboers1988 (approved):

orchestrator/api/helpers.py
tjeerddie: oke

orchestrator/migrations/versions/schema/2023-12-06_048219045729_add_workflow_id_to_processes_table.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+    # Set the processes.workflow column values to the corresponding workflow_id
+    op.execute("UPDATE processes SET workflow = wf.workflow_id FROM workflows wf WHERE processes.workflow = wf.name;")
+    # Update (column_name, column_type) to (processes.workflow_id, UUID)
+    op.execute("ALTER TABLE processes RENAME COLUMN workflow TO workflow_id;")
+    op.execute("ALTER TABLE processes ALTER COLUMN workflow_id TYPE uuid USING workflow_id::uuid;")
+    op.execute(
```
tjeerddie, line 29: probably needs to be dropped in downgrade

dmarjenburgh: It will be altered to workflow again. The names are gone, they will be the workflow_ids

dmarjenburgh: Had a bug. Should work now

PR #434, opened 2023-12-20
Add settings endpoint to refresh search index

### LINKED ISSUE #432: [Feature]: Add TSV reset endpoint
### Contact Details

_No response_

### Summary

Some times the TSV might be a bit behind on the last changes in coreDB. It would be good for the end-user to have the option to reset / restart the TSV data. Adding ths function to the settings page of the FE would be a nice improvement to the user.

- [ ] Add an endpoint which can be called from the orchestratio-ui FE

### Relevant pseudo code

_No response_

dmarjenburgh:
Resolves #432

### DIFF
--- a/orchestrator/api/api_v1/endpoints/settings.py
+++ b/orchestrator/api/api_v1/endpoints/settings.py
@@ -56,8 +56,7 @@ def get_cache_names() -> dict[str, str]:
 
 @router.post("/search-index/reset")
 async def reset_search_index() -> None:
-    settings.reset_search_index()
-    return
+   return settings.reset_search_index()
 
 
 @router.put("/status", response_model=EngineSettingsSchema)
--- a/orchestrator/api/api_v1/endpoints/settings.py
+++ b/orchestrator/api/api_v1/endpoints/settings.py
@@ -56,7 +56,7 @@ def get_cache_names() -> dict[str, str]:
 
 @router.post("/search-index/reset")
 async def reset_search_index() -> None:
-   return settings.reset_search_index()
+    return settings.reset_search_index()
 
 
 @router.put("/status", response_model=EngineSettingsSchema)
--- a/orchestrator/api/api_v1/endpoints/settings.py
+++ b/orchestrator/api/api_v1/endpoints/settings.py
@@ -54,6 +54,12 @@ def get_cache_names() -> dict[str, str]:
     return CACHE_FLUSH_OPTIONS
 
 
+@router.post("/search-index/reset")
+async def reset_search_index() -> None:
+    settings.reset_search_index()
+    return
+
+
 @router.put("/status", response_model=EngineSettingsSchema)
 async def set_global_status(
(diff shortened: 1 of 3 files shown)


test/unit_tests/api/test_settings.py
```diff
@@ -43,3 +43,20 @@ def test_set_engine_status(test_client):
...
+    assert len(subscriptions) == 0
+
+    # Refresh index
+    response = test_client.post("/api/settings/search-index/reset")
+    assert response.status_code == HTTPStatus.OK
+
+    # Subscriptions are now found
+    response = test_client.get("/api/subscriptions/search?query=status:active")
+    subscriptions = response.json()
+    assert {sub["subscription_id"] for sub in subscriptions} == {generic_subscription_1, generic_subscription_2}
```
mrijk, line 62: I'd rather not have for loops in tests. Better to have 2 explicit asserts.

mrijk (changes requested):
LGTM, but please make the changes in the unit test

orchestrator/api/api_v1/endpoints/settings.py
```diff
@@ -54,6 +54,12 @@ def get_cache_names() -> dict[str, str]:
     return CACHE_FLUSH_OPTIONS
 
 
+@router.post("/search-index/reset")
+async def reset_search_index() -> None:
+    settings.reset_search_index()
+    return
```
tjeerddie, line 60: this should also work
```suggestion
   return settings.reset_search_index()
```

tjeerddie (approved):

test/unit_tests/api/test_settings.py
dmarjenburgh: It's a single collection assert. Two separate assertions for the two ids would not be the same. (I want exactly those two). Maybe the following is clearer

```suggestion
    subscription_ids  = {sub["subscription_id"] for sub in subscriptions}
    assert subscription_ids == {generic_subscription_1, generic_subscription_2}
```

mrijk: Them maybe something like:

```python
assert len(subscriptions) == 2
assert subscriptions[0]["subscription_id"] == generic_subscription_1
assert subscriptions[1]["subscription_id"] == generic_subscription_2
```

dmarjenburgh: Now it is order dependent. I think the set equality check communicates the intent most clearly. What is the problem with the original solution exactly?

mrijk: Mental overhead instead of simple asserts. Maybe not a big deal in this case, but we have loads of tests that contain code (by which I mean for-loops, conditionals, etc.) in several places. I want to avoid testing the test code, having to execute code in my head to understand what we are actually testing, etc.

Of ordering is not important, you could also use:

```
actual_subscriptions = {subscriptions[0]["subscription_id"], subscriptions[1]["subscription_id"]}
assert generic_subscription_1 in actual_subscriptions
assert generic_subscription_2 in actual_subscriptions
```

mrijk (approved):
LGTM

Issue #393, opened 2023-10-19
Make it possible to run the graphql unit_tests with the entire suite


Mark90:
Currently you have to run the tests in a two-step approach like this

```
pytest test/unit_tests/ --ignore=test/unit_tests/graphql && pytest test/unit_tests/graphql
```

This is because of the autouse fixture `fastapi_app_graphql()` in `test/unit_tests/graphql/conftest.py` which attempts to register graphql models, which somehow causes strawberry to find duplicate models. I tried to dig into it a bit with a debugger but could not find the reason.

Would be great if we can make these tests run properly as part of the main suite.

tjeerddie:
This gets fixed with my PR were the models are re-created with `register_graphql` and update the GraphQL schema with it.
https://github.com/workfloworchestrator/orchestrator-core/pull/440

Mark90:
> This gets fixed with my PR were the models are re-created with `register_graphql` and update the GraphQL schema with it. #440

Awesome :)

Issue #310, opened 2023-07-20
Fix GraphQL tests failing when executed alongside all tests


tjeerddie:
Recently, we introduced GraphQL some tests as a separate step in the unit test GitHub action due to their failure when run alongside other tests. This issue is to look for a way to fix this.

When running the GraphQL test along with other tests, an error is encountered: `The type "<class 'test.unit_tests.fixtures.products.product_blocks.product_sub_block_two.test_product_sub_block_two.<locals>.SubBlockTwoForTestInactive'>" cannot be resolved for the field "listUnionBlocks" , are you using a strawberry.field`.

Reference: [Link to the pull request discussion](https://github.com/workfloworchestrator/orchestrator-core/pull/295#discussion_r1259634240)

I suspected  that the reason for these GraphQL tests failing is that the product model classes are contained within fixtures and are re-created in every test. On the other hand, GraphQL is registered once as part of a session fixture, which auto-generates all models in `SUBSCRIPTION_MODEL_REGISTRY` during that time. As a result, only the first test has the original models, while any subsequent tests receive re-created classes.

PR #440, opened 2024-01-10
Improve ability to override existing graphql models.

tjeerddie:
- Change `register_graphql` to re-create autogenerated models and update the graphql router schema.
- add strawberry class override to change field resolvers of existing classes.
- add unit tests for override_class function and add test that overrides `Query`.
- move register_domain_models to create_graphql_router.
- change register_domain_models to take existing models with a parameter instead of a global dict.
- move `StrawberryPydanticModel`, `StrawberryModelType` and `GRAPHQL_MODELS` to graphql.schemas init.
- create default `Subscription` inside the create_graphql_router to recreate it when interface is updated.

Fixes: #393, #310
Both fixed by the ability to update schema with running `app.register_graphql` multiple times.

### DIFF
--- a/orchestrator/app.py
+++ b/orchestrator/app.py
@@ -40,7 +40,8 @@ from orchestrator.db.database import DBSessionMiddleware
 from orchestrator.distlock import init_distlock_manager
 from orchestrator.domain import SUBSCRIPTION_MODEL_REGISTRY, SubscriptionModel
 from orchestrator.exception_handlers import problem_detail_handler
-from orchestrator.graphql import Mutation, Query, create_graphql_router, register_domain_models
+from orchestrator.graphql import Mutation, Query, create_graphql_router
+from orchestrator.graphql.schemas.subscription import SubscriptionInterface
 from orchestrator.services.processes import ProcessDataBroadcastThread
 from orchestrator.settings import AppSettings, ExecutorType, app_settings
 from orchestrator.version import GIT_COMMIT_HASH
@@ -60,6 +61,8 @@ sentry_integrations: List[Integration] = [
 
 
 class OrchestratorCore(FastAPI):
+    graphql_router: Optional[Any] = None
+
     def __init__(
         self,
         title: str = "The Orchestrator",
@@ -175,10 +178,19 @@ class OrchestratorCore(FastAPI):
         """
         SUBSCRIPTION_MODEL_REGISTRY.update(product_to_subscription_model_mapping)
 
-    def register_graphql(self: "OrchestratorCore", query: Any = Query, mutation: Any = Mutation) -> None:
-        register_domain_models()
-        new_router = create_graphql_router(query, mutation)
-        self.include_router(new_router, prefix="/api/graphql")
+    def register_graphql(
+        self: "OrchestratorCore",
+        query: Any = Query,
+        mutation: Any = Mutation,
+        register_models: bool = True,
+        subscription_interface: Any = SubscriptionInterface,
+    ) -> None:
+        new_router = create_graphql_router(query, mutation, register_models, subscription_interface)
+        if not self.graphql_router:
+            self.graphql_router = new_router
+            self.include_router(new_router, prefix="/api/graphql")
+        else:
+            self.graphql_router.schema = new_router.schema
 
 
 main_typer_app = typer.Typer()
--- a/orchestrator/graphql/autoregistration.py
+++ b/orchestrator/graphql/autoregistration.py
@@ -80,7 +80,7 @@ def create_block_strawberry_type(
     return strawberry_wrapper(new_type)
 
 
-def create_subscription_strawberry_type(strawberry_name: str, model: Type[DomainModel], interface: Any) -> Any:
+def create_subscription_strawberry_type(strawberry_name: str, model: Type[DomainModel], interface: Type) -> Type:
     base_type = type(strawberry_name, (interface,), {})
     directives = [Key(fields="subscriptionId", resolvable=UNSET)]
 
@@ -95,7 +95,7 @@ def add_class_to_strawberry(
     model: Type[DomainModel],
     strawberry_models: StrawberryModelType,
     strawberry_enums: EnumDict,
-    interface: Any = None,
+    interface: Union[Type, None] = None,
 ) -> None:
     if model_name in strawberry_models:
         logger.debug("Skip already registered strawberry model", model=repr(model), strawberry_name=model_name)
@@ -118,7 +118,7 @@ def add_class_to_strawberry(
 
 
 def register_domain_models(
-    interface: Any, existing_models: Union[StrawberryModelType, None] = None
+    interface: Union[Type, None], existing_models: Union[StrawberryModelType, None] = None
 ) -> StrawberryModelType:
     strawberry_models = existing_models if existing_models else {}
     strawberry_enums: EnumDict = {}
--- a/orchestrator/graphql/autoregistration.py
+++ b/orchestrator/graphql/autoregistration.py
@@ -45,11 +45,11 @@ def is_enum(field: Any) -> bool:
 
 def create_strawberry_enums(model: Type[DomainModel], strawberry_enums: EnumDict) -> EnumDict:
     enums = {
-        key: field
+        key: create_strawberry_enum(field)
         for key, field in model._non_product_block_fields_.items()
         if is_enum(field) and is_not_strawberry_enum(key, strawberry_enums)
     }
-    return strawberry_enums | {key: create_strawberry_enum(field) for key, field in enums.items()}
+    return strawberry_enums | enums
 
 
 def graphql_name(name: str) -> str:
--- a/orchestrator/graphql/autoregistration.py
+++ b/orchestrator/graphql/autoregistration.py
@@ -80,6 +80,16 @@ def create_block_strawberry_type(
     return strawberry_wrapper(new_type)
 
 
+def create_subscription_strawberry_type(strawberry_name: str, model: Type[DomainModel], interface: Any) -> Any:
+    base_type = type(strawberry_name, (interface,), {})
+    directives = [Key(fields="subscriptionId", resolvable=UNSET)]
+
+    pydantic_wrapper = strawberry.experimental.pydantic.type(
+        model, all_fields=True, directives=directives, description=f"{strawberry_name} Type"
+    )
+    return type(strawberry_name, (pydantic_wrapper(base_type),), {})
+
+
 def add_class_to_strawberry(
     model_name: str,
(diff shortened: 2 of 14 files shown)


Mark90:
> * move register_domain_models to create_graphql_router.
> * move `StrawberryPydanticModel`, `StrawberryModelType` and `GRAPHQL_MODELS` to graphql.schemas init.

Will orchestrator implementations need to adjust imports for this?

orchestrator/app.py
```diff
@@ -175,10 +178,19 @@ def register_subscription_models(product_to_subscription_model_mapping: Dict[str
...
+        self: "OrchestratorCore",
+        query: Any = Query,
+        mutation: Any = Mutation,
+        register_models: bool = True,
+        subscription_interface: Any = SubscriptionInterface,
+    ) -> None:
+        new_router = create_graphql_router(query, mutation, register_models, subscription_interface)
+        if not self.graphql_router:
+            self.graphql_router = new_router
+            self.include_router(new_router, prefix="/api/graphql")
```
Mark90, line 191: Shouldn't the router be created only in`if not self.graphql_router` ?

Mark90: Ah, you're using to update the schema

orchestrator/graphql/autoregistration.py
```diff
@@ -44,6 +43,15 @@ def is_enum(field: Any) -> bool:
     return inspect.isclass(field) and issubclass(field, Enum)
 
 
+def create_strawberry_enums(model: Type[DomainModel], strawberry_enums: EnumDict) -> EnumDict:
+    enums = {
+        key: field
+        for key, field in model._non_product_block_fields_.items()
+        if is_enum(field) and is_not_strawberry_enum(key, strawberry_enums)
+    }
+    return strawberry_enums | {key: create_strawberry_enum(field) for key, field in enums.items()}
```
Mark90, line 52: The field can be turned into an enum in the same comprehension, to eliminate the extra comprehension at the bottom

```suggestion
    enums = {
        key: create_strawberry_enum(field)
        for key, field in model._non_product_block_fields_.items()
        if is_enum(field) and is_not_strawberry_enum(key, strawberry_enums)
    }
    return strawberry_enums | enums
```

orchestrator/graphql/autoregistration.py
```diff
@@ -81,22 +80,22 @@ def create_block_strawberry_type(
...
 def add_class_to_strawberry(
     model_name: str,
     model: Type[DomainModel],
     strawberry_models: StrawberryModelType,
     strawberry_enums: EnumDict,
-    with_interface: bool = False,
+    interface: Any = None,
 ) -> None:
+    def create_subscription_strawberry_type(strawberry_name: str, model: Type[DomainModel]) -> Type[interface]:
+        base_type = type(strawberry_name, (interface,), {})
```
Mark90, line 91: As codecov mentions these lines probably aren't hit by a unittest. When `interface` has its default value of `None` this would raise a TypeError


```
In [18]: base = type('bla', (None,), {})
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [18], in <cell line: 1>()
----> 1 base = type('bla', (None,), {})

```

orchestrator/graphql/resolvers/subscription.py
```diff
@@ -65,6 +65,13 @@ def get_subscription_details(subscription: SubscriptionTable) -> SubscriptionInt
     return strawberry_type.from_pydantic(subscription_details)
 
 
+def format_default_subscriptions(subscription: SubscriptionTable) -> SubscriptionInterface:
```
Mark90, line 68: Singular?
```suggestion
def format_default_subscription(subscription: SubscriptionTable) -> SubscriptionInterface:
```

Mark90: And maybe "base subscription" is a better word than "default subscription"?

orchestrator/graphql/schema.py
```diff
@@ -150,12 +147,33 @@ async def get_context(custom_context=Depends(custom_context_dependency)) -> Orch
     return custom_context
 
 
-def create_graphql_router(query: Any = Query, mutation: Any = Mutation) -> OrchestratorGraphqlRouter:
+GRAPHQL_MODELS_INITIAL = dict(GRAPHQL_MODELS)
```
Mark90, line 150: `GRAPHQL_MODELS_INITIAL` will always have the same content as `GRAPHQL_MODELS`, so I don't think it has any purpose?

orchestrator/graphql/schema.py
```diff
@@ -150,12 +147,33 @@ async def get_context(custom_context=Depends(custom_context_dependency)) -> Orch
...
+    )
+    class Subscription(SubscriptionInterface):
+        pass
+
+    models = GRAPHQL_MODELS_INITIAL
+    models["subscription"] = Subscription
+
+    if register_models:
+        models = register_domain_models(subscription_interface, models)
+    GRAPHQL_MODELS.update(models)
```
Mark90, line 170: The intermediate `models` variable doesn't add much imho. Also the `register_domain_models` function modifies the dictionary itself, this can become a problem sometime. We could simplify and make it more explicit;

```suggestion
    GRAPHQL_MODELS["subscription"] = Subscription

    if register_models:
        register_domain_models(subscription_interface, models=GRAPHQL_MODELS)
```

Also, `register_domain_models` will modify `GRAPHQL_MODELS` directly if the `models` kwarg is None, I think it shouldn't do that

orchestrator/graphql/utils/is_query_detailed.py
```diff
@@ -29,6 +32,8 @@ def _is_model_info_detailed(info: OrchestratorInfo) -> bool:
 
         def has_details(selection: Selection) -> bool:
             if isinstance(selection, InlineFragment):
+                if type_conditions and selection.type_condition not in type_conditions:
+                    return True
```
Mark90, line 36: Could we figure out a small unittest to hit this line?

orchestrator/graphql/utils/override_class.py
```diff
@@ -0,0 +1,31 @@
...
+            field.base_resolver = custom_field.base_resolver
+            return field
+        return field
+
+    default_class_field_names = [field.name for field in strawberry_class.__strawberry_definition__._fields]
+
+    new_field_list = [override_fn(field) for field in strawberry_class.__strawberry_definition__._fields]
+    new_field_list.extend([field for field in fields.values() if field.name not in default_class_field_names])
+
+    strawberry_class.__strawberry_definition__._fields = new_field_list
```
Mark90, line 30: Does this mean that any field not overridden/added in `fields` is removed from the strawberry class?

test/unit_tests/graphql/utils/test_override_class.py
```diff
@@ -0,0 +1,258 @@
...
+    page {{
+      customer {{
+        customerId
+        fullname
+        shortcode
+        { "newField" if override_fields else ""}
+      }}
+    }}
+  }}
+}}
```
Mark90, line 106: You could use the old `"foo { %s }" % ("bar", )` formatting to avoid having to escape the curlies

orchestrator/graphql/autoregistration.py
tjeerddie: This never gets called without a interface

orchestrator/graphql/schema.py
tjeerddie: it doesn't, because the `GRAPHQL_MODELS_INITIAL` only has all the models before starting the auto generate. the `GRAPHQL_MODELS` includes the auto registered models and the auto generate wouldn't re-create the models that already exist in the `GRAPHQL_MODELS`.

test/unit_tests/graphql/utils/test_override_class.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # reset the overriden fields.
+
+    @strawberry.field(description="Returns customer_id")  # type: ignore
+    def customer_id(self) -> str:
+        return self.customer_id
+
+    customer_id.name = "customer_id"
+
+    override_class(CustomerType, {"customer_id": customer_id})
```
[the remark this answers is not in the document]
tjeerddie: I tried and gave up after a while. don't remember the error.

orchestrator/graphql/schema.py
Mark90: In the create_graphql_router it is referenced to because of `models = GRAPHQL_MODELS_INITIAL`, after which `models` is modified. Finally there is `GRAPHQL_MODELS.update(models)` at which point they contain the same data

orchestrator/graphql/utils/is_query_detailed.py
tjeerddie: this fixes a problem of not getting any data with a detailed subscription when only asking for base properties, because the detail check becomes false and returns a `Subscription` instead of a `DetailedSubscription`.

test/unit_tests/graphql/utils/test_override_class.py
```diff
@@ -0,0 +1,293 @@
...
+    customer_id_override_field = strawberry.field(
+        name="customerId", resolver=customer_id_override, description="Returns customer_id"
+    )
+    customer_id_override_field.name = "customer_id"
+
+    def new(self) -> UUID:
+        return self.customer_id
+
+    new_field = strawberry.field(name="newField", resolver=new, description="Returns new_field")
+    new_field.name = "new_field"
```
dmarjenburgh, line 32: Can new_field just be constructed with `name="new_field"` directly?

orchestrator/graphql/autoregistration.py
```diff
@@ -81,21 +80,22 @@ def create_block_strawberry_type(
...
 
 
-def create_strawberry_enums(model: Type[DomainModel], strawberry_enums: EnumDict) -> EnumDict:
-    enums = {
-        key: field
-        for key, field in model._non_product_block_fields_.items()
-        if is_enum(field) and is_not_strawberry_enum(key, strawberry_enums)
-    }
-    return strawberry_enums | {key: create_strawberry_enum(field) for key, field in enums.items()}
+def create_subscription_strawberry_type(strawberry_name: str, model: Type[DomainModel], interface: Any) -> Any:
```
dmarjenburgh, line 83: I think `interface` should be of type `Type`?

Mark90 (approved):

test/unit_tests/graphql/utils/test_override_class.py
tjeerddie: tested again, the `strawberry.field(name="newField")` doesn't seem necessary.

dmarjenburgh (approved):
LGTM

PR #441, opened 2024-01-11
Sqlalchemy queries v2

### LINKED ISSUE #386: Update SQLAlchemy queries from v1 to v2
Changing the `Table.query` queries to use `select(Table)`

- [ ] orchestrator.api folder
- [ ] orchestrator.cli folder
- [ ] orchestrator.db folder
  - [ ] models
  - [x] sorting
  - [x] filters
- [ ] orchestrator.domain.base
- [x] orchestrator.graphql folder
- [ ] orchestrator.services folder
- [ ] orchestrator.schedules folder
- [ ] orchestrator.workflows folder

dmarjenburgh:
Migrate nearly all sqlachemy queries to the v2 syntax
https://docs.sqlalchemy.org/en/20/changelog/migration_20.html

A few cases where queries are made in methods of model objects themselves remain. These should be revisited at another point.

Closes #386


orchestrator/api/api_v1/endpoints/processes.py
```diff
@@ -200,8 +206,8 @@ async def resume_all_processess_endpoint(request: Request, user: str = Depends(r
             )
         )
         .filter(ProcessTable.is_task.is_(True))
-        .all()
     )
+    processes_to_resume = list(db.session.scalars(stmt))
```
Mark90, line 210: Apparently they moved the `.all()` method into the `ScalarResult` object. Or does it not work the same?

```suggestion
    processes_to_resume = db.session.scalars(stmt).all()
```

Mark90: Ah right, that gives the Sequence result..

Mark90: Looks like they're using [`typing.Sequence`](https://github.com/sqlalchemy/sqlalchemy/blob/2c8a05244a8f7919c3954e0b6e1efaa996792772/lib/sqlalchemy/engine/result.py#L29) though, so should be fine for mypy?

dmarjenburgh: Yeah. It's `_async_resume_processes` that wants a list, which is why I just converted. A better way is to to change `_async_resume_processes` (and similar functions) to accept `Sequence` if that is sufficient. `Sequence` is more generic (accepts lists, strings and tuples) and does not support mutations in its interface.

Shall I update function signatures?

orchestrator/api/api_v1/endpoints/subscriptions.py
```diff
@@ -293,27 +304,24 @@ def subscription_instance_in_use_by(
     if filter_statuses:
         in_use_by_instances = [sub for sub in in_use_by_instances if sub.subscription.status in filter_statuses]
 
-    return list(
-        filter(
-            lambda sub_id: sub_id != subscription_instance.subscription_id,
-            {sub.subscription_id for sub in in_use_by_instances},
-        )
-    )
+    unique_ids = {sub.subscription_id for sub in in_use_by_instances}
```
Mark90, line 307: Small thought/remark (optional since it's existing code)

This endpoint needs to return subscription ID's, but to do that it instantiates the entire SubscriptionTable model, even including duplicates.

Perhaps we can at least eliminate duplicates by using [`.unique()`](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Result.unique)? Like `subscription_instance.in_use_by.unique()` (might be the wrong syntax)

orchestrator/cli/domain_gen_helpers/product_block_helpers.py
```diff
@@ -23,35 +25,31 @@ def get_product_block_id(block_name: str) -> ScalarSelect:
...
 
     Args:
         product_blocks: Dict of product blocks mapped by product block name.
 
     Returns: Dict of product blocks by product block name to create.
     """
-    _existing_product_blocks = ProductBlockTable.query.with_entities(ProductBlockTable.name).all()
-    existing_product_blocks = {block_name[0] for block_name in _existing_product_blocks}
+    _existing_product_blocks = db.session.scalars(select(ProductBlockTable.name))
+    existing_product_blocks = set(_existing_product_blocks)
```
Mark90, line 52: Does this work?

```suggestion
    existing_product_blocks = db.session.scalars(select(ProductBlockTable.name)).unique()
```

Mark90: Or perhaps [`from sqlalchemy import distinct`](https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.distinct) which should be even better; 

```suggestion
    existing_product_blocks = db.session.scalars(select(distinct(ProductBlockTable.name)))
```

orchestrator/db/models.py
```diff
@@ -408,10 +404,6 @@ class WorkflowTable(BaseModel):
     )
     processes = relationship("ProcessTable", lazy="select", cascade="all, delete-orphan", back_populates="workflow")
 
-    @staticmethod
-    def find_by_workflow_name(name: str) -> WorkflowTable:
-        return WorkflowTable.query.filter(WorkflowTable.name == name).scalar()
```
Mark90, line 413: Are we sure these removed methods aren't used by others?

orchestrator/services/settings.py
```diff
@@ -11,22 +11,30 @@
...
+from orchestrator.db import EngineSettingsTable, db
 from orchestrator.schemas.engine_settings import EngineSettingsSchema
-from orchestrator.services.processes import SYSTEM_USER, resume_process
 from orchestrator.settings import app_settings
 
 logger = structlog.get_logger(__name__)
 
 
+def get_engine_settings() -> EngineSettingsTable:
+    """Returns the EngineSettingsTable object. Raises an exception query does not return exactly one row."""
```
Mark90, line 29:
```suggestion
    """Returns the EngineSettingsTable object. Raises an exception if the query does not return exactly one row."""
```

Mark90 (approved):

orchestrator/db/models.py
dmarjenburgh: No. This would be a breaking change.
Unfortunately, we can't rewrite this to use db.session. That would defeat the purpose of the refactor and create a cyclic dependency between modules

But there's an easy migration: Just use `orchestrator.services.workflows.get_workflow_by_name`

orchestrator/workflows/tasks/validate_products.py
```diff
@@ -123,9 +123,9 @@ def check_that_products_have_create_modify_and_terminate_workflows() -> State:
 
 @step("Check that all active products have a modify note")
 def check_that_active_products_have_a_modify_note() -> State:
-    modify_workflow = WorkflowTable.query.filter(WorkflowTable.name == "modify_note").first()
+    modify_workflow = db.session.scalars(WorkflowTable.select().filter(WorkflowTable.name == "modify_note")).first()
```
tjeerddie, line 126: to keep it consistent with the rest.
```suggestion
    modify_workflow = db.session.scalars(select(WorkflowTable).filter(WorkflowTable.name == "modify_note")).first()
```

orchestrator/api/api_v1/endpoints/subscriptions.py
dmarjenburgh: Hm. It already does a join with calling `subscription_instance.in_use_by`.
It's clearer to understand this way (less coupled to the db schema) and might be fast enough because of the existing filtering.

I suggest to revisit this as a separate issue if performance problems arise

orchestrator/cli/domain_gen_helpers/product_block_helpers.py
dmarjenburgh: The `name` column on `product_blocks` is already unique by database index. So it's actually not even necessary to make a set out of it...

orchestrator/workflows/tasks/validate_products.py
dmarjenburgh: The `WorkflowTable.select()` is different in that it  filters out soft deleted workflows, so it is to be preferred.

I have yet to find a way to do filter workflows automatically, but there is no construct in sqlalchemy out-of-the-box that can do that.

orchestrator/db/models.py
Mark90: Ok, makes sense. We'll have to remember to put this in the release notes then.

orchestrator/api/api_v1/endpoints/processes.py
Mark90: Yeah, that makes sense.

Interestingly, SqlAlchemy's `ScalarResult.all()` has return type `Sequence` but it calls `ResultInternal._allrows()` which has return type `List` (and it does always return a list). I'm not sure why they would make the return type _less_ specific 🤔

orchestrator/workflows/tasks/validate_products.py
tjeerddie: ohh oke

tjeerddie (approved):

PR #443, opened 2024-01-15
Fix search_query. Only apply camel_casing to key-term

dmarjenburgh:
Selectively applies camel_case conversion to key-terms in tsv search queries.

This allows searching for `productType:X` instead of `product_type:X`, but prevents arbitrary conversion of e.g. `L2V` to `l2_v`.

### DIFF
--- a/orchestrator/utils/search_query.py
+++ b/orchestrator/utils/search_query.py
@@ -97,6 +97,11 @@ class Lexer:
         yield Token.END,
 
 
+def camel_to_snake(s: str) -> str:
+    name = re.sub("(.)([A-Z][a-z]+)", r"\1_\2", s)
+    return re.sub("([a-z0-9])([A-Z])", r"\1_\2", name).lower()
+
+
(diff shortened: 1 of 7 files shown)


orchestrator/utils/search_query.py
```diff
@@ -97,6 +97,11 @@ def lex(self) -> Iterable[Lexeme]:
         yield Token.END,
 
 
+def camel_to_snake(s: str) -> str:
```
mrijk, line 100: I think we now have this function in multiple locations

mrijk (changes requested):
I think we can get rid of multiple implementations of snake_to_camel

tjeerddie (approved):

orchestrator/utils/search_query.py
dmarjenburgh: Moved to utils.helpers

mrijk (approved):
LGTM

Issue #446, opened 2024-01-17
[Feature]: Drop 3.9, 3,10 and add 3.12 version support.


pboers1988:
### What should we build?

Drop 3.9, 3.10 support and activate tests (non mandatory on 3.12)

Mark90:
Dependent on https://github.com/workfloworchestrator/oauth2-lib/issues/59

PR #448, opened 2024-01-18
403 graphql error handling

Mark90:


Mark90:
The error handler extension contains 3.10 syntax.

I suggest to rebase this PR after dropping 3.9/3.10 in #446 which is planned for the same milestone

tjeerddie (approved):

PR #468, opened 2024-01-22
generic enum support in generator

hanstrompert:
Do not only allow the use of int and str enums in fixed inputs but also for resource types.


orchestrator/cli/generator/generator/enums.py
```diff
@@ -0,0 +1,64 @@
...
+def is_int_enum(field: dict) -> bool:
+    return is_enum_of_type(field, "int")
+
+
+def has_default(field: dict) -> bool:
+    return "default" in field
+
+
+def convert_enum(field: dict, separator: str = ".") -> dict:
+    return (
```
mrijk, line 40: Sometimes I use an inner generator function to make this a bit more readable. You end up with something like:

```python
def convert_enum(field: dict, separator: str = ".") -> dict:
  def _convert_enum() -> Generator:
     yield from field.items
     enum_type = field["name"].title().replace("_", ""))
     yield "type", enum_type
     if has_default_field(field)
        yield "default", enum_type + separator + str(field["default"])

    return dict(_convert_enum())
```

mrijk (approved):
LGTM

orchestrator/cli/generator/generator/product.py
```diff
@@ -45,20 +40,20 @@ def generate_product(context: dict) -> None:
     fixed_inputs = config.get("fixed_inputs", [])
     product_blocks = config.get("product_blocks", [])
 
-    non_standard_fixed_inputs = get_non_standard_fixed_inputs(fixed_inputs)
-    int_enums = get_int_enum_fixed_inputs(fixed_inputs)
-    str_enums = get_str_enum_fixed_inputs(fixed_inputs)
+    non_standard_fixed_inputs = get_non_standard_fields(fixed_inputs)
+    int_enums = get_int_enums(fixed_inputs)
+    str_enums = get_str_enums(fixed_inputs)
```
tjeerddie, line 45: I think it would be cleaner to have a function that handles both. I see this done twice, here and in product_block.

orchestrator/cli/generator/generator/enums.py
```diff
@@ -0,0 +1,64 @@
...
+        | ({"default": enum_type + separator + str(field["default"])} if has_default(field) else {})
+    )
+
+
+def convert_str_enum(field: dict) -> dict:
+    return convert_enum(field, ".")
+
+
+def convert_int_enum(field: dict) -> dict:
+    return convert_enum(field, "._")
```
Mark90, line 52: Kwargs should be passed by name instead of position

```suggestion
def convert_str_enum(field: dict) -> dict:
    return convert_enum(field, separator=".")


def convert_int_enum(field: dict) -> dict:
    return convert_enum(field, separator="._")
```

orchestrator/cli/generator/generator/enums.py
```diff
@@ -0,0 +1,64 @@
...
+def get_str_enums(fields: list[dict]) -> list[dict]:
+    return [convert_str_enum(field) for field in fields if is_str_enum(field)]
+
+
+def get_int_enums(fields: list[dict]) -> list[dict]:
+    return [convert_int_enum(field) for field in fields if is_int_enum(field)]
+
+
+def to_dict(list_of_dicts: list[dict]) -> dict[str, dict]:
+    return {dict["name"]: dict for dict in list_of_dicts}
```
Mark90, line 64: Is this function related to the others?

Mark90 (approved):

orchestrator/cli/generator/generator/product.py
hanstrompert: But I do need two separate lists, one for the int_enums and one for the str_enums. When I combine both into one than the new function should either have an argument to switch between int an str, or should return both lists at once, both solutions do make it cleaner in my opinion.

orchestrator/cli/generator/generator/enums.py
hanstrompert: Yes, it acts on `field` like dict's and only works if the dict contains the `name` key.

orchestrator/cli/generator/generator/enums.py
hanstrompert: I see your point, but I'm not sure if this will increase readability 😃

Issue #465, opened 2024-01-22
GraphQL: use cached Subscription model where possible


Mark90:
https://github.com/workfloworchestrator/orchestrator-core/blob/9fa532cfdc7d8102303602b386e0bb59e8a56e1d/orchestrator/graphql/utils/get_subscription_product_blocks.py#L62

https://github.com/workfloworchestrator/orchestrator-core/blob/9fa532cfdc7d8102303602b386e0bb59e8a56e1d/orchestrator/graphql/resolvers/subscription.py#L59

These functions reconstruct the domain model from the database while it is usually already in the redis cache.

* Change to the same approach as the domain-model endpoint in the REST API
* Use a dataloader function to optimize bulk queries (depends_on, in_use_by)
* Check if [`SubscriptionModel.model_construct()`](https://docs.pydantic.dev/2.5/concepts/models/#creating-models-without-validation) is even faster

PR #469, opened 2024-01-22
Get graphql subscription from redis if present

Mark90:
Related to #465

### DIFF
--- a/orchestrator/graphql/utils/get_subscription_product_blocks.py
+++ b/orchestrator/graphql/utils/get_subscription_product_blocks.py
@@ -8,6 +8,7 @@ from strawberry.scalars import JSON
 
 from orchestrator.domain.base import SubscriptionModel
 from orchestrator.services.subscriptions import build_extended_domain_model
+from orchestrator.utils.redis import from_redis
 
 
 @strawberry.type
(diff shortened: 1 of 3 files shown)


orchestrator/graphql/utils/get_subscription_product_blocks.py
```diff
@@ -58,8 +59,11 @@ def new_product_block(item: dict[str, Any]) -> Generator:
 async def get_subscription_product_blocks(
     subscription_id: UUID, tags: Optional[list[str]] = None, product_block_instance_values: Optional[list[str]] = None
 ) -> list[ProductBlockInstance]:
-    subscription_model = SubscriptionModel.from_subscription(subscription_id)
-    subscription = build_extended_domain_model(subscription_model)
+    if cached_model := from_redis(subscription_id):
```
tjeerddie, line 62: might be good to do the same in `subscriptions` resolver `get_subscription_details`

mrijk: I would probably move this to a separate function, because for get_subscription_product_blocks it's not relevant where the subscription comes from. That also makes it easy to reuse it in other resolvers/functions

Mark90: I couldn't figure out how to make a query that uses `get_subscription_details()`

Also it uses the subscription as a model, not as dict. As far as I can tell this (get_subscription_product_blocks) is the only place where we need the subscription as a dict

Mark90: Made it a function

PR #500, opened 2024-02-16
Replace GraphQL extensions with the nwastdlib versions

tjeerddie:


orchestrator/db/filters/search_filters/inferred_filter.py
```diff
@@ -123,15 +123,15 @@ def _clause_gen(node: Node) -> BinaryExpression | ColumnElement[bool]:
 
 def inferred_filter(field: ColumnClause) -> WhereCondGenerator:
     python_type = field.type.python_type
-    if python_type == str:
+    if python_type is str:
```
mrijk, line 126: I think this could be replaced by a match/case

mrijk (commented):
LGTM

PR #553, opened 2024-03-07
Add 'distinct' clause to subscription helper function

Sparrow1029:
During updates & testing of ESnet code while migrating to orchestrator-core v2.X, encountered an error from the `retrieve_subscription_by_subscription_instance_value` helper function:

```python
MultipleResultsFound: Multiple rows were found when one or none was required
```

This was for a Product which has multiple child product blocks which could each have a resource type with the same value, for example:

```text
ExampleProduct
 |_ExampleBlock
 | |_SubBlock
 |   + resource_type_1: 42
 |   + other_resource: "ingress"
 |_ExampleBlock
   |_SubBlock
     + resource_type_1: 42
     + other_resource: "egress"
```

In this example, the joins from the helper function would return the _*same*_ subscription object twice. If this is a valid product structure, where there can be multiple blocks of the same type reference in a subscription, then adding a `SELECT ... DISTINCT ON subscriptions.subscription_id` should still be valid for this helper query.


tjeerddie (approved):
LGTM

Mark90 (approved):

Issue #547, opened 2024-03-07
[CI]: Enable pre-commit before merge


crankynetman:
We should have running the pre-commit tasks that are defined in the repo be a requirement before we merge to main. We could use github actions to either run this on ever push or perhaps just every PR. Currently there are a lot of files in this repo that don't comply to our pre-commit checks, so we should also fix all of those at the time of us enabling pre-commit as a github action.

PR #554, opened 2024-03-07
ci: Run pre-commit before merge

crankynetman:
This should implement #547 . Only thing I'm not in love with here is that I disabled `blacken-docs` as it can't handle our overloading of the bitshift operator and has no way to ignore the error without changing the code blocks that were failing to not be `python` codeblocks. I looked through the blacken-docs codebase though, but it's a bit janky and I am not sure that I see the value in putting a PR over there; I talked w/ @pboers1988 and we figured we should just leave this commit hook out for now. I left it commented out for now in case we change our minds.


hanstrompert (changes requested):
This PR removes double spaces at the end of lines in MD files in de docs tree. But double spaces are used as line breaks to format the text. Can you maybe exclude the MD files?

crankynetman:
@hanstrompert excellent catch, thank you! I was not aware that markdown supports this, TIL. 

I found a feature in the pre-commit whitespace check that allows us to specify what our markdown file extension is (apparently it defaults to something other than `.md`) and it treats these spaces appropriately. I've gone and re-added the spaces too where I think that they were intentional. 


Now I just need to figure out why unit tests are failing and I'll mark this out of draft.

pboers1988 (approved):

PR #560, opened 2024-03-12
Fix GraphQL subscription status with detailed information

tjeerddie:
- status stays INITIAL because we convert subscriptions to the initial lifecycle but now we overwite it back with the original subscription status.
- update graphql single subscription tests to also check if the status is correct.
- add forgotten traceback to enrich_process.

### DIFF
--- a/orchestrator/graphql/resolvers/subscription.py
+++ b/orchestrator/graphql/resolvers/subscription.py
@@ -58,11 +58,12 @@ def get_subscription_details(subscription: SubscriptionTable) -> SubscriptionInt
 
     subscription_details = SubscriptionModel.from_subscription(subscription.subscription_id)
     base_model = subscription_details.__base_type__ if subscription_details.__base_type__ else subscription_details
-    subscription_details = base_model.from_other_lifecycle(  # type: ignore
+    base_subscription_details = base_model.from_other_lifecycle(  # type: ignore
         subscription_details, SubscriptionLifecycle.INITIAL, skip_validation=True
     )
+    base_subscription_details.status = subscription_details.status
     strawberry_type = GRAPHQL_MODELS[graphql_subscription_name(base_model.__name__)]  # type: ignore
-    return strawberry_type.from_pydantic(subscription_details)
+    return strawberry_type.from_pydantic(base_subscription_details)
 
 
 def format_base_subscription(subscription: SubscriptionTable) -> SubscriptionInterface:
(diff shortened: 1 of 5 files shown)


test/unit_tests/graphql/test_subscriptions.py
```diff
@@ -877,6 +877,7 @@ def test_single_subscription(test_client, product_type_1_subscriptions_factory,
         "totalItems": 1,
     }
     assert subscriptions[0]["subscriptionId"] == subscription_id
+    assert subscriptions[0]["status"] == "ACTIVE"
```
mrijk, line 880: Maybe start using enums for this?

test/unit_tests/graphql/test_subscriptions.py
```diff
@@ -877,6 +877,7 @@ def test_single_subscription(test_client, product_type_1_subscriptions_factory,
         "totalItems": 1,
     }
     assert subscriptions[0]["subscriptionId"] == subscription_id
```
mrijk, line 879: subscription = subscriptions[0]

and then using that is a bit easier on the eye.

mrijk (approved):
LGTM

test/unit_tests/graphql/test_subscriptions.py
tjeerddie: its done on many places and we have a story for changing tests to assert with snapshots. so don't find it usefull to update here.

orchestrator/graphql/resolvers/subscription.py
```diff
@@ -58,11 +58,12 @@ def get_subscription_details(subscription: SubscriptionTable) -> SubscriptionInt
 
     subscription_details = SubscriptionModel.from_subscription(subscription.subscription_id)
     base_model = subscription_details.__base_type__ if subscription_details.__base_type__ else subscription_details
-    subscription_details = base_model.from_other_lifecycle(  # type: ignore
+    base_subscription_details = base_model.from_other_lifecycle(  # type: ignore
         subscription_details, SubscriptionLifecycle.INITIAL, skip_validation=True
```
Mark90, line 62: Why do we convert subscriptions to the initial lifecycle again?

Mark90 (approved):

orchestrator/graphql/resolvers/subscription.py
tjeerddie: because it wants the nested models to be the exact instances.

`Expected value of type 'Sn8ServiceAttachPointBlockInactive' but got: <Sn8ServiceAttachPointBlock instance>.`

pboers1988 (approved):

PR #561, opened 2024-03-13
Fix search-index reset

dmarjenburgh:
Fixes #513

PR #563, opened 2024-03-13
Add refresh search index step post wf

dmarjenburgh:
NOTE:
Should wait for https://github.com/workfloworchestrator/orchestrator-core/pull/561 to be merged

### DIFF
--- a/orchestrator/workflows/utils.py
+++ b/orchestrator/workflows/utils.py
@@ -31,6 +31,7 @@ from orchestrator.utils.state import form_inject_args
 from orchestrator.workflow import StepList, Workflow, conditional, done, init, make_workflow, step
 from orchestrator.workflows.steps import (
     cache_domain_models,
+    refresh_subscription_search_index,
     remove_domain_model_from_cache,
     resync,
     set_status,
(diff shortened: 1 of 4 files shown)


orchestrator/workflows/utils.py
```diff
@@ -322,6 +326,7 @@ def _validate_workflow(f: Callable[[], StepList]) -> Workflow:
             >> f()
             >> resync
             >> push_subscriptions(cache_domain_models)
+            >> refresh_subscription_search_index
```
pboers1988, line 329: Curious to know what the impact of this will be on the engine, we used to have something similar to this, but we disabled it due to the performance cost (the database trigger). Even though it is not exactly the same, I'm a little concerned that this might be a bit too much. When running all validations in parallel, this will lock all tables way too often, imho and update the TSV, when there is no actual change to the subscription.

pboers1988 (changes requested):
As mentioned I do not think we should add the reset TSV to the validation workflow decorator due to the performance cost, and due to the fact that a subscription does not change during a validation workflow. If we feel we need a way to automatically trigger a TSV reset, we should create a built-in automatically run task.

orchestrator/workflows/utils.py
dmarjenburgh: I removed it

PR #579, opened 2024-03-21
Vendor broadcaster package

Mark90:
Adds directory `orchestrator/vendor/broadcaster` which is a clone of https://github.com/encode/broadcaster from commit 377b404, taken at 2024-03-21. 

The last release of broadster is 0.2.0 dating back to 2020. Since then the master branch contains a bugfix which we need for the `rediss://` backend.

As pypi does not allow direct github dependencies, we vendor broadcaster's source code into this codebase until broadcaster receives a new release. https://github.com/encode/broadcaster/issues/103

---

The above text has also been added to `orchestrator/vendor/broadcaster/README.md`.

The path `orchestrator/vendor` is excluded from all linters, formatters, type checkers, coverage and pre-commit.

Created #580 for follow-up.


orchestrator/vendor/broadcaster/broadcaster/_base.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def __repr__(self) -> str:
+        return f"Event(channel={self.channel!r}, message={self.message!r})"
+
+
+class Unsubscribed(Exception):
+    pass
+
+
+class Broadcast:
```
tjeerddie, line 27: You could have used a submodule or only add this file, since no changes are needed in other files :3

Mark90: I left the broadcaster code as-is in case there are bugs with this commit, then it's easy to go to a different commit.

Yeah I considered a git submodule, but that would mean other orchestrator-core users will need to learn about that as well. And they're not the easiest thing in the world to work with.

Mark90: Note that this is not meant to be permanent, created #580 for a long term solution

PR #597, opened 2024-04-03
Refactor db filtering

dmarjenburgh:


pboers1988 (approved):
LGTM!

orchestrator/db/filters/process.py
```diff
@@ -11,152 +11,97 @@
...
+    "subscription_id": make_subscription_id_clause(inferred_filter(SubscriptionTable.subscription_id)),
     "tag": make_product_clause(filter_exact(ProductTable.tag)),
+    "target": workflow_targets_clause,
+    "workflow_name": workflow_name_clause,
 }
 
 
-process_filter_fields = list(PROCESS_FILTER_FUNCTIONS_BY_COLUMN.keys())
-filter_processes = generic_filter(PROCESS_FILTER_FUNCTIONS_BY_COLUMN)
+process_filter_fields = list(PROCESS_TABLE_COLUMN_CLAUSES.keys())
```
tjeerddie, line 106: This should be a function to get filter fields, otherwise these keys will never from the initial `PROCESS_TABLE_COLUMN_CLAUSES ` keys, extra filters from projects that use this package will never be added to the list.

```suggestion
def get_process_filter_fields() -> list[str]:
    return list(PROCESS_TABLE_COLUMN_CLAUSES.keys())
```

should be changed everywhere

dmarjenburgh: Shall we change it for all filter and sort lists for all resources?

dmarjenburgh: Done for all filter/sorting field lists

tjeerddie: nice!

tjeerddie (approved):

orchestrator/db/filters/filters.py
```diff
@@ -32,57 +34,103 @@ class Filter(BaseModel):
...
+
+
+def _filter_to_node(filter_item: Filter) -> Node:
+    value = filter_item.value
+    if value and value[0] == "!":
+        updated_filter_item = Filter(field=filter_item.field, value=value[1:])
+        return "Negation", _filter_to_node(updated_filter_item)
+
+    # Workaround to deal with date fields. These should not be split like other fields.
+    # Fix after deprecating '-' as split-separator in favor of '|'
```
Mark90, line 62: Is there a follow-up ticket for this?

orchestrator/db/filters/filters.py
```diff
@@ -32,57 +34,103 @@ class Filter(BaseModel):
...
+    value_node: Node
+    if len(values) > 1:
+        value_node = "ValueGroup", [("Word", v) for v in values if v]
+    else:
+        value_node = "Word", value
+
+    if len(value_node[1]) == 0:
+        raise Exception("Invalid filter arguments")
+
+    return "KVTerm", (key_node, value_node)
```
Mark90, line 74: When the terms `ValueGroup` and `KVTerm` were only used in the `utils/search_query.py` module it was manageable, but now that we're using them in other modules I think they should become constant/enum variables to avoid typo's. Also the devs who don't know about the search query stuff can then navigate to the definition for context

orchestrator/db/filters/process.py
```diff
@@ -11,148 +11,101 @@
...
     "tag": make_product_clause(filter_exact(ProductTable.tag)),
+    "target": workflow_targets_clause,
+    "workflow_name": workflow_name_clause,
 }
 
 
-process_filter_fields = list(PROCESS_FILTER_FUNCTIONS_BY_COLUMN.keys())
-filter_processes = generic_filter(PROCESS_FILTER_FUNCTIONS_BY_COLUMN)
+process_filter_fields = create_memoized_field_list(PROCESS_TABLE_COLUMN_CLAUSES)
+filter_processes = generic_filter_from_clauses(PROCESS_TABLE_COLUMN_CLAUSES)
```
Mark90, line 111: Just to check whether I'm on the right page:
* These definitions  no longer contain camelCase keys but only snake_case keys
* The search query filters can be any of the 2: it is just normalized to snake_case
* In the error response to a query (in case of invalid filters) we send back the keys in camelCase because for GQL that makes the most sense

Mark90 (commented):
Nice!

orchestrator/db/filters/filters.py
dmarjenburgh: Will do

orchestrator/db/filters/process.py
dmarjenburgh: Correct. Everything concertning the FE/JSON is camelCase, but in the inner python code we work with snake_case

orchestrator/db/filters/filters.py
Mark90: created https://github.com/workfloworchestrator/orchestrator-core/issues/779

PR #598, opened 2024-04-03
Query a single subscription

pboers1988:


test/unit_tests/graphql/test_subscription.py
```diff
@@ -0,0 +1,80 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import json
+from http import HTTPStatus
+from uuid import uuid4
+
+
+def build_simple_query(subscription_id):
+    q = f"""
```
mrijk, line 19: I think we can make these queries a lot more readable by using graphql variables (like we do in Pythia)

mrijk (approved):
LGTM, apart from the difficult to read queries

test/unit_tests/graphql/test_subscription.py
tjeerddie: create a issue to update this :3

tjeerddie (dismissed):

test/unit_tests/graphql/test_subscription.py
mrijk: Why? This is a small change

PR #603, opened 2024-04-04
Fix search queries for all Postgres versions

### LINKED ISSUE #602: [Bug]: Fix Subscriptions search for the different postgres versions
### What happened?

Using more complex search syntax, like the `|` operator test fail as the database does not return a result. This test also used to be flaky, so it might be caused by a race condition somewhere, however higher versions of postgres do seem to make the results more predictable, indicating a relationship to the database version.

See the result of a matrix test run, where the Python version and Postgres version are manipulated.
https://github.com/workfloworchestrator/orchestrator-core/actions/runs/8555513507/attempts/2?pr=599

### Version

alpha

### What Postgres version are you seeing the problem on?

Postgres 11, Postgres 12, Postgres 13

dmarjenburgh:
Resolves #602 

- Add `filter_uuid_exact` clause matcher
- Implement Postgres version dependent splitting of search query terms

### DIFF
--- a/orchestrator/db/helpers.py
+++ b/orchestrator/db/helpers.py
@@ -1,7 +1,27 @@
+import functools
+
+import structlog
+from sqlalchemy import text
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.sql.elements import CompilerElement
+
+from orchestrator.db import db
+
+logger = structlog.get_logger(__name__)
 
 
 def to_sql_string(stmt: CompilerElement) -> str:
     dialect = postgresql.dialect()  # type: ignore
     return str(stmt.compile(dialect=dialect, compile_kwargs={"literal_binds": True}))
+
+
+@functools.cache
+def get_postgres_version() -> int:
+    """Returns the Postgres major version as an int."""
+    try:
+        # The pg_version_num is pg_major_version * 10000 + pg_minor_version
+        pg_version_num = int(db.session.scalar(text("show server_version_num")))
+        return pg_version_num // 10000
+    except ValueError:
+        logger.error("Unable to query Postgres version")
+        return 0
--- a/orchestrator/utils/search_query.py
+++ b/orchestrator/utils/search_query.py
@@ -5,10 +5,12 @@ from itertools import chain
 from typing import Any, Union, cast
 
 import structlog
-from sqlalchemy import BinaryExpression, ColumnElement, CompoundSelect, Select, false, not_, or_
+from sqlalchemy import BinaryExpression, ColumnElement, CompoundSelect, Select, column, false, func, not_, or_, select
 from sqlalchemy.orm import InstrumentedAttribute, MappedColumn
 
+from orchestrator.db import db
 from orchestrator.db.database import BaseModel
+from orchestrator.db.helpers import get_postgres_version
 from orchestrator.utils.helpers import camel_to_snake
 
 logger = structlog.getLogger(__name__)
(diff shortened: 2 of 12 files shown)


orchestrator/db/helpers.py
```diff
@@ -1,7 +1,27 @@
...
+
+@functools.cache
+def get_postgres_version() -> int:
+    """Returns the Postgres major version as an int."""
+    try:
+        # The pg_version_num is pg_major_version * 10000 + pg_minor_version
+        pg_version_num = int(db.session.scalar(text("show server_version_num")))
+        return pg_version_num // 10000
+    except ValueError:
+        logger.error("Unable to get query Postgres version")
```
mrijk, line 26: Just 'get' or 'query'

Josephine-Rutten (approved):

orchestrator/utils/search_query.py
```diff
@@ -255,6 +257,29 @@ def parse(self) -> Node:
...
+    _glue_chars = re.compile(r"[-_@#$%^&]")
+
+    @staticmethod
+    def _split_term(term: str) -> list[str]:
+        """Workaround for the way Postgres <14 parses text with to_tsquery.
+
+        For Postgres <14, we issue a database query to parse the text for us in a way that is compatible with to_tsvector. This is the same behavior as Postgres >=14.
+        """
+
+        # TODO: Remove this workaround when support for Postgres <14 is dropped
```
Mark90, line 270: Todo should be linked to a ticket

orchestrator/utils/search_query.py
```diff
@@ -255,6 +257,29 @@ def parse(self) -> Node:
...
+
+        if get_postgres_version() < 14:
```
Mark90, line 272: Maybe store the version in a global variable or contextvar, to reduce the overhead to the first query that an API instance processes? 

We won't ever do a major upgrade on the database while keeping the app running, so there's no concern of this value going stale. And it's temporary code anyway.

orchestrator/utils/search_query.py
```diff
@@ -287,16 +312,29 @@ def visit_kv_term(node: Node, acc: list[str]) -> None:
...
-        if node[0] == "Word":
-            acc.append(text)
-        elif node[0] == "PrefixWord":
-            acc.append(f"{text}:*")
-        else:
-            raise Exception(f"Invalid SearchWord type {node[0]}")
+
+        text = node[1].replace("'", '"')
+
+        parts = TSQueryVisitor._split_term(node[1])
```
Mark90, line 324: I think `node[1]` should be assigned to a variable on which the single quotes are replaced. Right now the `parts` result still contains single quotes which is probably not intended

orchestrator/utils/search_query.py
```diff
@@ -287,16 +312,29 @@ def visit_kv_term(node: Node, acc: list[str]) -> None:
...
+
+        if not parts:
+            return
+
+        if len(parts) > 1:
+            text = " <-> ".join(parts)
```
Mark90, line 330: The len check is not needed probably?

```suggestion
        text = " <-> ".join(parts)
```

orchestrator/db/helpers.py
dmarjenburgh: I don't think that works. `SHOW server_version_num` is the entire sql query. Not a select from a table

orchestrator/utils/search_query.py
dmarjenburgh: `get_postgres_version` is memoized.

orchestrator/utils/search_query.py
dmarjenburgh: Done

orchestrator/utils/search_query.py
dmarjenburgh: Done

orchestrator/db/helpers.py
Mark90: I think he referred to the wording of the log statement

Mark90 (approved):

orchestrator/db/helpers.py
dmarjenburgh: Aha :sweat_smile:

PR #608, opened 2024-04-10
DRAFT: Refactor OAUTH handling for configurable AuthN and AuthZ

torkashvand:


Mark90:
Can you change this in `pyproject.toml` ?

```
 # "oauth2-lib~=1.5.0",  # TODO uncomment and pin to new oauth2-lib release after https://github.com/workfloworchestrator/oauth2-lib/pull/61 is merged
 "oauth2-lib @ git+https://github.com/torkashvand/oauth2-lib@oauth2-enhancements",
```

This will allow us to test with and implement changes based on this branch

torkashvand:
And to give you @Mark90 and idea of usage it would be something like our usage in GEANT

```python
"""OpenID Connect and Open Policy Agent Integration for GSO Application.

This module provides helper functions and classes for handling OpenID Connect (OIDC) and
Open Policy Agent (OPA) related functionalities within the GSO application. It includes
implementations for OIDC-based user authentication and user information modeling. Additionally,
it facilitates making authorization decisions based on policies defined in OPA. Key components
comprise OIDCUser, OIDCUserModel, OPAResult, and opa_decision. These elements integrate with
FastAPI to ensure secure API development.
"""

import re
from functools import wraps
from http import HTTPStatus
from json import JSONDecodeError

from fastapi.exceptions import HTTPException
from fastapi.requests import Request
from httpx import AsyncClient
from oauth2_lib.fastapi import OIDCAuth, OIDCUserModel
from oauth2_lib.settings import oauth2lib_settings
from structlog import get_logger

logger = get_logger(__name__)

_CALLBACK_STEP_API_URL_PATTERN = re.compile('somethong')


def _is_callback_step_endpoint(request: Request) -> bool:
    """Check if the request is a callback step API call."""
    return re.match(_CALLBACK_STEP_API_URL_PATTERN, request.url.path) is not None


def ensure_openid_config_loaded(func):
    @wraps(func)
    async def wrapper(self, async_request: AsyncClient, *args, **kwargs):
        await self.check_openid_config(async_request)
        return await func(self, async_request, *args, **kwargs)

    return wrapper


class OIDCAuthentication(OIDCAuth):
    """OIDCUser class extends the :term:`HTTPBearer` class to do extra verification.

    The class will act as follows:
        1. Validate the Credentials at :term: `AAI` proxy by calling the UserInfo endpoint
    """

    _instance = None

    def __new__(cls, *args, **kwargs):
        if cls._instance is None:
            cls._instance = super(OIDCAuthentication, cls).__new__(cls)
        return cls._instance

    @staticmethod
    async def is_bypassable_request(request: Request) -> bool:
        return _is_callback_step_endpoint(request=request)

    @ensure_openid_config_loaded
    async def userinfo(self, async_request: AsyncClient, token: str) -> OIDCUserModel:
        """Get the userinfo from the openid server.

        :param AsyncClient async_request: The async request
        :param str token: the access_token
        :return: OIDCUserModel: OIDC user model from openid server

        """
        await self.introspect_token(async_request, token)

        response = await async_request.post(
            self.openid_config.userinfo_endpoint,
            data={"token": token},
            headers={"Authorization": f"Bearer {token}"},
        )
        try:
            data = dict(response.json())
        except JSONDecodeError as err:
            logger.debug(
                "Unable to parse userinfo response",
                detail=response.text,
                resource_server_id=self.resource_server_id,
                openid_url=self.openid_url,
            )
            raise HTTPException(status_code=HTTPStatus.UNAUTHORIZED, detail=response.text) from err
        logger.debug("Response from openid userinfo", response=data)

        if response.status_code not in range(200, 300):
            logger.debug(
                "Userinfo cannot find an active token, user unauthorized",
                detail=response.text,
                resource_server_id=self.resource_server_id,
                openid_url=self.openid_url,
            )
            raise HTTPException(status_code=HTTPStatus.UNAUTHORIZED, detail=response.text)

        return OIDCUserModel(data)

    @ensure_openid_config_loaded
    async def introspect_token(self, async_request: AsyncClient, token: str) -> dict:
        """Introspect the access token to see if it is a valid token.

        :param async_request: The async request
        :param token: the access_token
        :return: dict from openid server
        """
        endpoint = self.openid_config.introspect_endpoint or self.openid_config.introspection_endpoint or ""
        response = await async_request.post(
            endpoint,
            data={"token": token, "client_id": self.resource_server_id},
            headers={"Content-Type": "application/x-www-form-urlencoded"},
        )

        try:
            data = dict(response.json())
        except JSONDecodeError as err:
            logger.debug(
                "Unable to parse introspect response",
                detail=response.text,
                resource_server_id=self.resource_server_id,
                openid_url=self.openid_url,
            )
            raise HTTPException(status_code=HTTPStatus.UNAUTHORIZED, detail=response.text) from err

        logger.debug("Response from openid introspect", response=data)

        if response.status_code not in range(200, 300):
            logger.debug(
                "Introspect cannot find an active token, user unauthorized",
                detail=response.text,
                resource_server_id=self.resource_server_id,
                openid_url=self.openid_url,
            )
            raise HTTPException(status_code=HTTPStatus.UNAUTHORIZED, detail=response.text)

        if "active" not in data:
            logger.error("Token doesn't have the mandatory 'active' key, probably caused by a caching problem")
            raise HTTPException(status_code=HTTPStatus.UNAUTHORIZED, detail="Missing active key")
        if not data.get("active", False):
            logger.info("User is not active", user_info=data)
            raise HTTPException(status_code=HTTPStatus.UNAUTHORIZED, detail="User is not active")

        return data


oidc_instance = OIDCAuthentication(
    openid_url=oauth2lib_settings.OIDC_BASE_URL,
    openid_config_url=oauth2lib_settings.OIDC_CONF_URL,  # Corrected parameter name
    resource_server_id=oauth2lib_settings.OAUTH2_RESOURCE_SERVER_ID,
    resource_server_secret=oauth2lib_settings.OAUTH2_RESOURCE_SERVER_SECRET,
    oidc_user_model_cls=OIDCUserModel,
)
```

and your OIDCUserModel I think would be like this:

```python
class SURFOIDCUserModel(OIDCUserModel):
    """
    Below this line are SURFnet specific properties that are used often in code to either display or check on.
    """

    @property
    def user_name(self) -> str:
        if "user_name" in self.keys():
            return cast(str, self["user_name"])
        if "unspecified_id" in self.keys():
            return cast(str, self["unspecified_id"])
        return ""

    @property
    def display_name(self) -> str:
        return self.get("display_name", "")  # type: ignore

    @property
    def principal_name(self) -> str:
        return self.get("eduperson_principal_name", "")  # type: ignore

    @property
    def memberships(self) -> list[str]:
        return self.get("edumember_is_member_of", [])  # type: ignore

    @property
    def teams(self) -> set[str]:
        prefix = "urn:collab:group:surfteams.nl:nl:surfnet:diensten:"
        length = len(prefix)
        return {urn[length:] for urn in self.memberships if urn.startswith(prefix)}

    @property
    def entitlements(self) -> list[str]:
        return self.get("eduperson_entitlement", [])  # type: ignore

    @property
    def roles(self) -> set[str]:
        prefix = "urn:mace:surfnet.nl:surfnet.nl:sab:role:"
        length = len(prefix)
        return {urn[length:] for urn in self.entitlements if urn.startswith(prefix)}

    @property
    def organization_codes(self) -> set[str]:
        prefix = "urn:mace:surfnet.nl:surfnet.nl:sab:organizationCode:"
        length = len(prefix)
        return {urn[length:] for urn in self.entitlements if urn.startswith(prefix)}

    @property
    def organization_guids(self) -> set[str]:
        prefix = "urn:mace:surfnet.nl:surfnet.nl:sab:organizationGUID:"
        length = len(prefix)
        return {urn[length:] for urn in self.entitlements if urn.startswith(prefix)}

    @property
    def scopes(self) -> set[str]:
        if isinstance(self.get("scope"), list):
            return set(self.get("scope"))  # type: ignore
        return set(re.split("[ ,]", self.get("scope", "")))

    @property
    def is_resource_server(self) -> bool:
        return self.get("is_resource_server", False)  # type: ignore

    @property
    def surf_crm_id(self) -> str:
        return self.get("surf-crm-id", "")  # type: ignore

```

Mark90:
> And to give you @Mark90 an idea of usage it would be something like our usage in GEANT

Thanks!


It seems your `OIDCAuthentication.user_info` calls `OIDCAuthentication.introspect_token` but does not use the return value. They also do quite similar things. Is this intended?

I'm also wondering if you are using a custom Authorization class?

orchestrator/security.py
```diff
@@ -10,41 +10,103 @@
...
 oauth_client_credentials.register(
     "connext",
-    server_metadata_url=well_known_endpoint / ".well-known" / "openid-configuration",
-    client_id=oauth2_settings.OAUTH2_RESOURCE_SERVER_ID,
-    client_secret=oauth2_settings.OAUTH2_RESOURCE_SERVER_SECRET,
+    server_metadata_url=URL(oauth2lib_settings.OIDC_CONF_URL),
+    client_id=oauth2lib_settings.OAUTH2_RESOURCE_SERVER_ID,
+    client_secret=oauth2lib_settings.OAUTH2_RESOURCE_SERVER_SECRET,
     request_token_params={"grant_type": "client_credentials"},
     client_kwargs={"verify": HTTPX_SSL_CONTEXT},
```
Mark90, line 37: It seems we're not using this `oauth_client_credentials` object or its return value anywhere, also checked in our surf orchestrator. If it's the same for you then I think we can remove this?

Or perhaps move it to oauth2-lib and use it for interacting with the OIDC backend. It has some built-in logic for verifying tokens https://docs.authlib.org/en/latest/client/starlette.html#register-remote-apps I'm not sure how to use it though.

torkashvand: We also don't use it in our code and I wasn't sure if it causes bug if I remove it so that I decided to leave it and adapt it with new changes 

> It seems we're not using this `oauth_client_credentials` object or its return value anywhere, also checked in our surf orchestrator. If it's the same for you then I think we can remove this?
> 
> Or perhaps move it to oauth2-lib and use it for interacting with the OIDC backend. It has some built-in logic for verifying tokens https://docs.authlib.org/en/latest/client/starlette.html#register-remote-apps I'm not sure how to use it though.

orchestrator/__init__.py
```diff
@@ -16,13 +16,12 @@
 __version__ = "2.2.2rc2"
```
pboers1988, line 16: We should make this at least a Minor version bump.

orchestrator/app.py
```diff
@@ -207,6 +210,15 @@ def register_graphql(
         else:
             self.graphql_router.schema = new_router.schema
 
+    def register_authentication(self, authentication_instance: OIDCAuth) -> None:
+        self.auth_manager.authentication = authentication_instance
```
pboers1988, line 214: Docs?

orchestrator/app.py
```diff
@@ -207,6 +210,15 @@ def register_graphql(
...
+
+    def register_authorization(self, authorization_instance: Authorization) -> None:
```
pboers1988, line 216: Docs?

orchestrator/app.py
```diff
@@ -207,6 +210,15 @@ def register_graphql(
...
+        self.auth_manager.authorization = authorization_instance
+
+    def register_graphql_authorization(self, graphql_authorization_instance: GraphqlAuthorization) -> None:
```
pboers1988, line 219: Docs?

pboers1988 (changes requested):
@torkashvand codewise it LGTM. I would greatly appreciate some documentation added to the Docs and that all pending conversations are resolved. We are reviewing this internally and should get any other QA issues to you by early nextweek.

Thanks!

orchestrator/app.py
torkashvand: Done!

orchestrator/app.py
torkashvand: Done!

orchestrator/app.py
torkashvand: Done!

pyproject.toml
```diff
@@ -61,7 +61,8 @@ dependencies = [
     "typer==0.12.0",
     "uvicorn[standard]~=0.29.0",
     "nwa-stdlib~=1.7.0",
-    "oauth2-lib~=1.5.0",
+ # "oauth2-lib~=1.5.0",  # TODO uncomment and pin to new oauth2-lib release after https://github.com/workfloworchestrator/oauth2-lib/pull/61 is merged
```
torkashvand, line 64: We should remove this when we get approval!

orchestrator/app.py
torkashvand: I found an empty file named `auth.md` in the docs folder, and I added a document to it. You can view the final version here:

https://github.com/workfloworchestrator/orchestrator-core/blob/feature/configurable-auth-handling/docs/reference-docs/auth.md

Is that what you want?

pyproject.toml
torkashvand: Done!

pboers1988 (approved):
LGTM

PR #609, opened 2024-04-12
Add subscription customer description mutations

tjeerddie:
- upsert and remove subscription customer descriptions.

### DIFF
--- a/orchestrator/api/api_v1/endpoints/subscription_customer_descriptions.py
+++ b/orchestrator/api/api_v1/endpoints/subscription_customer_descriptions.py
@@ -11,18 +11,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from datetime import datetime
 from http import HTTPStatus
 from uuid import UUID
 
 from fastapi.param_functions import Body
 from fastapi.routing import APIRouter
-from pytz import timezone
-from sqlalchemy import select
 
 from orchestrator.api.error_handling import raise_status
-from orchestrator.api.models import delete, save, update
+from orchestrator.api.models import delete
 from orchestrator.db import SubscriptionCustomerDescriptionTable, db
+from orchestrator.domain.customer_description import (
+    create_subscription_customer_description,
+    get_customer_description_by_customer_subscription,
+    update_subscription_customer_description,
+)
 from orchestrator.schemas import SubscriptionDescriptionBaseSchema, SubscriptionDescriptionSchema
 from orchestrator.utils.redis import delete_from_redis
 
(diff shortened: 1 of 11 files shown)


orchestrator/graphql/types.py
```diff
@@ -116,3 +116,14 @@ class GraphqlFilter:
...
+
+@strawberry.type(description="Generic class to capture errors")
+class MutationError:
+    message: str = strawberry.field(description="Error message")
+    details: str | None = strawberry.field(description="Details of error cause", default=None)
+
+
+@strawberry.type(description="Error class if a resource couldn't be found (404)")
+class NotFound(MutationError):  # noqa: N818
+    pass
```
Mark90, line 129: I suggest to suffix all error types with Error just to make it easily distinguishable. It's a relatively common pattern as well

```suggestion
class NotFoundError(MutationError):  # noqa: N818
    pass
```

orchestrator/api/api_v1/endpoints/subscription_customer_descriptions.py
```diff
@@ -11,36 +11,38 @@
...
 
 router = APIRouter()
 
 
 @router.post("/", response_model=None, status_code=HTTPStatus.NO_CONTENT)
-def save_subscription_customer_description(data: SubscriptionDescriptionBaseSchema = Body(...)) -> None:
-    save(SubscriptionCustomerDescriptionTable, data)
+def save_subscription_customer_description_endpoint(data: SubscriptionDescriptionBaseSchema = Body(...)) -> None:
+    create_subscription_customer_description(data.customer_id, data.subscription_id, data.description)
     delete_from_redis(data.subscription_id)
```
Mark90, line 37: I think updating the redis cache should also be moved into the domain function

Mark90 (approved):

orchestrator/domain/customer_description.py
```diff
@@ -0,0 +1,71 @@
...
+from orchestrator.utils.redis import delete_from_redis
+
+router = APIRouter()
+
+
+def get_customer_description_by_customer_subscription(
+    customer_id: str, subscription_id: UUID
+) -> SubscriptionCustomerDescriptionTable | None:
+    stmt = select(SubscriptionCustomerDescriptionTable).filter(
+        SubscriptionCustomerDescriptionTable.customer_id == str(customer_id),
```
mrijk, line 32: Why cast a str to a str? (since customer_id is already a str)

orchestrator/domain/customer_description.py
```diff
@@ -0,0 +1,71 @@
...
+    customer_id: str, subscription_id: UUID, description: str
+) -> SubscriptionCustomerDescriptionTable:
+    customer_description = SubscriptionCustomerDescriptionTable(
+        customer_id=customer_id,
+        subscription_id=subscription_id,
+        description=description,
+    )
+    db.session.add(customer_description)
+    db.session.commit()
+    delete_from_redis(subscription_id)
```
mrijk, line 48: Hm... . this might be something that would be better done in a decorator

mrijk (approved):
LGTM, 2 minor comments

orchestrator/domain/customer_description.py
tjeerddie: added a decorator and issue to update it in other places where `delete_from_redis` is used. [issue](https://github.com/workfloworchestrator/orchestrator-core/issues/615)

Issue #529, opened 2024-03-07
[Feature]: Add high priority documentation for app.py


srichmond:
### Contact Details

srichmond@es.net

### What should we build?

Please create reference documentation for the following topics on https://workfloworchestrator.org/orchestrator-core/ using the "how to write docs" information located in https://docs.google.com/document/d/10Fz98moy-pMyyk__PTR8xg4aeHuIv5VegJa99ZpzvkI/edit

- [ ] App setup
- [ ] Generator

pboers1988:
https://github.com/workfloworchestrator/orchestrator-core/blob/main/orchestrator/app.py

hanstrompert:
Current generator documentation: https://workfloworchestrator.org/orchestrator-core/architecture/application/cli/#generate.

Issue #545, opened 2024-03-07
[Feature]: Add high priority documentation for Subscriptions


srichmond:
### Contact Details

srichmond@es.net

### What should we build?

Please create reference documentation for the following topics on https://workfloworchestrator.org/orchestrator-core/ using the "how to write docs" information located in https://docs.google.com/document/d/10Fz98moy-pMyyk__PTR8xg4aeHuIv5VegJa99ZpzvkI/edit

- [ ] Lifecycle of a Customer facing service
- [ ] Product-blocks are resource-facing
- [ ] Graph of all interlinked resources
- [ ] Depends_on
- [ ] In_use_by
- [ ] Customers need to be implemented as first class citizen

pboers1988:
https://workfloworchestrator.org/orchestrator-core/architecture/application/graphql/?h=cu#overriding-customertype-and-resolvers Customers

Issue #543, opened 2024-03-07
[Feature]: Add high priority documentation for Domain Models


srichmond:
### Contact Details

srichmond@es.net

### What should we build?

Please create reference documentation for the following topics on https://workfloworchestrator.org/orchestrator-core/ using the "how to write docs" information located in https://docs.google.com/document/d/10Fz98moy-pMyyk__PTR8xg4aeHuIv5VegJa99ZpzvkI/edit

- [ ] Product Types
- [ ] Product Blocks
- [ ] Resource Types
- [ ] Product model generator - will generate all workflows, db migrations, test stubs and validations
- [ ] From_product
- [ ] From_subscription
- [ ] Validation
- [ ] Pydantic hooks
- [ ] Computed_property
- [ ] Title property
- [ ] Type-casting and serialisation
- [ ] Union types Most specific to least specific

hanstrompert:
A lot of these topics are already, more or less, documented in either the orchestrator-core documentation and/or the example orchestrator documentation. Most of the work will probably exist of trying to merge both sets of documentation into one consistent piece of documentation. Also don't forget that slide decks that were used last year for presentations or workshops may also contain useful things like nice diagrams. After all existing documentation has been merged, the missing pieces can be added.

hanstrompert:
And this document can also be of interest: GEANT 5-1 Milestone M7.3 Common NREN Network Service Product Models (https://resources.geant.org/wp-content/uploads/2023/06/M7.3_Common-NREN-Network-Service-Product-Models.pdf).

Issue #540, opened 2024-03-07
[Feature]: Add high priority documentation for Database management


srichmond:
### Contact Details

srichmond@es.net

### What should we build?

Please create reference documentation for the following topics on https://workfloworchestrator.org/orchestrator-core/ using the "how to write docs" information located in https://docs.google.com/document/d/10Fz98moy-pMyyk__PTR8xg4aeHuIv5VegJa99ZpzvkI/edit

Schema Migrations
- [ ] Schema Migrations: Use the CLI tools
- [ ] Schema Migrations: Two “heads”

pboers1988:
https://workfloworchestrator.org/orchestrator-core/getting-started/prepare-source-folder/#migrations-folder-and-local-alembic-head

PR #616, opened 2024-04-15
docs: AutoCon1 Documentation Updates

crankynetman:
This implements all of the high priority documentation that we wanted to get done in advance of AutoCon1. 

This re-uses a lot of existing docs but also adds in the following new things:

- `mkdocstrings` which allows us to pull in documentation from our actual code and use our docstrings and code examples instead of re-writing or copy/pasting them in our documentation.
- `mkdocs-macros-plugin` Allows us to add jinja (lol) templates into our docs. We can then use this to pull in examples of implementations from the example orchestrator's main branch. With this, we can show real running examples of code but not have to maintain it in two places.
- `abbr` Allows us to add abbreviations to mkdocs that show up anywhere that abbreviation is used as a tooltip.
- I also went ahead and added all of the skeleton files for all of the medium and low priority documentation that we want to add in the future. 
  - Reminder: the full list of things we wanted docuemented is [here](https://docs.google.com/document/d/15OfipvnlHG7eZnZnVQv0oviF5Qbxkc75RuTv4CT3s3k/edit). 
- I guess I added a bunch of docs or something too.
  - Domain Models
  - Forms
  - Subscriptions (Lives on the domain model docs)
  - App.py
  - Workflow Steps
  - I cleaned up and refactored a bunch of docs along the way.


crankynetman:
Closes #529, #532, #539, #540, #543, and #545.

pyproject.toml
```diff
@@ -118,6 +118,9 @@ doc = [
     "mkdocs-material[imaging]",
```
Sparrow1029, line 118: Locally, I finally got this to work by updating the `mkdocs-material[imaging]` version string to `~=9.5.17` (which installed version `9.5.18` using python 3.11 conda environment + `flit install --deps develop`).

I think that likely fixed & later versions will be installed for mkdocs plugins for fresh installs of orchestrator-core local dev environment, but wanted to make a note here about this issue, which was causing

```python
zipfile.BadZipFile: File is not a zip file
```

errors I think because of some Google CDN/Font file changes.

[Here's a link to relevant "social" plugin issue on Github](https://github.com/squidfunk/mkdocs-material/issues/6983)

docs/reference-docs/forms.md
```diff
@@ -219,7 +225,7 @@ export function autoFieldFunction(props: GuaranteedProps<unknown> & Record<strin
 
 ```
 
-### custom field example
+#### Custom Field Example
 
 example custom field to select a user by group.
```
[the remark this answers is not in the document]
crankynetman: I think that the idea was that we'd **_generally_** do the v2 docs entirely separately (I think Wouter is finding someone to do that on the SURF side) since it's a separate library. The cool thing is, if we do that, we can still use mkdocstrings if we turn on the inventory for that repo:

https://mkdocstrings.github.io/usage/#cross-references-to-other-projects-inventories


(also assumes there is a js handler, which I'm not sure on, but boy, that'd be fun to build!)

pyproject.toml
crankynetman: Good call. I think that you will only see this if you installed `mkdocs-material` after the bug was introduced, but before it was fixed, which is a ~1 week window, right? Or do you still see this with a fresh install?

Sparrow1029: Yes, it was a pretty narrow window! My intent was to make a note in case anyone else ran into it, however unlikely. Gonna resolve the the comment though

crankynetman: Really is a shame that flit doesn't have an upgrade option. When testing this, the only real workaround I could find was to upgrade with pip. But my skill level on these things is not great 😅

docs/reference-docs/cli.md
```diff
@@ -1,61 +1,40 @@
...
+      show_signature: false
+      show_root_heading: false
+      docstring_style: google
+      show_docstring_parameters: false
+      show_docstring_returns: false
+      show_root_toc_entry: false
+      show_symbol_type_toc: false
+      show_symbol_type_heading: false
+      members:
+        - downgrade
```
tjeerddie, line 29: I think this would be a more logical order for the commands
```suggestion
        - init
        - revision
        - upgrade
        - downgrade
        - heads
        - history
        - merge
        - migrate_workflows
```

docs/reference-docs/domain_models/product_types.md
```diff
@@ -0,0 +1,56 @@
...
+After defining all of the components of a Product type, you'll also need to create a database migration to properly wire-up the product in the orchestrator's database. A migration file for this example Node model looks like this:
+
+??? example "Example: `example-orchestrator/migrations/versions/schema/2023-10-27_a84ca2e5e4db_add_node.py`"
+    ```python linenums="1"
+    {% include 'https://raw.githubusercontent.com/workfloworchestrator/example-orchestrator/master/migrations/versions/schema/2023-10-27_a84ca2e5e4db_add_node.py' %}
+    ```
+
+Thankfully, you don't have to write these database migrations by hand, you can simply use the `main.py db migrate-domain-models` command that is part of the [orchestrator CLI, documented here.](../cli.md#migrate-domain-models)
+
+## Automatically Generating Product Types
```
tjeerddie, line 54: might be better to move this above `Creating Database Migrations` because you want to create products before creating migrations

docs/contributing/documentation.md
```diff
@@ -0,0 +1,9 @@
+# Writing Docs for orchestrator-core
+
+## Setting up Documentation Environment
+
+1. Install flit `pip install flit` inside of the venv you will use for orch development.
+1. setup dev env for writing docs `flit install`
+1. Write some documentation by editing the markdown files in `orchestrator-core/docs/`
+1. Build the documentation with `mkdocs build`.
+1. View the documentation by opening `orchestrator-core/site/index.html` in your browser.
```
tjeerddie, line 9: it is also possible to use `mkdocs serve`

mkdocs.yml
```diff
@@ -74,8 +99,45 @@ nav:
     - Base Application: getting-started/base.md
     - Development setup: getting-started/development.md
     - Preparing source folder: getting-started/prepare-source-folder.md
+  - Reference Documentation:
+    - TL;DR: reference-docs/tldr.md
+    - API:
+      - Rest API: reference-docs/api.md
+      - GraphQL: reference-docs/graphql.md
+    - App.py : reference-docs/app/app.md
+    # - Auth(n|z): reference-docs/auth.md
```
tjeerddie, line 108: commented out docs?

I think its better to be able to read them somewhere untill updated. (atleast for this Auth)

tjeerddie (approved):
can this be linked to issues and already close some?

mkdocs.yml
crankynetman: Agreed, however, these are all placeholders as we don't have any docs written for these yet :(

docs/contributing/documentation.md
crankynetman: Good call, this is FAR more convenient!

crankynetman:
Thanks for the review @tjeerddie ! I think I addressed all of your comments, which are all great feedback. I **_think_** I linked the issues appropriately via this comment, but I'm not 100% sure on that. 

https://github.com/workfloworchestrator/orchestrator-core/pull/616#issuecomment-2059110995

anieuwland (approved):

PR #617, opened 2024-04-17
Improve the way to add strawberry models to the schema

tjeerddie:
- fix incorrect process id federation key
- change customer to a federated type
- Improve the way to add strawberry models to the schema
- - instead of overriding or adding strawberry types to the `GRAPHQL_MODELS` with `.updates()`, you can now extend it or use a different variable using parameter `graphql_models` inside `register_graphql`.
- - [x] update documentation
- Change `SCALAR_OVERRIDE` to a parameter instead of magical dict update (same as `GRAPHQL_MODELS`)
- split Query into OrchestratorQuery and CustomerQuery class, so customers can be left out if it comes from somewhere else in a federation.

- `GRAPHQL_MODELS` is replaced by `DEFAULT_GRAPHQL_MODELS`.

### DIFF
--- a/docs/architecture/application/graphql.md
+++ b/docs/architecture/application/graphql.md
@@ -72,6 +72,8 @@ app.register_graphql(query=NewQuery)
 
 ## Adding federated types to the graphql
 
+federation introduction: https://strawberry.rocks/docs/federation/introduction
+
 Within a federation, it is possible to add orchestrator data to graphql types from other sources by extending the `DEFAULT_GRAPHL_MODELS` dictionary with your own federated classes and adding them as parameter to `app.register_graphql(graphql_models={})`. Here is an example for when instead of overriding the customers resolver, you instead use a different graphql source (know that not storing any customer data in the orchestator will make filtering and sorting unavailable and very tricky to implement):
 
 ```python
--- a/docs/architecture/application/graphql.md
+++ b/docs/architecture/application/graphql.md
@@ -27,11 +27,18 @@ app.register_graphql()
 
 ## Extending the Query and Mutation
 
+You are not able to remove resolvers from a Query, so we split the Query into 2 and merged them back for a default Query.
+Our usecase for this is that we use an external graphql source as our customers root.
+
+- `OrchestratorQuery` all resolvers except for customers.
+- `CustomerQuery` only has `customers` resolver.
+- `Query` Merges of `OrchestratorQuery` and `CustomerQuery` and serves as the default.
+
 This is an basic example of how to extend the query.
 You can do the same to extend Mutation.
 
 ```python
-from orchestrator.graphql import Query, Mutation
+from orchestrator.graphql import Query, Mutation, OrchestratorQuery
 
 
 # Queries
@@ -39,8 +46,18 @@ def resolve_new(info) -> str:
     return "resolve new..."
 
 
+# with customers.
 @strawberry.federation.type(description="Orchestrator queries")
 class NewQuery(Query):
+    other_processes: Connection[ProcessType] = authenticated_field(
+        resolver=resolve_processes,
+        description="resolve_processes used for another field",
+    )
+    new: str = strawberry.field(resolve_new, description="new resolver")
+
+# without customers.
+@strawberry.federation.type(description="Orchestrator queries")
+class NewQueryWithoutCustomers(OrchestratorQuery):
     other_processes: Connection[ProcessType] = authenticated_field(
         resolver=resolve_processes,
         description="resolve_processes used for another field",
@@ -52,6 +69,53 @@ app = OrchestratorCore(base_settings=AppSettings())
 # register SUBSCRIPTION_MODEL_REGISTRY
 app.register_graphql(query=NewQuery)
 ```
+
+## Adding federated types to the graphql
+
+Within a federation, it is possible to add orchestrator data to graphql types from other sources by extending the `DEFAULT_GRAPHL_MODELS` dictionary with your own federated classes and adding them as parameter to `app.register_graphql(graphql_models={}). Here is an example for when instead of overriding the customers resolver, you instead use a different graphql source (know that not storing any customer data in the orchestator will make filtering and sorting unavailable and very tricky to implement):
+
+```python
+import strawberry
+from sqlalchemy import select
+
+from oauth2_lib.strawberry import authenticated_field
+from orchestrator.db import db
+from orchestrator.graphql.pagination import Connection
+from orchestrator.graphql.schemas.subscription import SubscriptionInterface
+from orchestrator.graphql.types import GraphqlFilter, GraphqlSort, OrchestratorInfo
+
+
+@strawberry.federation.type(description="Customer", keys=["customer_id"])
+class Customer:
+    customer_id: str
+
+    @classmethod
+    async def resolve_reference(cls, customer_id: str) -> "Customer":  # noqa: N803
+        return Customer(customer_id=customer_id)
+
+    @authenticated_field(description="Returns subscriptions of a customer")  # type: ignore
+    async def subscriptions(
+        self,
+        info: OrchestratorInfo,
+        filter_by: list[GraphqlFilter] | None = None,
+        sort_by: list[GraphqlSort] | None = None,
+        first: int = 10,
+        after: int = 0,
+    ) -> Connection[SubscriptionInterface]:
+        from orchestrator.graphql.resolvers.subscription import resolve_subscriptions
+
+        filter_by_customer_id = (filter_by or []) + [GraphqlFilter(field="customerId", value=str(self.uuid))]  # type: ignore
+        return await resolve_subscriptions(info, filter_by_customer_id, sort_by, first, after)
+
+UPDATED_GRAPHQL_MODELS = DEFAULT_GRAPHQL_MODELS | {
+    "Customer": Customer,
+}
+
+app.register_graphql(query=OrchestratorQuery, graphql_models=UPDATED_GRAPHQL_MODELS)
+```
+
+Types that are added in this way but aren't used in a resolver, will be viewable outside of a federation inside the types in the graphql ui interface.
+Adding product or product block strawberry types to the `graphql_models` will skip their generation inside `autoregister_domain_models`. more info [here](#domain-models-auto-registration-for-graphql)
 
 ## Add Json schema for metadata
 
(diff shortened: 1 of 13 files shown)


orchestrator/graphql/schema.py
```diff
@@ -130,52 +135,64 @@ def custom_context_dependency(
...
+INTERNAL_GRAPHQL_MODELS: StrawberryModelType = {}
+
+
+def internal_graphql_model_test_logic(graphql_models: dict) -> dict:
+    """Necessary Graphql logic for unit tests since we create and delete subscription models in each test and graphql expect static models.
+
+    We create and destroy the subscription models in each test. (which makes the pydantic model technically not the same after its re-created)
+    Strawberry will error saying that a nested product block can't be resolver and if you are actually using strawberry.field, This should never happen outside of testing.
+
+    returns INTERNAL_GRAPHQL_MOEDLS if environment is 'TESTING' and graphql_models if isn't
```
tjeerddie, line 160:
```suggestion
    returns INTERNAL_GRAPHQL_MOEDLS if environment is 'TESTING' and graphql_models if it isn't
```

orchestrator/graphql/resolvers/subscription.py
```diff
@@ -66,24 +78,23 @@ def get_subscription_details(subscription: SubscriptionTable) -> SubscriptionInt
...
 
-def format_base_subscription(subscription: SubscriptionTable) -> SubscriptionInterface:
-    from orchestrator.graphql import GRAPHQL_MODELS
+def format_subscription(info: OrchestratorInfo, subscription: SubscriptionTable) -> SubscriptionInterface:
+    if _is_subscription_detailed(info):
+        return get_subscription_details(info, subscription)
 
-    strawberry_type = GRAPHQL_MODELS["subscription"]
-    return strawberry_type.from_pydantic(subscription)
+    strawberry_type = get_subscription_graphql_type(info, "subscription")
```
mrijk, line 89: The parameter value "subscription" looks a bit weird / redundant. I guess we also have a class with name Subscription?

orchestrator/graphql/resolvers/subscription.py
```diff
@@ -119,10 +130,7 @@ async def resolve_subscriptions(
     graphql_subscriptions = []
```
mrijk, line 130: I guess this is to avoid an if/else? Personally I dislike initialising variables and the overwriting them in the next line. Too many times bitten in my life with this mutable code.

orchestrator/graphql/schema.py
```diff
@@ -130,52 +135,64 @@ def custom_context_dependency(
...
+    if app_settings.ENVIRONMENT == "TESTING":
+        INTERNAL_GRAPHQL_MODELS.update(graphql_models)
+        return INTERNAL_GRAPHQL_MODELS
+    return graphql_models
 
 
 def create_graphql_router(
     query: Any = Query,
     mutation: Any = Mutation,
     register_models: bool = True,
```
mrijk, line 171: This was already there, but please don't add booleans to function defs. Just create another function, and have them call some shared code.

orchestrator/graphql/resolvers/subscription.py
mrijk: Ok... I saw some code later and this seems to be the base class. Why not have that very explicit in this code:

```
if _is_subscription_detailed(info):
     return get_subscription_details(info, subscription)
return get_subscription_base(info, subscription)
```

mrijk (approved):
Couple of minor comments, mostly about readability/understandability of the code

orchestrator/graphql/resolvers/subscription.py
tjeerddie: Its the base class and also needs to be build from the `subscription_interface` parameter that is given to `create_graphql_router`

orchestrator/graphql/resolvers/subscription.py
tjeerddie: I don't see a reason to add the else only for a variable assignment, how do/did you get bitten by this?

if the logic in the if statement was more then 2 lines, I would refactor it into a function that returns the subscriptions or an empty list.

orchestrator/graphql/schema.py
tjeerddie: will keep it in mind

orchestrator/graphql/resolvers/subscription.py
mrijk: You would be amazed I have been bitten while using mutable variables in the past 30 years ;) Mostly when the initialisation starts to move during refactoring to a different place, etc. etc. Feel free to leave it like this.

docs/architecture/application/graphql.md
```diff
@@ -47,12 +55,68 @@ class NewQuery(Query):
...
+
 
 app = OrchestratorCore(base_settings=AppSettings())
 # register SUBSCRIPTION_MODEL_REGISTRY
 app.register_graphql(query=NewQuery)
 ```
 
+## Adding federated types to the graphql
+
+Within a federation, it is possible to add orchestrator data to graphql types from other sources by extending the `DEFAULT_GRAPHL_MODELS` dictionary with your own federated classes and adding them as parameter to `app.register_graphql(graphql_models={}). Here is an example for when instead of overriding the customers resolver, you instead use a different graphql source (know that not storing any customer data in the orchestator will make filtering and sorting unavailable and very tricky to implement):
```
Mark90, line 75: have we explained the concept of federation anywhere in the docs? if not, maybe add a short line about it somewhere, and link to https://strawberry.rocks/docs/federation/introduction

Mark90, line 75:
```suggestion
Within a federation, it is possible to add orchestrator data to graphql types from other sources by extending the `DEFAULT_GRAPHL_MODELS` dictionary with your own federated classes and adding them as parameter to `app.register_graphql(graphql_models={})`. Here is an example for when instead of overriding the customers resolver, you instead use a different graphql source (know that not storing any customer data in the orchestator will make filtering and sorting unavailable and very tricky to implement):
```

docs/architecture/application/graphql.md
```diff
@@ -47,12 +55,68 @@ class NewQuery(Query):
...
+
+UPDATED_GRAPHQL_MODELS = DEFAULT_GRAPHQL_MODELS | {
+    "Customer": Customer,
+}
+
+app.register_graphql(query=OrchestratorQuery, graphql_models=UPDATED_GRAPHQL_MODELS)
+```
+
+Types that are added in this way but aren't used in a resolver, will be viewable outside of a federation inside the types in the graphql ui interface.
+Adding product or product block strawberry types to the `graphql_models` will skip their generation inside `autoregister_domain_models`. more info [here](#domain-models-auto-registration-for-graphql)
```
Mark90, line 118:
```suggestion
Adding product or product block strawberry types to the `graphql_models` will skip their generation inside `register_domain_models`. More info [here](#domain-models-auto-registration-for-graphql)
```

docs/architecture/application/graphql.md
```diff
@@ -94,18 +158,19 @@ This will result in json schema:
 
 ## Domain Models Auto Registration for GraphQL
 
-When using the `register_graphql()` function, all products in the `SUBSCRIPTION_MODEL_REGISTRY` will be automatically converted into GraphQL types.
+When using the `app.register_graphql()` function, all products in the `SUBSCRIPTION_MODEL_REGISTRY` will be automatically converted into GraphQL types.
+You are able to turn this off with `app.register_graphql(register_models=False)` but will only be able to fetch the default `Subscription` data.
```
Mark90, line 162:
```suggestion
You are able to turn this off with `app.register_graphql(register_models=False)`, but then you can only query fields from the default `SubscriptionModel`.
```

orchestrator/graphql/schemas/__init__.py
```diff
@@ -10,16 +10,10 @@
...
 
-StrawberryPydanticModel = TypeVar("StrawberryPydanticModel", bound=StrawberryTypeFromPydantic)
-StrawberryModelType = dict[str, StrawberryPydanticModel]
-
-
-GRAPHQL_MODELS: StrawberryModelType = {
+DEFAULT_GRAPHQL_MODELS: StrawberryModelType = {
     "ProductModelGraphql": ProductModelGraphql,
 }
+GRAPHQL_MODELS: StrawberryModelType = DEFAULT_GRAPHQL_MODELS
```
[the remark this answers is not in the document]
Mark90: If there is an easy change that users can make if they did use it, I'd say describe the change in the PR and let's remove it

orchestrator/graphql/schema.py
```diff
@@ -130,52 +135,64 @@ def custom_context_dependency(
...
-    class Subscription(SubscriptionInterface):
-        pass
-
-    models = dict(GRAPHQL_MODELS_INITIAL)
-    models["subscription"] = Subscription
 
     if register_models:
         models = register_domain_models(subscription_interface, existing_models=models)
-        GRAPHQL_MODELS.update(models)
+        models = internal_graphql_model_test_logic(models)
```
Mark90, line 185: Adding this fixture to `test/unit_tests/graphql/conftest.py` seem to remove the need for this. Does it work for you as well?

```python
from unittest import mock
from orchestrator.graphql.autoregistration import register_domain_models

@pytest.fixture(scope="session", autouse=True)
def fix_graphql_model_registration():
    internal_graphql_models = {}

    def patched_register_domain_models(*args, **kwargs):
        graphql_models = register_domain_models(*args, **kwargs)
        internal_graphql_models.update(graphql_models)
        return internal_graphql_models

    with mock.patch("orchestrator.graphql.schema.register_domain_models", side_effect=patched_register_domain_models):
        yield
```

I still don't understand why it's necessary, but better to have as little test-specific code in the actual code

Mark90 (approved):
Few minor doc remarks and a different way to fix the model registration during testcases. Otherwise, LGTM!

orchestrator/graphql/schema.py
tjeerddie: It has something to do with re-creating the classes per test, quick example:
```
CLASS_DICT = {}

def create_class():
    class Test:
        name: str = "1"

    CLASS_DICT['TEST'] = Test
    return Test()

test = create_class()
test2 = create_class()

print(isinstance(test, CLASS_DICT["TEST"]))  # this returns False
print(isinstance(test2, CLASS_DICT["TEST"]))  # this returns True
```

Issue #626, opened 2024-04-24
Enhance Subscription Request Response with outsideSubscriptionBoundary Property


DutchBen:
**Description:**
As a frontend developer tasked with implementing the functionality described in ticket workfloworchestrator/orchestrator-ui-library#367, I propose an enhancement to the subscription request response. Currently, when a subscription is requested, the return value includes productBlockInstances. To better facilitate frontend development and provide more context, I suggest adding a new property to each productBlockInstances object: outsideSubscriptionBoundary.

**Details:**
- **Property Name**: outsideSubscriptionBoundary
- **Type**: Boolean
- **Purpose**: This property will indicate whether the associated productBlockInstance is outside the subscription boundary.
- **Usage**: Frontend developers can utilize this property to adjust the display or behavior of elements based on their placement relative to the subscription boundary.

By adding this property, we can streamline frontend development and enhance the user experience by providing valuable information about the subscription boundary directly within the response data.

PR #627, opened 2024-04-24
Add outside subscription boundary prop to ProductBlockInstance

tjeerddie:
Closes: https://github.com/workfloworchestrator/orchestrator-core/issues/626

### DIFF
--- a/orchestrator/graphql/utils/get_subscription_product_blocks.py
+++ b/orchestrator/graphql/utils/get_subscription_product_blocks.py
@@ -20,6 +20,7 @@ class ProductBlockInstance:
     owner_subscription_id: UUID
     in_use_by_relations: list[JSON]
     product_block_instance_values: JSON
+    outside_subscription_boundary: bool
 
 
 def is_product_block(candidate: Any) -> bool:
@@ -88,6 +89,7 @@ async def get_subscription_product_blocks(
                 if included(k, v)
             ],
             in_use_by_relations=product_block.get("in_use_by_relations", []),
+            outside_subscription_boundary=product_block["owner_subscription_id"] == subscription_id,
         )
 
     product_blocks = (to_product_block(product_block) for product_block in get_all_product_blocks(subscription, tags))


orchestrator/graphql/utils/get_subscription_product_blocks.py
```diff
@@ -88,6 +89,7 @@ def included(key: str, value: Any) -> bool:
                 if included(k, v)
             ],
             in_use_by_relations=product_block.get("in_use_by_relations", []),
+            outside_subscription_boundary=product_block["owner_subscription_id"] == subscription_id,
```
DutchBen, line 92: Let's wait for @wouter1975 to discuss the exact definition of a subscription boundary

mrijk: Yes, and I would probably add that check to a separate function for clarity and because I'm pretty sure it will get more complicated in the future.

pboers1988: This seems to be "going a little too short through the corner" as the Dutch saying goes ;) Lets discuss next week.

pboers1988: Seems like this could also be calculated by using the owner_subscription_subscription_id, furthermore we still have a logic error. Outside_subscription_boundary should be `True` when !=

pboers1988 (changes requested):

orchestrator/graphql/utils/get_subscription_product_blocks.py
wouter1975: Logic should be like:
- if owner(product_block) != subscription_owner and inuse-by(product_block) != subscription_owner: outsideboundary is true

Best illustrated with an example: in case of IP service, which has a PB relation to a Serviceport (SP). The SP has a different subscription-owner thus within subscription boundary.

The SP has PB relation to a node, which has a different subscription_owner. The node is has not PB relation to a PB which has the same subscription_owner, so it falls outside the subscription boundary.

PR #634, opened 2024-04-30
Adapt type of str and int enums

mrijk:


Mark90 (approved):

orchestrator/graphql/utils/modify_class.py
```diff
@@ -0,0 +1,121 @@
...
+        return len(args) == 2 and is_type_enum(args[0]) and args[1] is type(None)
+    return False
+
+
+is_optional_int_enum = partial(is_optional_enum_of_type, is_type_enum=is_int_enum)
+is_optional_str_enum = partial(is_optional_enum_of_type, is_type_enum=is_str_enum)
+
+
+def map_type(annotation: type | None, modify_map: ModifyMap) -> type | None:
+    match first_true(modify_map, pred=lambda match: match[0](annotation)):
```
tjeerddie, line 59: do you expect extra cases? if not it could just be an if statement

tjeerddie (approved):

orchestrator/graphql/utils/modify_class.py
mrijk: The match is a nice way to do the destructuring of the return value (a tuple). Pattern often used in Scala and other languages :)

orchestrator/graphql/utils/modify_class.py
```diff
@@ -0,0 +1,142 @@
...
+    def map_field(fi: FieldInfo) -> None:
+        annotation = fi.annotation
+        if mapped_type := map_type(annotation, modify_map):
+            fi.annotation = mapped_type
+        if is_list_type(annotation):
+            orig_type = get_args(annotation)[0]
+            if mapped_type := map_type(orig_type, modify_map):
+                fi.annotation = list[mapped_type]  # type: ignore
+        if is_union_type(annotation) and not is_union_type(annotation, klass):
+            fi.annotation = reduce(or_, map_union_members(annotation, modify_map))
```
Mark90, line 95: I added this to handle unions, but this unfortunately doesn't solve the problem which lies deeper in Strawberry

Mark90: https://git.ia.surfsara.nl/netdev/automation/projects/orchestrator/-/issues/2005

PR #637, opened 2024-05-02
Add submit formpage

dmarjenburgh:
Ensures `FormNotCompleteError.meta.hasNext` data is passed to the frontend.

Updated pydantic-forms dependency
Updated reference documentation.

### DIFF
--- a/docs/reference-docs/forms.md
+++ b/docs/reference-docs/forms.md
@@ -104,7 +104,7 @@ class CreateLightPathForm(FormPage):
 
 #### Multi-step Form Input
 
-To do a multi-step form in your workflow, you simply yield multiple times and then combine the results at the end:
+To do a multistep form in your workflow, you simply yield multiple times and then combine the results at the end:
 
 ```python
 def initial_input_form_generator(product: UUIDstr, product_name: str) -> FormGenerator:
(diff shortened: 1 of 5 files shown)


Josephine-Rutten (approved):

docs/reference-docs/forms.md
```diff
@@ -131,6 +131,8 @@ def initial_input_form_generator(product: UUIDstr, product_name: str) -> FormGen
     return {**user_input.dict(), **user_input_node.dict()}
 ```
 
+For multistep forms especially, it can be useful to use the `orchestrator.forms.SubmitFormPage` class, which is just a subclass of `orchestraot.forms.FormPage` that has some metadata informing the frontend that this form is the last page in the flow so it can style the submit button differently. This is entirely optional.
```
Mark90, line 134:
```suggestion
For multistep forms especially, it can be useful to use the `orchestrator.forms.SubmitFormPage` class, which is just a subclass of `orchestrator.forms.FormPage` that has some metadata informing the frontend that this form is the last page in the flow so it can style the submit button differently. This is entirely optional.
```

Mark90 (approved):

Mark90:
Hangs on outdated checks again..

PR #647, opened 2024-05-13
Improved processes filtering and sorting

### LINKED ISSUE #648: [Bug]: Processes can't be filtered or sorted on workflow name
### Contact Details

_No response_

## Description

Searching processes with GraphQL does not allow sorting the results on `workflowName`. Also, it is hard to find a process by workflow name, since it only looks at full exact matches.

## Desired behaviour

- We should be able to sort the processes list results using `"orderBy": [{"field": "workflowName", "order": "ASC"}]`

-We should be able to filter the processes by workflow name using the `"query"` parameter and match if the query is contained in the workflow name

### Version

2.3.0rc1

All

### Relevant log output

_No response_

dmarjenburgh:
Closes #648

### DIFF
--- a/orchestrator/db/sorting/process.py
+++ b/orchestrator/db/sorting/process.py
@@ -23,6 +23,9 @@ def generic_process_relation_sort(
     join_tables = join_tables or []
 
     def sort_function(query: QueryType, order: SortOrder) -> QueryType:
+        # SQLAlchemy can't resolve the query correctly when we join the same table multiple times. A refactor for the sorting logic that
+        # ensures each table is joined once for the sort order is required. For now, we access an internal member `_setup_join` to check whether a table
+        # has already been joined
         if isinstance(query, Select):
             joined_tables = [str(t[0]) for t in query._setup_joins]
 
--- a/orchestrator/db/sorting/process.py
+++ b/orchestrator/db/sorting/process.py
@@ -4,7 +4,7 @@ from sqlalchemy import Column, Select
 from sqlalchemy.inspection import inspect
 from sqlalchemy.sql import expression
 
-from orchestrator.db import ProcessSubscriptionTable, ProcessTable, ProductTable, SubscriptionTable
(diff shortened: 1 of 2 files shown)


pboers1988 (approved):

orchestrator/db/sorting/process.py
```diff
@@ -17,22 +17,29 @@
...
-            # Can't join on CompoundSelect
-            joined_query = query
-        if order == SortOrder.DESC:
-            return joined_query.order_by(expression.desc(field))
-        return joined_query.order_by(expression.asc(field))
+            joined_tables = [str(t[0]) for t in query._setup_joins]
+
+            for table in join_tables:
+                if (tablename := getattr(table, "__tablename__", "")) and tablename not in joined_tables:
+                    query = query.join(table)
```
Mark90, line 31: What's so bad about multiple joins that we hook into 2 sqlalchemy internal variables to prevent it?

dmarjenburgh: If you repeat the same join, sqlalchemy will get confused. That can happen when we use a composite sort order, like

`order_by=[{"field":"productTag", "order":"ASC"},{"field":"productName","value":"firewall"}]`

Maybe it can be solved a different way that I overlooked, like join aliases.

Mark90: I think ideally these sort functions should return the order_by expression and additionally the table(s) that it needs to join, and that a higher level function gathers the unique tables from all sort functions and creates the joins. But that might be a pretty big change :) 

Scanning the sqlalchemy docs/code I don't suspect join aliases will solve this but maybe I'm overlooking something. Anyway, I'm also fine with a small comment in the code explaining the necessity, and possibly outlining a better solution as mentioned above - if you think it's plausible.

dmarjenburgh: That would be a quite bigger change. I'm fine with a small comment for now

dmarjenburgh: Issue created #656

Mark90 (approved):

PR #663, opened 2024-05-24
Improve subscription caching with domain model endpoint and graphql subscriptions

tjeerddie:
- change get_subscription_dict to a util function and use it for the get domain model endpoint and in graphql get_subscription_details

### DIFF
--- a/orchestrator/graphql/resolvers/subscription.py
+++ b/orchestrator/graphql/resolvers/subscription.py
@@ -20,6 +20,7 @@ from pydantic.alias_generators import to_camel as to_lower_camel
 from sqlalchemy import Select, func, select
 from strawberry.experimental.pydantic.conversion_types import StrawberryTypeFromPydantic
 
+from nwastdlib.asyncio import gather_nice
 from orchestrator.db import ProductTable, SubscriptionTable, db
 from orchestrator.db.filters import Filter
 from orchestrator.db.filters.subscription import (
@@ -33,7 +34,6 @@ from orchestrator.db.sorting.subscription import (
     sort_subscriptions,
     subscription_sort_fields,
 )
-from orchestrator.domain.base import SubscriptionModel
 from orchestrator.graphql.pagination import Connection
 from orchestrator.graphql.schemas.product import ProductModelGraphql
 from orchestrator.graphql.schemas.subscription import SubscriptionInterface
@@ -48,7 +48,7 @@ from orchestrator.graphql.utils import (
     is_querying_page_data,
     to_graphql_result_page,
 )
-from orchestrator.types import SubscriptionLifecycle
+from orchestrator.utils.get_subscription_dict import get_subscription_dict
 
 logger = structlog.get_logger(__name__)
 # Note: we can make this more fancy by adding metadata to the field annotation that indicates if a resolver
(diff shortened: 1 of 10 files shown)


orchestrator/graphql/resolvers/subscription.py
```diff
@@ -65,26 +65,34 @@
 def get_subscription_graphql_type(info: OrchestratorInfo, subscription_name: str) -> StrawberryTypeFromPydantic:
     subscription_graphql_type = info.context.graphql_models.get(subscription_name)
     if not subscription_graphql_type:
-        raise GraphQLError(message=f"No graphql type found for {subscription_name}")
+        logger.warn(message=f"No graphql type found for {subscription_name}")
```
Mark90, line 68:
```suggestion
        logger.warning(message=f"No graphql type found for {subscription_name}")
```

orchestrator/graphql/resolvers/subscription.py
```diff
@@ -65,26 +65,34 @@
...
-    base_subscription_details = base_model.from_other_lifecycle(  # type: ignore
-        subscription_details, SubscriptionLifecycle.INITIAL, skip_validation=True
-    )
-    base_subscription_details.status = subscription_details.status
-    strawberry_type = get_subscription_graphql_type(info, graphql_subscription_name(base_model.__name__))  # type: ignore
-    return strawberry_type.from_pydantic(base_subscription_details)  # type:ignore
+    subscription_dict_data, _ = await get_subscription_dict(subscription.subscription_id)
+
+    domain_model_type = SUBSCRIPTION_MODEL_REGISTRY[subscription.product.name]
+    base_model = domain_model_type.__base_type__ if domain_model_type.__base_type__ else domain_model_type
```
Mark90, line 83:
```suggestion
    base_model = domain_model_type.__base_type__ or domain_model_type
```

orchestrator/graphql/resolvers/subscription.py
```diff
@@ -65,26 +65,34 @@
...
+
+    subscription_name = graphql_subscription_name(base_model.__name__)
+    subscription_details = base_model.model_validate(subscription_dict_data, strict=False)
```
Mark90, line 86: https://docs.pydantic.dev/latest/concepts/models/#creating-models-without-validation

Mark90: Oh.

> In Pydantic V2, the performance gap between BaseModel.__init__ and BaseModel.model_construct has been narrowed considerably. For simple models, calling BaseModel.__init__ may even be faster. If you are using [model_construct()](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_construct) for performance reasons, you may want to profile your use case before assuming that [model_construct()](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_construct) is faster.

Nvm then, maybe worth a shot in the future though if some subscriptions seem particularly slow to resolve

orchestrator/utils/get_subscription_dict.py
```diff
@@ -0,0 +1,17 @@
...
+
+async def get_subscription_dict(subscription_id: UUID) -> tuple[dict, str]:
+    if cached_model := from_redis(subscription_id):
+        return cached_model  # type: ignore
+
+    subscription_model = SubscriptionModel.from_subscription(subscription_id)
+    subscription = build_extended_domain_model(subscription_model)
+    etag = to_redis(subscription)
+    if not etag:
+        etag = _generate_etag(subscription)
```
Mark90, line 16: I would say just rely on `to_redis` to do the right thing, as it checks if domain model caching is enabled

```suggestion
    etag = to_redis(subscription)
```

tjeerddie: but the domain model endpoint needs a etag currently, so thought doing it here is the best place to always have an etag

Mark90: Ah wait. `to_redis()` is actually updating the cache! This is not meant to be called on the fly, this is only done as part of workflows

tjeerddie: do we not want to call it after having it fetched once, when it isn't in the cache yet?
mostly a dev problem with importing DB's

Mark90: The step

```python
@step("Cache Subscription and related subscriptions")
def cache_domain_models(workflow_name: str, subscription: SubscriptionModel | None = None) -> State:
```

Is responsible for updating a subscription in the cache. This is called at the end of workflows.

If a subscription is not found in the cache, this means that something is wrong with it and that the workflow is stuck. When that happens, we should not have some other part of the code secretly fill the cache

tjeerddie: oke, removed it from `get_subscription_dict`

Mark90 (approved):

Issue #574, opened 2024-03-18
[Feature]: Extend websocket endpoint for subscription detail pages


wouter1975:
### What should we build?

This ticket builds upon the ticket https://github.com/workfloworchestrator/orchestrator-core/issues/502.

Tasks
- [ ] Make the `/api/ws/events` endpoint emit the message `{"name": "invalidateCache", "value": ["subscriptions", "<subscriptionId>"]}` to trigger subscription detail page invalidation.

This event should be emitted from
* The workflow step which also updates the subscription's cached domain model
* The 'set in sync' endpoint
* The update customer subscription description endpoint
* ... ?

PR #671, opened 2024-06-05
send Invalidate cache of subscription/s through websocket

tjeerddie:
- invalidate subscription/s cache in cache domain model step.
- invalidate subscription/s cache in 'set in sync' endpoint.
- invalidate subscription/s cache in create and update subscription description.
- Change invalidation to return a dict instead of a list to have it in line with frontend cache invalidation, [check the discussion](https://github.com/workfloworchestrator/orchestrator-core/pull/671#discussion_r1628930882)

Closes: https://github.com/workfloworchestrator/orchestrator-core/issues/574

### DIFF
--- a/orchestrator/utils/redis.py
+++ b/orchestrator/utils/redis.py
@@ -86,8 +86,8 @@ def delete_subscription_from_redis(
 ) -> Callable[[Callable[..., Any]], Callable[..., Any]]:
     def _delete_subscription_from_redis(func: Callable[..., Any]) -> Callable[..., Any]:
         @functools.wraps(func)
-        def wrapper(*args: tuple, **kwargs: dict[str, Any]) -> Any:
-            data = func(*args, **kwargs)
+        async def wrapper(*args: tuple, **kwargs: dict[str, Any]) -> Any:
+            data = await func(*args, **kwargs)
             key = extract_fn(data)
             delete_from_redis(key)
             return data
--- a/orchestrator/websocket/__init__.py
+++ b/orchestrator/websocket/__init__.py
@@ -23,6 +23,7 @@ from orchestrator.types import BroadcastFunc
 from orchestrator.utils.enrich_process import enrich_process
 from orchestrator.websocket.websocket_manager import WebSocketManager
 from orchestrator.workflow import ProcessStat, ProcessStatus
+from pydantic_forms.types import UUIDstr
 
 logger = get_logger(__name__)
 
@@ -102,6 +103,17 @@ async def broadcast_invalidate_cache(*cache_key: str) -> None:
     await _broadcast_event("invalidateCache", list(cache_key))
 
 
+def sync_invalidate_subscription_cache(subscription_id: UUID | UUIDstr, invalidate_all: bool = True) -> None:
+    cache_keys = {
+        f"subscription:{subscription_id}",
+    }
+    if invalidate_all:
+        cache_keys = cache_keys | {
+            "subscriptions",
+        }
+    sync_broadcast_invalidate_cache(*cache_keys)
+
+
 def send_process_data_to_websocket(
     process_id: UUID,
     data: dict,
@@ -118,7 +130,7 @@ def send_process_data_to_websocket(
         logger.debug("Broadcast process data directly to websocket_manager", process_id=str(process_id))
 
         anyio.run(websocket_manager.broadcast_data, [WS_CHANNELS.ALL_PROCESSES], data)
-        sync_broadcast_invalidate_cache("processes", str(process_id))
+        sync_broadcast_invalidate_cache("processes", f"process:{process_id}")
 
 
 async def empty_handler() -> None:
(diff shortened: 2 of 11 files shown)


Josephine-Rutten (approved):

orchestrator/workflows/steps.py
```diff
@@ -206,10 +207,12 @@ def _cache_other_subscriptions(product_block: ProductBlockModel) -> None:
     for subscription_id in cached_subscription_ids:
         subscription_model = SubscriptionModel.from_subscription(subscription_id)
         to_redis(build_extended_domain_model(subscription_model))
+        sync_broadcast_invalidate_cache(str(subscription.subscription_id))
```
dmarjenburgh, line 210: So, are we only sending a UUID to the frontend? How does the frontend know it's about a subscription?

tjeerddie: yea, good spot...

changed to `subscription:uuid` and also changed process to `process:uuid`

orchestrator/websocket/__init__.py
```diff
@@ -118,7 +130,7 @@ def send_process_data_to_websocket(
         logger.debug("Broadcast process data directly to websocket_manager", process_id=str(process_id))
 
         anyio.run(websocket_manager.broadcast_data, [WS_CHANNELS.ALL_PROCESSES], data)
-        sync_broadcast_invalidate_cache("processes", str(process_id))
+        sync_broadcast_invalidate_cache("processes", f"process:{process_id}")
```
Mark90, line 133: As far as I remember we were using arrays as cache keys, so something like

```
["processes", <uuid>]
["subscriptions", <uuid>]
```

Did that format change to colon-separated strings now? 

@DutchBen to verify from frontend perspective

dmarjenburgh: An array as cache key sounds like a bad design choice. Are you sure it's not for sending multiple cache keys in a single send?

Mark90: It's not meant for sending multiple keys, just an alternative to a delimited string. What's particularly bad about it?

I don't have a preference for one or the other, I'm fine withing a string delimited by `:` or `.` or something else. We'd just have to do this in accordance with the UI.

DutchBen: I was interpreting the message ["processes", <uuid>] as one message and the uuid as a processes uuid. I understand from @tjeerddie that it is in fact 2 messages 'processes' and a uuid that is not necessarily a process UUID. I can go either way with a slight preference for one message per message.

dmarjenburgh: Cache keys need to be checked for equality and checking arrays for equality in JS can easily lead to ambiguities, since they are not values like strings are.

Or, if you stringify them to JSON, is `["process","123"] == ["process", "123"]`?

tjeerddie: @DutchBen showed me how we use the caching on the frontend.
which is a object `{ type: 'processes' }` for list and `{ type: 'processes', id: "uuid" }` for a process detail.
I suggest we send it exactly like that, so its consistent on both frontend and backend, removing confusion on how one would add their own cache invalidation with both projects.

and do we keep the list or send one object per websocket message?

Mark90: So just to make sure we're on the same page, the messages as a whole currently look like this;

```json
{"name": "invalidateCache", "value": ["processes", "(the id here)"]}

{"name": "invalidateCache", "value": ["engineStatus"]}
```

And we'd want to change them to look like below?

```json
{"name": "invalidateCache", "value": {"type": "processes", "id": "(the id here)"}}

{"name": "invalidateCache", "value": {"type": "engineStatus"}}
```

If so: it's more explicit, I think it's great. 

> and do we keep the list or send one object per websocket message?

I think one object is fine.

Mark90: > Cache keys need to be checked for equality and checking arrays for equality in JS can easily lead to ambiguities, since they are not values like strings are.

Fair point, I didn't know this was ambiguous in JS. I agree to change to an object as above.

orchestrator/utils/redis.py
```diff
@@ -86,8 +86,8 @@ def delete_subscription_from_redis(
 ) -> Callable[[Callable[..., Any]], Callable[..., Any]]:
     def _delete_subscription_from_redis(func: Callable[..., Any]) -> Callable[..., Any]:
         @functools.wraps(func)
-        def wrapper(*args: tuple, **kwargs: dict[str, Any]) -> Any:
-            data = func(*args, **kwargs)
+        async def wrapper(*args: tuple, **kwargs: dict[str, Any]) -> Any:
+            data = await func(*args, **kwargs)
```
Mark90, line 90: Why is this changed to async?  In surf orchestrator this function is used as a decorator on sync functions. And even if we change those to async, then this will break it for other users

PR #672, opened 2024-06-05
Add functions to backfill new resource types and delete them

Josephine-Rutten:
For the orchestrator I needed to add a new resource type.  

These functions were defined but to be able to reuse them, it would be nicer to have them in the orchestrator core.


pboers1988 (approved):
Approve

PR #678, opened 2024-06-13
Pass websocket logic through in broadcast_fn

dmarjenburgh:


tjeerddie (approved):

orchestrator/websocket/__init__.py
```diff
@@ -81,8 +80,10 @@ def init_websocket_manager(settings: AppSettings) -> WebSocketManager:
     return websocket_manager
 
 
-def create_process_websocket_data(process: ProcessTable, pStat: ProcessStat) -> dict:
-    return {"process": enrich_process(process, pStat)}
+def create_process_websocket_data(process_id: UUID) -> dict:
+    process = _get_process(process_id)
+    p_stat = load_process(process)
+    return {"process": enrich_process(process, p_stat)}
```
Mark90, line 86: @tjeerddie What do you think, can we delete this?

If I'm not mistaken, this was exclusively used to broadcast the entire process data to the V1 GUI; as opposed to now that we're only sending the process ID to the V2 GUI.

And we've already dropped support for the V1 GUI in a previous core release..

orchestrator/services/process_broadcast_thread.py
```diff
@@ -66,35 +72,58 @@ def stop(self) -> None:
...
+    except Exception as e:
+        logger.exception(e)
+
+
+def _broadcast_queue_put_fn(broadcast_queue: queue.Queue, process_id: UUID) -> None:
+    # Catch all exceptions as broadcasting failure is noncritical to workflow completion
+    try:
+        websocket_data = create_process_websocket_data(process_id)
+        logger.info("Putting data in queue", data=websocket_data)
+        broadcast_queue.put((str(process_id), websocket_data))
```
Mark90, line 94: i think that the ProcessDataBroadcastThread should be changed to call `send_process_data_to_websocket()` as well, as right now it wouldn't send the cache invalidation event. (seems like a bug in the existing code)

But see my other comment first about whether or not to remove process data entirely..

orchestrator/websocket/__init__.py
tjeerddie: that is true, with dropped V1 this can be removed.

dmarjenburgh: Aaand it's gone

Mark90: 🔥

orchestrator/services/process_broadcast_thread.py
```diff
@@ -66,35 +72,58 @@ def stop(self) -> None:
...
+
+
+def _broadcast_ws_fn(process_id: UUID) -> None:
+    # Catch all exceptions as broadcasting failure is noncritical to workflow completion
+    try:
+        websocket_data = create_process_websocket_data(process_id)
+        logger.info("Send process data to ws", data=websocket_data)
+        send_process_data_to_websocket(process_id, websocket_data)
+    except Exception as e:
+        logger.exception(e)
```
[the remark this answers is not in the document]
Mark90: Closed accidentally? Logging the exception object is kind of useless

Mark90 (approved):

PR #679, opened 2024-06-13
Update processes and subscriptions list invalidation dict with { id: "LIST" }

tjeerddie:
- using only "type" it will invalidate everything with that type and adding the id list fixes that.
- - when one subscription updates, we don't want to invalidate all other subscription details.


DutchBen (approved):

pboers1988 (approved):
LGTM

PR #686, opened 2024-06-24
Refactor processes_ws unit tests

dmarjenburgh:
Update `test_processes_ws.py` which was testing that the correct process data was sent through the websocket,.
Since we have removed sending process_data through the websocket (and only send cache invalidation data), the test was waiting forever to complete.


pboers1988 (approved):

Josephine-Rutten (approved):

mrijk (commented):
LGTM

PR #690, opened 2024-06-27
Correct order_id entry for list types in migration helper

Sparrow1029:
When using the `convert_resource_type_relations_to_instance_relations` function to migrate domain models to the new `subscription_instance_relation`/`domain_model_attr` method of fetching related subscriptions via the ORM models, any `list` attributes are incorrectly given an `order_id` row value of `'0'`.

This small change uses the `row_number()` + `PARTITION_BY` ([reference](https://www.postgresqltutorial.com/postgresql-window-function/postgresql-row_number/)) method to automatically increment the value of `order_id` for insertions.


pboers1988 (approved):
LGTM

PR #695, opened 2024-07-01
Bump psycopg

pboers1988:


RichieB2B:
Is using `psycopg[binary]` best practice? You'll get a compiled `.so` of the PostgreSQL code with unknown version. The psycopg author says using a [local installation](https://www.psycopg.org/psycopg3/docs/basic/install.html#local-installation) is the preferred method.

pboers1988:
closing in favor of #696

PR #696, opened 2024-07-01
Bump psycopg to 3.2.1

RichieB2B:
This is an alternative to #695 which also changes the connection strings and fixes the unit tests.


pboers1988 (approved):

orchestrator/settings.py
```diff
@@ -48,7 +48,7 @@ class AppSettings(BaseSettings):
     EXECUTOR: str = ExecutorType.THREADPOOL
     WORKFLOWS_SWAGGER_HOST: str = "localhost"
     WORKFLOWS_GUI_URI: str = "http://localhost:3000"
-    DATABASE_URI: PostgresDsn = "postgresql://nwa:nwa@localhost/orchestrator-core"  # type: ignore
```
pboers1988, line 51: Any chance of making this change backwards compatible that will save a lot of people some pain.

pboers1988 (changes requested):

orchestrator/settings.py
RichieB2B: The whole reason for this MR is to switch to `psycopg` v3 so we can't assume the `psycopg2` module is available at all. We could convert `postgresql://` connection strings to `postgresql+psycopg://` and perhaps `postgresql+psycopg2://` as well? Perhaps issuing a warning that the conversion happened and the `DATABASE_URI` should be fixed.

pboers1988: Yes the conversion and the warning would be great, that gives some time for people to migrate

pboers1988: thanks @RichieB2B! I ran some tests internally, it seems to work 👍🏼

pboers1988 (approved):
LGTM

PR #735, opened 2024-08-19
Improve Websocket Broadcasting Stability

Mark90:
* Make `RedisBroadcast` retry in case of connection errors/timeouts
* Fix `RedisBroadcast` not releasing redis connection
* Extend logging in `BroadcastWebsocketManager` and catch exceptions
* Fixed some tests, updated pre-commit
* Bump version to 2.7.4

### DIFF
--- a/orchestrator/utils/redis.py
+++ b/orchestrator/utils/redis.py
@@ -17,9 +17,13 @@ from os import getenv
 from typing import Any, Callable
 from uuid import UUID
 
+import redis.exceptions
+from anyio import CancelScope, get_cancelled_exc_class
 from redis import Redis
 from redis.asyncio import Redis as AIORedis
 from redis.asyncio.client import Pipeline, PubSub
+from redis.asyncio.retry import Retry
+from redis.backoff import EqualJitterBackoff
 from structlog import get_logger
 
 from orchestrator.services.subscriptions import _generate_etag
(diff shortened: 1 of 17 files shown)


orchestrator/utils/redis.py
```diff
@@ -152,12 +161,24 @@ async def subscriber(self, *channels: str) -> AsyncGenerator[PubSub, None]:
...
             yield pubsub
+        except get_cancelled_exc_class():
+            # https://anyio.readthedocs.io/en/3.x/cancellation.html#finalization
+            with CancelScope(shield=True):
+                await do_teardown()
+            raise
         finally:
-            await pubsub.unsubscribe(*channels)
-            await pubsub.aclose()  # type: ignore[attr-defined]
+            await do_teardown()
```
tjeerddie, line 181: good to know of the finalization cancel scope, is this finally also nessecary?

tjeerddie (approved):

orchestrator/utils/redis.py
Mark90: Yes. It's recommended practice to use the`try/finally` construct in contextmanagers, in case we leave the context for any other reason. The cancel scope is just an addition to prevent anyio from throwing the async task into the void :)

surf-anieuwland (approved):

PR #751, opened 2024-09-30
Added broadcast invalidate status counts to processes

Georgi2704:
From issue:

https://github.com/workfloworchestrator/orchestrator-ui-library/issues/1411

### DIFF
--- a/docs/reference-docs/websockets.md
+++ b/docs/reference-docs/websockets.md
@@ -81,10 +81,10 @@ Where `name` and `value` are one of the following combinations:
 Notes:
 1. The `LIST` and `<uuid>` combinations currently mean one and the same. The reason to keep them separate is that we may want to implement throttling on the `LIST` event.
 2. The process status count event is triggered when a process:
-   * Transitions from failed state to non-failed state -> count goes down:
-     1. Suspended process is scheduled to be resumed
-     2. Inactive process is deleted from database
-   * Transitions from non-failed state to a failed state -> count goes up:
+   * Transitions to non-failed state -> count may go down:
+     1. Non-running process is scheduled to be resumed
+     2. Non-running process is deleted from database
+   * Transitions to a failed state -> count goes up:
      1. Running process finishes with an error
 
 Example of a complete message:
--- a/docs/reference-docs/websockets.md
+++ b/docs/reference-docs/websockets.md
@@ -46,3 +46,62 @@ Backend processes are executed in a threadpool and therefore access the same WSM
 To solve this there is a dedicated `ProcessDataBroadcastThread` (attached to and managed by the `OrchestratorCore` app) to perform the actual `broadcast_data()` call.
 
 The API endpoints which start/resume/abort a process, call `api_broadcast_process_data(request)` to acquire a function that can be used to submit process updates into a `threading.Queue` on which `ProcessDataBroadcastThread` listens.
+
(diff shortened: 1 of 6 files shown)


orchestrator/workflow.py
```diff
@@ -1393,6 +1404,7 @@ def _exec_steps(steps: StepList, starting_process: Process, dblogstep: StepLogFu
         # as bare exceptions are not JSON serializable
         result_to_log = step_result_process.on_failed(error_state_to_dict).on_waiting(error_state_to_dict)
         result_to_log.on_success(mutationlogger).on_failed(errorlogger).on_waiting(errorlogger)
+        result_to_log.on_failed(broadcast_failed_step)
```
Mark90, line 1407: Nice. We also need to run at after successful completion of a process though? For when a retried process transitions from failed to successful.

Which makes me think that we might as well call `broadcast_failed_step()` one level up in runwf() after it returns from `_exec_steps(steps, next_state, _logstep)`.

Mark90: I like @tjeerddie suggestion better to do it on the resume endpoint + here on step failure. Then we don't need what I suggested

tjeerddie (approved):

orchestrator/websocket/__init__.py
```diff
@@ -105,6 +105,16 @@ async def invalidate_subscription_cache(subscription_id: UUID | UUIDstr, invalid
     await broadcast_invalidate_cache({"type": "subscriptions", "id": str(subscription_id)})
 
 
+def broadcast_invalidate_status_counts() -> None:
+    """Broadcast data of the current process to connected websocket clients."""
```
Mark90, line 109: Can you update this docstring?

orchestrator/api/api_v1/endpoints/processes.py
```diff
@@ -115,6 +120,7 @@ def delete(process_id: UUID) -> None:
     if not process:
         raise_status(HTTPStatus.NOT_FOUND)
 
+    broadcast_invalidate_status_counts()
```
Mark90, line 123: I wonder why the old broadcast statement was done above the database commit. Do you have any idea perhaps @tjeerddie 

I think we should move them to below the commit, to ensure that the broadcasting action can never delay the (more important) DB action, and eliminate the (very small) possibility that the UI has done something on the broadcasted event _before_ the DB action completed.

Mark90: (This would apply to all other broadcast statements as well)

tjeerddie: yea, haven't reviewed previous MR's strict enough, I agree that they should happen after DB actions

docs/reference-docs/websockets.md
```diff
@@ -46,3 +46,62 @@ Backend processes are executed in a threadpool and therefore access the same WSM
 To solve this there is a dedicated `ProcessDataBroadcastThread` (attached to and managed by the `OrchestratorCore` app) to perform the actual `broadcast_data()` call.
 
 The API endpoints which start/resume/abort a process, call `api_broadcast_process_data(request)` to acquire a function that can be used to submit process updates into a `threading.Queue` on which `ProcessDataBroadcastThread` listens.
+
```
Georgi2704, line 49: Nice addtion !

PR #752, opened 2024-10-03
Fix a bug that did not show lists in the product blocks

Igoranze:
Issue: 2112


orchestrator/graphql/utils/get_subscription_product_blocks.py
```diff
@@ -67,21 +67,29 @@ async def get_subscription_product_blocks(
...
             return not product_block_instance_values or key in product_block_instance_values
 
         def included(key: str, value: Any) -> bool:
             return is_resource_type(value) and requested_resource_type(key) and key not in pb_instance_property_keys
 
+        def value_parser(value: Any) -> str:
+            if isinstance(value, (str, int, float, type(None))):
+                return value
+            elif isinstance(value, list):
+                return ", ".join([str(v) for v in value if is_resource_type(v)])
```
Mark90, line 82: Perhaps we can just give the list back as an array, so the frontend (or other consuming application) can decide how to format it?

```suggestion
            elif isinstance(value, list):
                return [value_parser(v) for v in value if is_resource_type(v)]
```

Unless there's a reason that this should be a string (@tjeerddie  ?)

tjeerddie: nope, you're correct, this gives a better representation of the data and the frontend can visualize it however it wants

tjeerddie (approved):

orchestrator/graphql/utils/get_subscription_product_blocks.py
Igoranze: @Mark90 and @tjeerddie fixed and pushed changes

PR #753, opened 2024-10-03
Add depth recursion to inUseBySubscriptions and dependsOnSubscriptions

tjeerddie:
This update introduces the ability to fetch subscription relations multiple layers deep by adding depth recursion to the subscription/s relations query.

Other changes:
- Move code that is independant of the graphql query loader to its own file `orchestrator/services/subscription_relations.py`.
- Fixed an issue in the `_get_in_use_by_instance_relations` and `_get_depends_on_instance_relations` functions where the `filter_statuses` argument was incorrectly applied to the subscriptions from the `subscription_ids` argument, rather than to their relations. This issue had no impact since `filter_statuses` was not in use yet.
- With the addition of the `statuses` filter in the `SubscriptionRelationFilter`, I noticed that the loader couldn't handle lists properly in combination with its caching. This is fixed by changing the list to a tuple.

More indepth description of the recursion feature:
- Use the `inUseByFilter` and dependsOnFilter with `.recurse_product_types` filter to recurse through specified product types. By default, recursion will go through the first 10 layers of the specified `recurse_product_types`. 
- To discourage inefficient queries, recursion without specifying `recurse_product_types` is not allowed, as recursing through all relations could significantly impact performance.

For `inUseBySubscriptions` and `dependsOnSubscriptions`, the following filters were added:
```python
SubscriptionRelationFilter:
    statuses: tuple[str, ...] | None = strawberry.field(default=None, description="Search by statuses")
    recurse_depth_limit: int = strawberry.field(default=10, description="the limited depth to recurse through")
    recurse_product_types: list[str] | None = strawberry.field(
        default=None, description="List of product types to recurse into"
    )
```

example query that would recurse through `Test` product type, more examples can be found in `tests/unit_tests/graphql/test_subscription_relations.py`:
```graphql
query SubscriptionRelationsRecursive {
  subscriptions {
    page {
      subscriptionId
      description
      inUseBySubscriptions(
        inUseByFilter: {statuses: ["active"], recurseProductTypes: ["Test"]}
      ) {
        page {
          subscriptionId
          description
          status
        }
      }
    }
  }
}
```
The `inUseByFilter`/`dependsOnFilter` works in combination with `filterBy` and `sortBy` to filter and sort the recursively discovered subscriptions. For example:
* Filter by product tag: `filterBy: [{ field: "tag", value: "MyTag" }]`
* Sort by tag `sortBy: [{ field: "tag", value: ASC }]`.


pboers1988 (commented):
Bumpversion LGTM

pboers1988 (approved):

orchestrator/graphql/schemas/subscription.py
```diff
@@ -123,10 +150,27 @@ async def depends_on_subscriptions(
...
+            return await info.context.core_depends_on_subs_loader.load_many(load_mapping)
+
+        subscriptions = await get_recursive_relations(
+            [self.subscription_id],
+            tuple(_depends_on_filter.statuses or ()),
+            _depends_on_filter.recurse_product_types or [],
+            _depends_on_filter.recurse_depth_limit,
+            get_depends_on_subscriptions_from_loader,
+        )
+
```
Mark90, line 173: You can move this into a local helper function in this module, something like

```py
    async def _load_recursive_relations(subscription_id: UUID, relation_filter: SubscriptionRelationFilter | None, data_loader: Callable[...]) -> list[SubscriptionTable]:
        sub_relation_filter = relation_filter or SubscriptionRelationFilter()

        async def get_subscriptions_from_loader(subscription_ids: list[UUID], filter_statuses: tuple[str, ...]) -> list[list[SubscriptionTable]]:
            load_mapping = [(sub_id, filter_statuses) for sub_id in subscription_ids]
            return await data_loader.load_many(load_mapping)

        subscriptions = await get_recursive_relations(
            [subscription_id],
            tuple(sub_relation_filter.statuses or ()),
            sub_relation_filter.recurse_product_types or [],
            sub_relation_filter.recurse_depth_limit,
            get_subscriptions_from_loader,
        )
```


And then use it here like 

```suggestion
        subscriptions = await _load_recursive_relations(self.subscription_id, depends_on_filter, info.context.core_depends_on_subs_loader)

```

orchestrator/graphql/schemas/subscription.py
```diff
@@ -36,6 +37,15 @@
 static_metadata_schema = {"title": "SubscriptionMetadata", "type": "object", "properties": {}, "definitions": {}}
 
 
+@strawberry.input(description="Filter recursion")
+class SubscriptionRelationFilter:
+    statuses: tuple[str, ...] | None = strawberry.field(default=None, description="Search by statusses")
```
Mark90, line 42: Can't this be a list?

orchestrator/graphql/loaders/subscriptions.py
```diff
@@ -1,224 +1,37 @@
...
-    ).scalars()
-    in_use_by_subs = {subscription.subscription_id: subscription for subscription in _in_use_by_subs}
+    filter_statuses: tuple[str, ...] = keys[0][1] or tuple(SubscriptionLifecycle.values())
 
-    # group (more_itertools.bucket doesn't seem to work for tuple of uuids)
-    subscription_in_use_by_ids: dict[UUID, list[UUID]] = {}
-    for relation in in_use_by_relations:
-        subscription_in_use_by_ids.setdefault(relation.depends_on_sub_id, []).append(relation.in_use_by_sub_id)
+    if len({key[1] for key in keys}) > 1:
+        raise Exception("Only one inUseByFilter.statuses can be defined")
```
Mark90, line 21: Maybe nicer to move this into a reusable function, and use some itertools-fu, so 

```py
def to_subscription_ids_and_filter_statuses(
    keys: list[tuple[UUID, tuple[str, ...] | None]], direction: Literal["dependsOn", "inUseBy"]
) -> tuple[list[UUID], tuple[str, ...]]:
    subscription_ids, filter_statuses_values = unzip(keys)
    filter_statuses = (
        one(
            unique_everseen(filter_statuses_values),
            too_long=Exception(f"{direction}Filter.statuses must be set to the same value"),
        )
        or SubscriptionLifecycle.values()
    )
    return list(subscription_ids), tuple(filter_statuses)
```


Then use like

```py
subscription_ids, filter_statuses = to_subscription_ids_and_filter_statuses(keys, "inUseBy")
```

Then we could have a few isolated unittests for `to_subscription_ids_and_filter_statuses` i.e. for these scenarios


```
Valid example - same status filter set on all load calls: [('subid1', ['active']), ('subid2', ['active']), ('subid3', ['active'])]
subscription_ids=['subid1', 'subid2', 'subid3']
filter_statuses=('active',)

Valid example - no status filter set on any load calls: [('subid1', None), ('subid2', None), ('subid3', None)]
subscription_ids=['subid1', 'subid2', 'subid3']
filter_statuses=('initial', 'active', 'migrating', 'disabled', 'terminated', 'provisioning')

Invalid example - mix of status filter and no status: [('subid1', None), ('subid2', ['active']), ('subid3', None)]
Exception: inUseByFilter.statuses must be set to the same value

Invalid example - mix of status filters: [('subid1', ['active', 'terminated']), ('subid2', ['active']), ('subid3', ['active'])]
Exception: inUseByFilter.statuses must be set to the same value
```

orchestrator/graphql/loaders/subscriptions.py
```diff
@@ -1,224 +1,37 @@
...
-        resource_type_relations = await _get_depends_on_resource_type_relations(subscription_ids, filter_statuses)
-    else:
-        resource_type_relations = []
-    # Find relations through instance hierarchy
-    instance_relations = await _get_depends_on_instance_relations(subscription_ids, filter_statuses)
-    return list(chain(resource_type_relations, instance_relations))
-
-
-async def in_use_by_subs_loader(keys: list[tuple[UUID, list[str] | None]]) -> list[list[SubscriptionTable]]:
+async def in_use_by_subs_loader(keys: list[tuple[UUID, tuple[str, ...]]]) -> list[list[SubscriptionTable]]:
```
Mark90, line 15: If I'm not mistaken the second param can also be None
```suggestion
async def in_use_by_subs_loader(keys: list[tuple[UUID, tuple[str, ...] | None]]) -> list[list[SubscriptionTable]]:
```

orchestrator/services/subscription_relations.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+
+    used_subscription_ids = used_subscription_ids or set(subscription_ids)
+    relations = await relation_fetcher(subscription_ids, filter_statuses)
+    flat_relations_list = list(flatten(relations))
+
+    def get_related_subscription_ids() -> list[UUID]:
+        return [
+            r.subscription_id
+            for r in flat_relations_list
```
Mark90, line 253: Small suggestion to rename variables; the intermediate result and the fact that something is "flat" is not that relevant. 

Also since we can already get duplicates here, we can deduplicate them while we're iterating over the flattened result

```suggestion
    _relations = await relation_fetcher(subscription_ids, filter_statuses)
    relations = list(unique_everseen(flatten(_relations), key=lambda s: s.subscription_id))

    def get_related_subscription_ids() -> list[UUID]:
        return [
            r.subscription_id
            for r in relations
```

orchestrator/services/subscription_relations.py
```diff
@@ -0,0 +1,269 @@
...
+        nested_relations = await get_recursive_relations(
+            related_subscription_ids,
+            filter_statuses,
+            recurse_product_types,
+            recurse_depth_limit,
+            current_depth=current_depth + 1,
+            used_subscription_ids=used_subscription_ids,
+            relation_fetcher=relation_fetcher,
+        )
+    return list(unique(flat_relations_list + nested_relations, lambda s: s.subscription_id))
```
Mark90, line 269: Some suggestions:
* Use `unique_everseen` instead of `unique` here to save time on sorting
* Use `itertools.chain` to avoid creating a new list
* Pass sort-key parameter as a keyword arg
```suggestion
    return list(unique_everseen(chain(relations, nested_relations), key=lambda s: s.subscription_id))
```

test/unit_tests/graphql/conftest.py
```diff
@@ -38,7 +41,7 @@ class Metadata(BaseModel):
 
 @pytest.fixture(scope="session", autouse=True)
 def fix_graphql_model_registration():
-    internal_graphql_models = {}
+    internal_graphql_models = {"ProductBlockListNestedForTestInactive": ProductBlockListNestedForTestInactiveGraphql}
```
Mark90, line 44: small comment explaining why this one is hardcoded?

test/unit_tests/graphql/test_subscription_relations.py
```diff
@@ -0,0 +1,766 @@
...
+            productId
+            name
+            description
+            productType
+            status
+            tag
+            createdAt
+            endDate
+          }
+        }
```
Mark90, line 93: Are these 2 blocks relevant for the test? I see that you're inspecting productBlockInstances in one testcase as an extra check, but I think just verifying the end result (as in, which related subscriptions are returned) is sufficient here. So maybe remove productBlocklnstances/processes (the file is long enough already :) )

test/unit_tests/graphql/test_subscription_relations.py
```diff
@@ -0,0 +1,766 @@
...
+    result = response.json()
+    subscriptions_data = result["data"]["subscriptions"]
+    subscriptions = subscriptions_data["page"]
+
+    assert "errors" not in result
+    assert subscriptions[0]["subscriptionId"] == subscription_id
+    result_depends_on_ids = [
+        subscription["subscriptionId"] for subscription in subscriptions[0]["dependsOnSubscriptions"]["page"]
+    ]
+    assert_result_ids_against_expected_ids(result_depends_on_ids, expected_depends_on_ids)
```
Mark90, line 766: (optional) Maybe see if you can generalize and parametrize a few tests here

test/unit_tests/graphql/test_subscription_relations.py
tjeerddie: there is a test for `test_single_subscription_with_processes` which uses the same query function, but will split it up so only the processes test fetches the processes.

test/unit_tests/graphql/conftest.py
tjeerddie: something like this
```suggestion
    # This block caches the "ProductBlockListNestedForTestInactive" model to avoid re-instantiation in each test case.
    # This is necessary because this product block has a self referencing property, which strawberry can't handle correctly,
    # and lead to an error expecting the `ProductBlock` strawberry type to exist.
    internal_graphql_models = {"ProductBlockListNestedForTestInactive": ProductBlockListNestedForTestInactiveGraphql}
```

Mark90: 👍

Mark90 (approved):
Works like a charm 🚀  Nice!

PR #756, opened 2024-10-07
Fix that a process resumed through callback would not broadcast events

### LINKED ISSUE #744: [Bug]: Process detail page does not update with a callback_step in ThreadPool Executor
This issue was addressed in frontend https://github.com/workfloworchestrator/orchestrator-ui-library/issues/1456 and after looking into it we found out that it is a backend issue.

> Mahammad@Geant reports that the process detail page doens't update throught the websocket connection when there are callback steps in his form. The websocket connection is established and the backend sends the correct cache invalidation requests, it appears to stop doing that when a callback step is reached.

When using the ThreadPool executor, the callback_step does not send a websocket message to the frontend for the validation step. The frontend does not update and you don't see that it has failed until refreshing.

I have not been able to reproduce the problem inside Celery executor, so I would assume the problem has to do with `ProcessDataBroadcastThread`.

## steps to reproduce

1. create a workflow with a callback_step
2. run the workflow and make sure that the validation step in the callback_step raises an error.
3. see that the frontend does not receive a websocket message to update the process detail page data.

### Version

2.7.5

_No response_

### Relevant log output

_No response_

torkashvand:
* Add `broadcast_func` to `continue_awaiting_process_endpoint` so that when using ThreadPool executor, processes resumed through callback URL will broadcast events

Closes #744

### DIFF
--- a/orchestrator/api/api_v1/endpoints/processes.py
+++ b/orchestrator/api/api_v1/endpoints/processes.py
@@ -183,7 +183,8 @@ def continue_awaiting_process_endpoint(
         raise_status(HTTPStatus.CONFLICT, "This process is not in an awaiting state.")
 
     try:
-        continue_awaiting_process(process, token=token, input_data=json_data)
+        broadcast_func = api_broadcast_process_data(request)
+        continue_awaiting_process(process, token=token, input_data=json_data, broadcast_func=broadcast_func)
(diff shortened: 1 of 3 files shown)


Mark90 (approved):

tjeerddie (approved):
LGTM

Issue #670, opened 2024-06-03
[Bug]: "Authorize" button not shown anymore on /api/docs


Mark90:
The "Authorize" button used to be on the `/api/docs` page in the top-right. This allowed to test endpoints with JWT or Basic auth.

This button has disappeared now, this may be a side effect of the recent AuthN/AuthZ refactor.

When comparing the `/api/openapi.json`, the following top-level key is missing

```json
 "securitySchemes": {
   "OIDCUser": {
     "type": "http",
     "scheme": "bearer"
   }
 }
```

Also these keys are missing on each individual endpoint 
```
 "security": [
   {
     "OIDCUser": []
   }
 ],
 "parameters": [
   {
     "name": "token",
     "in": "query",
     "required": false,
     "schema": {
       "anyOf": [
         {
           "type": "string"
         },
         {
           "type": "null"
         }
       ],
       "title": "Token"
     }
   }
 ],
```

pboers1988:
This change has happend as the Depens function no longer returns a HTTPBearer subclass

Mark90:
https://git.ia.surfsara.nl/netdev/automation/projects/orchestrator/-/issues/2079

PR #758, opened 2024-10-10
HTTPBearer token is set, Auth button not shown on /api/docs

Igoranze:
Part of https://github.com/workfloworchestrator/orchestrator-core/issues/670

This PR adds a Depends and extracts the token from the request in order to allow for HTTPBearer requests.
Part of the missing HTTPBearer request was that the /api/docs auth button was not shown.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ dependencies = [
     "orjson==3.10.7",
     "psycopg2-binary==2.9.9",
     "pydantic[email]~=2.7.4",
-    "pydantic-settings~=2.4.0",
+    "pydantic-settings~=2.5.2",
     "python-dateutil==2.8.2",
     "python-rapidjson>=1.18,<1.20",
     "pytz==2024.1",


orchestrator/security.py
```diff
@@ -36,8 +33,10 @@
 )
 
 
-async def authenticate(request: Request) -> OIDCUserModel | None:
-    return await request.app.auth_manager.authentication.authenticate(request)
+async def authenticate(
+    request: Request, token: Annotated[str | None, Depends(TokenExtractor())] = None
```
torkashvand, line 37:
```suggestion
    request: Request, token: Annotated[str | None, Depends(HttpBearerExtractor.extract)] = None
```

Why not this? am I missing something? So we don't need your PR on the oauth-lib?

because I think we had this `HttpBearerExtractor` in oauth-lib exactly for this. Hope I am not wrong

torkashvand: of course, you have to instantiate `HttpBearerExtractor` or you can make it a static method in base class

pyproject.toml
```diff
@@ -61,7 +61,8 @@ dependencies = [
     "typer==0.12.5",
     "uvicorn[standard]~=0.32.0",
     "nwa-stdlib~=1.7.3",
-    "oauth2-lib~=2.2.0",
+#    "oauth2-lib~=2.2.0",
+    "oauth2-lib @ git+https://github.com/workfloworchestrator/oauth2-lib.git@auth-b-shown",
```
Mark90, line 65: To be changed to oauth2-lib 2.3.0 when it's done

Mark90 (approved):
LGTM!

torkashvand (approved):
LGTM!

PR #772, opened 2024-10-28
Revert pydantic-forms 1.1.1 to 1.1.0 summary not shown

Igoranze:
Reverting the pydantic forms to old version because there were some issues with the workflow summary not being shown correctly

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.8.1rc4
+current_version = 2.8.0rc5
 commit = False
 tag = False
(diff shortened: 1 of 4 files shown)


mrijk (approved):
LGTM, but we need a follow-up ticket so we can unblock this pinning

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.8.0rc4
+current_version = 2.8.1rc4
```
tjeerddie, line 2: we dont have a 2.8.0 release yet, this should be 2.8.0rc5

tjeerddie (approved):

PR #777, opened 2024-11-06
Upgrade strawberry graphql to 0.246.2 and above

tjeerddie:
upgrading to above `0.246.2` to be consistent with nwa-stdlib package. could also change it to above `0.236.0` to be consistent with the [strawberry breaking import change](https://github.com/strawberry-graphql/strawberry/releases/tag/0.236.0).

I found the upgrading problem in strawberry version `0.233.0`, where they refactor the federation schema to use strawberry typing instead of converting strawberry types to graphql types https://github.com/strawberry-graphql/strawberry/pull/3525

## Upgrade problem description

The problem was that the `autoregistration` unnecessarily created a type based on the output of the `pydantic_wrapper` function, which already generates the required Strawberry type.

This led to the same Strawberry type being created twice at runtime, but from different locations:
 1. Directly in the `pydantic_wrapper`, resulting in `strawberry.experimental.pydantic.object_type.YourStrawberryType`.
 2. Indirectly in `autoregistration`, which created a type using the result of `pydantic_wrapper`, resulting in `orchestrator.graphql.autoregistration.YourStrawberryType` 
  - with the function `create_block_strawberry_type` or `create_subscription_strawberry_type`.

Previously Strawberry handled this duplication because it already needed to convert Strawberry types to GraphQL types, so we never noticed that we created the type twice.

Strawberry refactored `strawberry.federation.schema` to use Strawberry types directly (removing GraphQL type conversion) in version `0.233.0`.
This implementation does not unintentionally fix our duplicate types and resulted in the runtime error: `Union type _Entity can only include type YourStrawberryType once`.

Removing the type create in `autoregistration` fixed this problem.

### DIFF
--- a/orchestrator/graphql/autoregistration.py
+++ b/orchestrator/graphql/autoregistration.py
@@ -109,6 +109,9 @@ def create_subscription_strawberry_type(strawberry_name: str, model: type[Domain
     return type(strawberry_name, (pydantic_wrapper(base_type),), {})
 
 
+EXISTING_MODELS = set()
+
+
 def add_class_to_strawberry(
     model_name: str,
     model: type[DomainModel],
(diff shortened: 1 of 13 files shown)


Igoranze (approved):

orchestrator/graphql/autoregistration.py
```diff
@@ -17,9 +17,9 @@
 import strawberry
```
tjeerddie, line 17: There is a flaky test `test_product_blocks_has_previous_page`

Mark90 (approved):
Nice investigation 👍

PR #786, opened 2024-12-11
Add documentation on backfilling new resource types in existing subscriptions

kvklink:


kvklink:
Do we want the ORM style to be part of this documentation as well?

hanstrompert:
I think it can be useful for people to see how you can use the ORM for this kind of things. But I would like the opinion of a couple of core developers as well, so I invited @Mark90 and @tjeerddie to review as well.

docs/architecture/product_modelling/backfilling.md
```diff
@@ -0,0 +1,77 @@
+# Backfilling Existing Subscriptions
+When updating a product block that already exists in your orchestrator, it could be the case that new attributes are
+added or removed. When removing resource types from a `ProductBlock`, the database migration generator is able to pick
```
Mark90, line 3: > the database migration generator

This refers to the `migrate-domain-models` command, right?

Maybe it's better to write it as such so people can find it (I can't find it when I search "Database migration generator")

Mark90 (approved):
Nice! Left a few comments/nitpicks.

I guess it shouldn't be too hard for the migrate-domain-models command to allow specifying a backfill value, might be a nice "Good First Issue" feature ticket

@hanstrompert I think migrations should always be written in plain SQL so that they are versioned in time. Otherwise a future breaking change in the SQLalchemy models, such as renaming a column, would break old migrations that expect the column by it's old name - but which the SQLAlchemy model will try to access by it's new name. (which, if data migrations are correctly linked to schema migrations, will not exist at that point in the migration chain)

PR #791, opened 2025-01-09
Add optional version control for subscriptions and subscription customer descriptions

tjeerddie:
- add optional version validation to the first form (`_generate_modify_form`) of modify workflows that validates if the version is the same as then the one in the db, when a version is given.
- add similar optional version validation for subscription customer description modifies that validates if the version is the same before executing the update.
- add and update unit tests.

### DIFF
--- a/orchestrator/api/api_v1/endpoints/processes.py
+++ b/orchestrator/api/api_v1/endpoints/processes.py
@@ -25,7 +25,7 @@ from fastapi.param_functions import Body, Depends, Header
 from fastapi.routing import APIRouter
 from fastapi.websockets import WebSocket
 from fastapi_etag.dependency import CacheHit
-from more_itertools import chunked, first
+from more_itertools import chunked
 from sentry_sdk.tracing import trace
 from sqlalchemy import CompoundSelect, Select, select
 from sqlalchemy.orm import defer, joinedload
@@ -40,7 +40,6 @@ from orchestrator.db.filters import Filter
 from orchestrator.db.filters.process import filter_processes
 from orchestrator.db.sorting import Sort, SortOrder
 from orchestrator.db.sorting.process import sort_processes
-from orchestrator.domain.base import update_subscription_version
 from orchestrator.schemas import (
     ProcessIdSchema,
     ProcessResumeAllSchema,
@@ -64,7 +63,6 @@ from orchestrator.services.settings import get_engine_settings
 from orchestrator.settings import app_settings
 from orchestrator.types import JSON, State
 from orchestrator.utils.enrich_process import enrich_process
-from orchestrator.utils.get_subscription_dict import get_subscription_dict
 from orchestrator.websocket import (
     WS_CHANNELS,
     broadcast_invalidate_status_counts,
@@ -129,18 +127,6 @@ def delete(process_id: UUID) -> None:
     broadcast_process_update_to_websocket(process.process_id)
 
 
-async def validate_subscription_version(workflow_data: dict[str, Any]) -> None:
-    subscription_id = workflow_data.get("subscription_id")
-    new_version = workflow_data.get("version")
-
-    if new_version is not None and subscription_id:
-        subscription, _ = await get_subscription_dict(subscription_id)
-        current_version = subscription.get("version")
-        if current_version > new_version:
-            raise_status(HTTPStatus.BAD_REQUEST, f"Stale data ({current_version} < {new_version})")
-        update_subscription_version(subscription_id, new_version)
-
-
 @router.post(
     "/{workflow_key}",
     response_model=ProcessIdSchema,
@@ -154,8 +140,6 @@ async def new_process(
     user: str = Depends(user_name),
 ) -> dict[str, UUID]:
     broadcast_func = api_broadcast_process_data(request)
-    if json_data and (first_wf_data := first(json_data, None)):
-        await validate_subscription_version(first_wf_data)
     process_id = start_process(workflow_key, user_inputs=json_data, user=user, broadcast_func=broadcast_func)
 
     return {"id": process_id}
--- a/orchestrator/api/api_v1/endpoints/processes.py
+++ b/orchestrator/api/api_v1/endpoints/processes.py
@@ -25,7 +25,7 @@ from fastapi.param_functions import Body, Depends, Header
 from fastapi.routing import APIRouter
 from fastapi.websockets import WebSocket
 from fastapi_etag.dependency import CacheHit
-from more_itertools import chunked
+from more_itertools import chunked, first
 from sentry_sdk.tracing import trace
 from sqlalchemy import CompoundSelect, Select, select
 from sqlalchemy.orm import defer, joinedload
@@ -40,6 +40,7 @@ from orchestrator.db.filters import Filter
 from orchestrator.db.filters.process import filter_processes
 from orchestrator.db.sorting import Sort, SortOrder
 from orchestrator.db.sorting.process import sort_processes
+from orchestrator.domain.base import update_subscription_version
 from orchestrator.schemas import (
     ProcessIdSchema,
     ProcessResumeAllSchema,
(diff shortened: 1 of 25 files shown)


orchestrator/db/models.py
```diff
@@ -624,6 +626,7 @@ class SubscriptionMetadataTable(BaseModel):
         index=True,
     )
     metadata_ = mapped_column("metadata", pg.JSONB(), nullable=False)  # type: ignore
+    version = mapped_column(Integer, nullable=False, server_default="1")
```
tjeerddie, line 629: accidentally left this in. metadata does not need version control, there is no api endpoint to update it. we use a workflow to update it.
```suggestion
```

orchestrator/api/api_v1/endpoints/processes.py
```diff
@@ -127,19 +129,33 @@ def delete(process_id: UUID) -> None:
...
 )
-def new_process(
+async def new_process(
     workflow_key: str,
     request: Request,
     json_data: list[dict[str, Any]] | None = Body(...),
     user: str = Depends(user_name),
 ) -> dict[str, UUID]:
     broadcast_func = api_broadcast_process_data(request)
+    if json_data and (first_wf_data := first(json_data, None)):
```
tjeerddie, line 157: should move this to `_generate_modify_form`

orchestrator/workflows/utils.py
```diff
@@ -140,6 +141,15 @@ def subscription_validator(cls, subscription_id: UUID) -> UUID:
 
             return subscription_id
 
+        @model_validator(mode="after")
+        def version_validator(self) -> Self:
+            if self.version is not None:
+                subscription = db.session.get(SubscriptionTable, self.subscription_id)
+                current_version = subscription.version  # type: ignore
+                if current_version > self.version:
+                    raise ValueError(f"Stale data ({current_version} < {self.version})")
```
mrijk, line 150: This seems to contradict the if statement

orchestrator/workflows/utils.py
```diff
@@ -157,11 +169,15 @@ def create_initial_input_form_generator(state: State) -> FormGenerator:
         user_input = yield _generate_modify_form(workflow_target, workflow_name)
 
         subscription = SubscriptionTable.query.get(user_input.subscription_id)
+        if user_input.version:
+            subscription.version = user_input.version
+            db.session.commit()
```
Mark90, line 174: I don't think we should update the database version in a form step..

Mark90 (changes requested):

orchestrator/domain/customer_description.py
```diff
@@ -53,8 +53,18 @@ async def create_subscription_customer_description(
...
+    description: str,
+    created_at: datetime | None = None,
+    version: int | None = None,
 ) -> SubscriptionCustomerDescriptionTable:
+    if version is not None:
+        if version < customer_description.version:
+            raise ValueError(
+                f"Stale data: given version ({version}) is lower than the current version ({customer_description.version})"
+            )
+        customer_description.version = version
```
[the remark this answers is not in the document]
tjeerddie: the ticket requirement was allowing the same or a newer version and I do believe we should update it to the given version, the trigger still adds a 1 to the updated version.

orchestrator/migrations/versions/schema/2025-01-08_4c5859620539_add_version_column_to_subscription.py
```diff
@@ -0,0 +1,64 @@
...
+
+    conn.execute(
+        sa.text(
+            """
+CREATE OR REPLACE FUNCTION increment_version()
+RETURNS TRIGGER AS $$
+BEGIN
+    NEW.version := COALESCE(NEW.version, OLD.version) + 1;
+    RETURN NEW;
+END;
```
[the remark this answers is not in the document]
tjeerddie: tricky to find but it is somewhere in here: https://www.postgresql.org/docs/current/sql-createtrigger.html

NEW refers to the updated row and `OLD` is the row before the update.

orchestrator/domain/customer_description.py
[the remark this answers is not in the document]
tjeerddie: only allowing exact version

orchestrator/utils/errors.py
```diff
@@ -74,6 +74,14 @@ class InconsistentDataError(ProcessFailureError):
     pass
 
 
+class StaleDataError(ValueError):
+    """The version of the update payload is older than the version in the database."""
```
surf-anieuwland, line 78: Comment is out of date compared to the message

test/unit_tests/api/test_processes.py
```diff
@@ -531,3 +531,51 @@ def test_create_process_reporter(test_client, fastapi_app, oidc_user, reporter,
...
+    new_version = db.session.get(SubscriptionTable, generic_subscription_1).version
+    assert new_version == 3
+
+
+def test_new_process_version_check(test_client, generic_subscription_1):
+    version = 2
+
+    response = test_client.post(
+        "/api/processes/modify_note",
+        json=[{"subscription_id": generic_subscription_1, "version": version}, {"note": "test note"}],
```
surf-anieuwland, line 552: I'm not sure I understand what is tested and happening here. Is this test trying to check that we can set the initial version? We don't want that.

test/unit_tests/api/test_subscription_customer_descriptions.py
```diff
@@ -83,6 +84,59 @@ def test_update(seed, test_client):
     assert new_desc == customer_description.description
 
 
+def test_update_with_version(seed, test_client):
```
surf-anieuwland, line 87: Can we have a test for updating without a version?

surf-anieuwland (approved):
Looks good to me. Added some comments for clarification and one more unit test

test/unit_tests/api/test_processes.py
tjeerddie: this checks that with sending version, it goes through the process and has an updated version

test/unit_tests/api/test_subscription_customer_descriptions.py
tjeerddie: that already exists above this

orchestrator/graphql/mutations/customer_description.py
```diff
@@ -30,20 +30,22 @@
...
 
 
 async def resolve_upsert_customer_description(
-    customer_id: str, subscription_id: UUID, description: str
+    customer_id: str, subscription_id: UUID, description: str, version: int | None = None
 ) -> CustomerDescription | NotFoundError | MutationError:
     try:
-        customer_description = await upsert_customer_description(customer_id, subscription_id, description)
+        customer_description = await upsert_customer_description(customer_id, subscription_id, description, version)
+    except ValueError as error:
```
Mark90, line 47: This is still matching for ValueError

Mark90 (approved):
Added 1 minor remark, otherwise LGTM!

test/unit_tests/api/test_subscription_customer_descriptions.py
surf-anieuwland: That's for processes/workflows right? This is a customer description endpoint which doesn't go through those

test/unit_tests/api/test_subscription_customer_descriptions.py
```diff
@@ -83,6 +84,59 @@ def test_update(seed, test_client):
     assert new_desc == customer_description.description
```
tjeerddie, line 84:
```suggestion
    assert new_desc == customer_description.description
    assert 2 == customer_description.version
```

test/unit_tests/api/test_subscription_customer_descriptions.py
tjeerddie: did add version assert in the existing unit test: https://github.com/workfloworchestrator/orchestrator-core/pull/791#discussion_r1914781532

PR #799, opened 2025-01-15
#2194 refactor to more abstract validation workflow

Igoranze:
The orchestrator-core contains a cronjob that validates workflows.
Refactored the validation code so that it can be used outside of the workflow and more abstract so that there can be a filter applied on product_type

### DIFF
--- a/orchestrator/migrations/versions/schema/2025-10-19_4fjdn13f83ga_add_validate_product_type_task.py
+++ b/orchestrator/migrations/versions/schema/2025-10-19_4fjdn13f83ga_add_validate_product_type_task.py
@@ -12,8 +12,6 @@ import sqlalchemy as sa
 from alembic import op
 
 # revision identifiers, used by Alembic.
-
-# revision identifiers, used by Alembic.
 revision = "4fjdn13f83ga"
 down_revision = "4c5859620539"
 branch_labels = None
--- a/orchestrator/migrations/versions/schema/2025-10-19_4fjdn13f83ga_add_validate_product_type_task.py
+++ b/orchestrator/migrations/versions/schema/2025-10-19_4fjdn13f83ga_add_validate_product_type_task.py
@@ -18,28 +18,24 @@ branch_labels = None
 depends_on = None
 
 
-workflows = [
-    {
-        "name": "task_validate_product_type",
-        "target": "SYSTEM",
-        "description": "Validate all subscriptions of Product Type",
-        "workflow_id": uuid4(),
-    }
-]
+workflow = {
+    "name": "task_validate_product_type",
+    "target": "SYSTEM",
+    "description": "Validate all subscriptions of Product Type",
+    "workflow_id": uuid4(),
+}
 
 
 def upgrade() -> None:
     conn = op.get_bind()
-    for workflow in workflows:
-        conn.execute(
-            sa.text(
-                "INSERT INTO workflows VALUES (:workflow_id, :name, :target, :description, now()) ON CONFLICT DO NOTHING"
-            ),
-            workflow,
-        )
+    conn.execute(
+        sa.text(
+            "INSERT INTO workflows VALUES (:workflow_id, :name, :target, :description, now()) ON CONFLICT DO NOTHING"
+        ),
+        workflow,
+    )
 
 
 def downgrade() -> None:
     conn = op.get_bind()
-    for workflow in workflows:
-        conn.execute(sa.text("DELETE FROM workflows WHERE name = :name"), {"name": workflow["name"]})
+    conn.execute(sa.text("DELETE FROM workflows WHERE name = :name"), {"name": workflow["name"]})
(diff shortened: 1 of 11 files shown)


orchestrator/services/subscriptions.py
```diff
@@ -670,3 +671,13 @@ def filter_instance_ids_on_productblock(path_to_block: str) -> None:
         filter_instance_ids_on_subscription()
 
     return subscription
+
+
+def get_subscriptions_on_product_table_in_sync(in_sync: bool = True) -> list[SubscriptionTable]:
+    if app_settings.VALIDATE_OUT_OF_SYNC_SUBSCRIPTIONS:
```
Mark90, line 677: This setting was meant to toggle the behavior of the `validate_subscriptions` workflow so it should remain there. Probably you can add another function `get_subscriptions_on_product_table()` )both in-sync and out-of-sync subs) for the workflow to call

orchestrator/services/workflows.py
```diff
@@ -42,3 +48,33 @@ def _add_filter(stmt: Select) -> Select:
...
+        default = TARGET_DEFAULT_USABLE_MAP[Target.SYSTEM]
+        usable_when = WF_USABLE_MAP.get(workflow, default)
+
+        if subscription.status in usable_when and (
+            product_type_filter is None or subscription.product.product_type == product_type_filter
+        ):
+            json = [{"subscription_id": str(subscription.subscription_id)}]
+
+            # against circular import
+            from orchestrator.services.processes import get_execution_context
```
Mark90, line 73: Preferably we do local imports at the start of a function

Mark90 (approved):

orchestrator/services/subscriptions.py
Igoranze: Thanks, reverted back to the original call but now with the two different functions

orchestrator/services/workflows.py
Igoranze: Unfortunately I can't do that in this file, because there seems to be an circular import.

`get_execution_context` imports `orchestrator/services/workflows.py`. And thus if I add this import at the top, it will import itself via a loop and crash.

orchestrator/migrations/versions/schema/2025-10-19_4fjdn13f83ga_add_validate_product_type_task.py
```diff
@@ -0,0 +1,45 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "4fjdn13f83ga"
+down_revision = "4c5859620539"
+branch_labels = None
+depends_on = None
+
+
+workflows = [
```
tjeerddie, line 21: no reason to use a list with a single workflow

orchestrator/schedules/validate_subscriptions.py
```diff
@@ -34,29 +37,19 @@
...
-                validate_func = get_execution_context()["validate"]
-                validate_func(validation_workflow, json=json)
-        else:
+        if not system_product_workflows:
             logger.warning(
                 "SubscriptionTable has no validation workflow",
                 subscription=subscription,
                 product=subscription.product.name,
             )
+            break
```
tjeerddie, line 53: I believe this should be `continue` to loop through the next iter instead of breaking the entire loop
```suggestion
            continue
```

orchestrator/migrations/versions/schema/2025-10-19_4fjdn13f83ga_add_validate_product_type_task.py
Igoranze: Fixed

orchestrator/schedules/validate_subscriptions.py
Igoranze: TY, nice catch. should be continue indeed

orchestrator/services/workflows.py
```diff
@@ -42,3 +48,43 @@ def _add_filter(stmt: Select) -> Select:
 
 def get_workflow_by_name(workflow_name: str) -> WorkflowTable | None:
     return db.session.scalar(select(WorkflowTable).where(WorkflowTable.name == workflow_name))
+
+
+def get_system_product_workflows_for_subscription(
+    subscription: SubscriptionTable,
+) -> list:
+    return [workflow.name for workflow in subscription.product.workflows if workflow.target == Target.SYSTEM]
```
tjeerddie, line 56: people can add more tasks to a product instead of having only `validate`

orchestrator/migrations/versions/schema/2025-10-19_4fjdn13f83ga_add_validate_product_type_task.py
```diff
@@ -0,0 +1,41 @@
+"""Validate Product Type.
```
tjeerddie, line 1: do a bumpversion to 2.9.2rc2

orchestrator/services/workflows.py
Igoranze: Got it,
This function is specifically only for getting the system workflows, in start_validation_workflow_for_workflows we do a check to make sure these are only validation workflows

PR #816, opened 2025-02-10
Remove workaround for supporting postgres vers <14

karimbenbourenane:
Removes the no-longer-needed workaround in `TSVisitorQuery` that was supporting `postgresql` versions older than `postgresql@14`.


pboers1988 (approved):
LGTM

kvklink (approved):

PR #818, opened 2025-02-10
Abort Task now sends a 'invalidateCache' message.

misterbox:


Mark90:
Tested it and works 🚀

PR #824, opened 2025-02-11
[376] build and push container image action fails

gstewart86:
`docker build` now installs library from local filesystem rather than pulling from pypi

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ WORKDIR /tmp/orchestrator-core
 COPY . .
 RUN pip install .
 RUN pip uninstall setup-tools -y
+RUN rm -rf /tmp/*
 
 RUN useradd orchestrator
 USER orchestrator
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,15 @@
 FROM python:3.11-slim
 
-ARG VERSION
-
 ENV PIP_ROOT_USER_ACTION=ignore
 
 RUN apt-get update && apt-get install --yes git
 
-RUN pip install pip --upgrade
-RUN pip install orchestrator-core==${VERSION}
-RUN pip uninstall setup-tools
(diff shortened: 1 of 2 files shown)


pboers1988 (approved):
LGTM

pboers1988 (approved):
LGTM

Dockerfile
```diff
@@ -1,14 +1,16 @@
...
-RUN pip install pip --upgrade
-RUN pip install orchestrator-core==${VERSION}
-RUN pip uninstall setup-tools
+RUN pip install --upgrade pip
+
+WORKDIR /tmp/orchestrator-core
+COPY . .
+RUN pip install .
+RUN pip uninstall setup-tools -y
+RUN rm -rf /tmp/*
```
Mark90, line 13: @gstewart86 I'm not sure if this cleanup is doing what's intended; building the image locally shows it becomes 850MB whereas before it was 459MB

Mark90: On a clean repository (without any cache folders) the size is only 486MB, which is then mainly due to the `.git` folder.

A fix for this could be to use a temporary build stage and copy the installed packages back, but depending on how much you've worked with docker files that might be a lot of work. It's also fine to leave it as it is!

gstewart86: interesting -- I tested both with and without this `rm` command and the size remained 751MB
```
REPOSITORY                                             TAG              IMAGE ID       CREATED         SIZE
<none>                                                 <none>           748871b39325   2 hours ago     751MB
<none>                                                 <none>           9e3a5d7819b5   2 hours ago     751MB
```
So I assumed the size of the library was negligible, but now that I review the size of the directory
```
310M    ./orchestrator-core
```
that is definitely false. So, good observation. Either way though I think your suggestion is the right way to go, Ill update shortly 👍

PR #825, opened 2025-02-11
Remove reimports

### LINKED ISSUE #428: Remove pydantic_forms.types re-imports
In #308 the `pydantic-forms` library was added as a dependency, and form types in orchestrator-core were removed. To remain backwards compatible we import most of `pydantic_forms.types` in `orchestrator.types`.

These imports should be removed, and a migration script should be added to rewrite the imports in orchestrator implementations. For this script one can take a look at the `migrate_pydantic_2x.py` script which performs similar rewrite actions.

# Tasks
* [ ] Remove pydantic_forms imports in `orchestrator.forms` that are not used in core
* [ ] Add a migration script for orchestrator implementors

kvklink:
Closes #428

!! Breaking change

* Removes repeat import statements in orchestrator.types
* Replaces import statements to orchestrator.types that should instead point to pydantic_forms.types
* Adds a migration script in devtools to automatically perform this change in a user's own orchestrator

### DIFF
--- a/orchestrator/devtools/populator.py
+++ b/orchestrator/devtools/populator.py
@@ -26,8 +26,8 @@ import structlog
 from more_itertools import first, first_true
 
 from nwastdlib.url import URL
-from orchestrator.types import State
 from pydantic_forms.types import InputForm as LegacyInputForm
+from pydantic_forms.types import State
 
 
 class JSONSubSchema(TypedDict, total=False):
--- a/orchestrator/devtools/scripts/migrate_pydantic_forms_imports.py
+++ b/orchestrator/devtools/scripts/migrate_pydantic_forms_imports.py
@@ -1,0 +1,61 @@
+"""Helper script to rewrite import statements in your orchestrator.
+
+Since types have been externalised in `pydantic_forms`, they were re-imported in `orchestrator.types` for backwards
+compatibility. These import statements have been removed, and therefore need to be updated in orchestrator
+implementations.
+"""
+
+import sys
+from pathlib import Path
+
+from shared import migrate, move_import
+
+
+def migrate_file(f: Path) -> bool:
+    imports = {
+        "JSON": move_import(f, "JSON", "orchestrator.types", "pydantic_forms.types"),
+        "AcceptData": move_import(f, "AcceptData", "orchestrator.types", "pydantic_forms.types"),
+        "AcceptItemType": move_import(f, "AcceptItemType", "orchestrator.types", "pydantic_forms.types"),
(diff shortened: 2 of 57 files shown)


orchestrator/devtools/populator.py
```diff
@@ -26,8 +26,8 @@
 from more_itertools import first, first_true
 
 from nwastdlib.url import URL
-from orchestrator.types import State
 from pydantic_forms.types import InputForm as LegacyInputForm
+from pydantic_forms.types import State
```
pboers1988, line 30: Lelluk

pboers1988 (changes requested):
Please add some docs and call the migrate script migrate to 3.0

orchestrator/devtools/scripts/migrate_pydantic_forms_imports.py
```diff
@@ -0,0 +1,61 @@
...
+
+Since types have been externalised in `pydantic_forms`, they were re-imported in `orchestrator.types` for backwards
+compatibility. These import statements have been removed, and therefore need to be updated in orchestrator
+implementations.
+"""
+
+import sys
+from pathlib import Path
+
+from shared import migrate, move_import
```
Mark90, line 11: This should be an absolute import, at least the script is failing on my machine when I'm trying to run it with 

```
python -m orchestrator.devtools.scripts.migrate_pydantic_forms_imports.py <dir>
```

Mark90:
```
ModuleNotFoundError: No module named 'shared'
```

orchestrator/devtools/scripts/shared.py
```diff
@@ -0,0 +1,108 @@
...
+    except FileNotFoundError:
+        print(f"{cmd} failed: could not find executable in the current venv")
+    return False
+
+
+def migrate(target_dir: Path, migrate_file: Callable[[Path], bool]) -> bool:
+    abs_path = str(target_dir.resolve())
+
+    def run_tools() -> bool:
+        return run_tool("ruff", "--fix", abs_path) and run_tool("black", "--quiet", abs_path)
```
Mark90, line 87: As it turns out this existing code was invalid because ruff's interface changed over time; the correct command is now `ruff check --fix`

orchestrator/devtools/populator.py
kvklink: Wat je zegt ben je zelf

pboers1988 (approved):

PR #830, opened 2025-02-12
build and push container image action fails

### LINKED ISSUE #376: build and push container image action fails most of the time
The `Build and push container image` action can fail because it needs the package build by the `Upload Python Package` action that might not be finished yet. Please fix this race condition.

gstewart86:
closes #376 
Removes the dependence between the container build and python build jobs -- building directly from source
separate build and final image to minimize image size
shrink image size by 50MB

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,22 @@
-FROM python:3.11-slim
+# syntax=docker/dockerfile:1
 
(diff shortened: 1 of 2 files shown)


Mark90 (approved):
Nice 🚀

Mark90 (approved):

PR #831, opened 2025-02-12
Adds a `FILTER_BY_MODE` config option to allow for exact or partial s…

misterbox:
…tring matching.

### DIFF
--- a/docs/reference-docs/search.md
+++ b/docs/reference-docs/search.md
@@ -53,13 +53,6 @@ Note that:
 * Ordering of words does not matter (unless it is a Phrase)
 * Search words cannot contain the characters `|-*():"` as they are part of the search query grammar
 
-### Config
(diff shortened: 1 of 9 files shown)


docs/reference-docs/search.md
```diff
@@ -53,6 +53,13 @@ Note that:
 * Ordering of words does not matter (unless it is a Phrase)
 * Search words cannot contain the characters `|-*():"` as they are part of the search query grammar
 
+### Config
```
Mark90, line 56: Ah sorry, this is not actually related to search.

A better place would be in the `graphql.md` file, probably in a section called something like `##  Behavior of filterBy`

Mark90 (approved):
Minor remark regarding the docs, LGTM otherwise 🚀

docs/reference-docs/search.md
misterbox: Fixed!

Issue #817, opened 2025-02-10
[Feature]: Design/Discuss a way to store user input as soon as a celery task is submitted


pboers1988:
### What should we build?

When using the Celery executor, the user input is not persisted in the database until the first step has executed. Please design and implement a change where the user_input is stored in the database and the loaded from db when the first step gets executed. At this point in time the user_input is transported to the worker through redis.

**Implementation suggestion**

Change the flow of code by persisting user_input on a one to many relationship to the ProcessTable. This is needed as a workflow may have multiple user input moments

The table should look like this:

- FK relationship to processes
- user_input JSONB
- step-name

### Relevant pseudo code

```shell

```

PR #832, opened 2025-02-12
Added an Initial Input state table

pboers1988:
Closes https://github.com/workfloworchestrator/orchestrator-core/issues/817

This table is used to store and persist the initial state of a process before it is submitted to celery or the user_input of a process when it is resumed.


orchestrator/db/models.py
```diff
@@ -81,6 +83,20 @@ def process_result_value(self, value: datetime | None, dialect: Dialect) -> date
         return value.astimezone(timezone.utc) if value else value
 
 
+class InputStateTable(BaseModel):
+    __tablename__ = "input_state"
```
Mark90, line 87:
```suggestion
    __tablename__ = "input_states"
```

(almost) All other table names are plural as well

orchestrator/db/models.py
```diff
@@ -81,6 +83,20 @@ def process_result_value(self, value: datetime | None, dialect: Dialect) -> date
...
+
+    class InputType(enum.Enum):
+        user_input = "user_input"
+        initial_state = "initial_state"
+
+    input_state_id = mapped_column(UUIDType, primary_key=True, server_default=text("uuid_generate_v4()"), index=True)
+    process_id = mapped_column("pid", UUIDType, ForeignKey("processes.pid"), nullable=False)
+    input_state = mapped_column(pg.JSONB(), nullable=False)  # type: ignore
+    input_time = mapped_column(UtcTimestamp, server_default=text("current_timestamp()"), nullable=False)
+    input_type = mapped_column(Enum(InputType), nullable=False)
```
Mark90, line 97: I think we haven't used enums in the db models so far for a reason. However I have forgotten that reason..

orchestrator/services/input_state.py
```diff
@@ -0,0 +1,65 @@
...
+        .filter(InputStateTable.input_type == input_type)
+        .order_by(InputStateTable.input_time.asc())
+    ).first()
+
+    if res:
+        return res
+    raise ValueError(f"No input state for pid: {process_id}")
+
+
+def _store_input_state(process_id: UUID, input_state: dict[str, Any] | list[dict[str, Any]], input_type: str) -> None:
```
Mark90, line 46: Prefixing with underscore is meant to signal that functions are only meant to be used within the same module, which is not the case here.

Also you can use `Literal` to restrict the input type string values

```suggestion
def store_input_state(process_id: UUID, input_state: dict[str, Any] | list[dict[str, Any]], input_type: Literal["input_state", "user_input"]) -> None:
```

orchestrator/services/input_state.py
```diff
@@ -0,0 +1,65 @@
...
+from typing import Any
+from uuid import UUID
+
+from sqlalchemy import select
+
+from orchestrator.db import db
+from orchestrator.db.models import InputStateTable
+
+
+def _retrieve_input_state(process_id: UUID, input_type: str) -> InputStateTable:
```
Mark90, line 22: Similar suggestion here as for `_store_input_state`

orchestrator/db/models.py
tjeerddie: having two places to maintain the enum, always requiring a migration to add new types.
I like to keep enums only inside the codebase

tjeerddie (approved):

orchestrator/db/models.py
Mark90: When I check the SA docs and google in general I don't see any mentions of it being a "bad" thing, so let's just try it out. If it ends up being bad we can change it back to a VARCHAR

orchestrator/services/input_state.py
```diff
@@ -0,0 +1,69 @@
...
+from typing import Any, Literal
+from uuid import UUID
+
+from sqlalchemy import select
+
+from orchestrator.db import db
+from orchestrator.db.models import InputStateTable
+
+
+def retrieve_input_state(process_id: UUID, input_type: str) -> InputStateTable:
```
Mark90, line 22: `input_type` can also be a Literal

orchestrator/services/input_state.py
```diff
@@ -0,0 +1,69 @@
...
+    res: InputStateTable | None = db.session.scalars(
+        select(InputStateTable)
+        .filter(InputStateTable.process_id == process_id)
+        .filter(InputStateTable.input_type == input_type)
+        .order_by(InputStateTable.input_time.asc())
+    ).first()
+
+    if res:
+        return res
+    raise ValueError(f"No input state for pid: {process_id}")
```
Mark90, line 43: So when deploying this, users should make sure that there are no incomplete workflows, because those can't be resumed. We should at least document this in the release, and perhaps consider making the migration check for and error on incomplete workflows?

Mark90 (approved):
🚀

PR #837, opened 2025-02-13
#782 add RBAC to running workflows - update decorators

### LINKED ISSUE #782: [Feature]: Add Role based access control to running workflows - update decorators
Refinement of ticket after discussion on 11-2:

## Requirements:
* Ability to assign privileges per group
* Ability to check privileges halfway a workflow
* Make it possible to check authorization at each input type step
  * like  suspend, callback input steps (iow when a user/proces changes the workflow into running state)
 
* Dropped requirement: ability to change the minimal rights in the UI. This should be a handled by the auth implementation of the application

## Suggested solution

- [ ]  Extend the stepdecorator to call an optional function, which is performing the isAllowed function by the attribute `workflow_name`, `workflow_step_name` (optional) and `user`. If this function returns true the workflow (step) can be executed.
  * Update the @update, @inputstep and @workflow decorators to include the isAllowed function.
 
**In other tickets:** 
* extend the graphql calls for populating the `Start` dropdown menu items, to include a state flag (eg isAllowed bool) to rendering the workflow item in normal or disabled state. 
#821 
* extend the graphql calls for populating the `Tasks` dropdown menu items, to include a state flag (eg isAllowed bool) to rendering the workflow item in normal or disabled state. 
#822 

---

### Contact Details

srichmond@es.net

### What should we build?

We would like to restrict access to running workflows to role based access control.  

This should be pluggable such that a user can integrate with their own selected auth solution.

### Summary
At the moment the task page has restricted access, while some to the workflows could be run/inspected by anybody, as the risk is minimal to none. Therefore it would be nice to be specific on defining execution rights per wf/task.
The new task button requires the access to the /orchestrator/processes/create/task. By adding an extra field to the task or workflow DB record, a minimal authorization level (MAL) can be specified.

- /orchestrator/processes/create/task/no_risk
- /orchestrator/processes/create/task/low_risk
- /orchestrator/processes/create/task/medium_risk
- /orchestrator/processes/create/task/major_risk
- /orchestrator/processes/create/task/service_affecting
- /orchestrator/processes/create/task/admin_only

By default all workflows and tasks are medium_risk.
When populating the new task dropdown, the user scope is matched with the MAL, so only allowed tasks are shown.

- [ ]  add MAL field to the workflows table in coreDB
  - [ ]  consider an extra field active/inactive per workflow?   
  - [ ]  remove column deleted_at

DB view of workflows

image

- [ ] add endpoint to change the MAL per workflow/task in the FE (Metadata pages), only allow valid fields (ENUM) --> ticket to be specced

image

- [ ] add logic that validates in the BE that requested workflow/task is allowed to be executed by the user based on the added field in the workflows table.
- [ ] Filter the `new task` dropdown with the valid task matching the users profile

Sparrow1029:
Extend the @step decorator to call an optional function, which is performing the isAllowed function by the attribute workflow_name, workflow_step_name (optional) and user. If this function returns true the workflow (step) can be executed.

Update the @inputstep and various `@*_workflow` decorators to include the isAllowed function.

Resolves #782


pboers1988 (changes requested):
@Sparrow1029 I feel like we probably need some tests to cover this feature. Looks okay to me!

orchestrator/workflow.py
```diff
@@ -178,12 +180,18 @@ def form_generator(state: State) -> FormGenerator:
...
+    return True
+
+
 def make_workflow(
     f: Callable,
     description: str,
     initial_input_form: InputStepFunc | None,
     target: Target,
     steps: StepList,
+    authorize_callback: Callable[[OIDCUserModel | None], bool] | None = None,
```
Mark90, line 194: Could this work too?

```suggestion
    authorize_callback: Callable[[OIDCUserModel | None], bool] = allow,
```

test/unit_tests/cli/test_generate_code.py
```diff
@@ -119,6 +120,7 @@ def get_expected_to_actual_migration_rev_ids(actual_folder, expected_folder):
     return dict(zip(expected_revision_ids, actual_revision_ids))
 
 
+@pytest.mark.skipif(os.getenv("LOCAL") in ("t", "true", "1", "True"), reason="Double quotes cause failure locally")
```
Mark90, line 123: I ran into this too the other day, this is because `ruff` added a [new rule](https://docs.astral.sh/ruff/rules/unquoted-type-alias/) in 0.10.0, and we use ruff for formatting the generated code. I think if you update ruff in your local virtualenv that it should pass

Mark90 (approved):
Left a few nitpicks but looks great!

test/unit_tests/cli/test_generate_code.py
Sparrow1029: @Mark90 Thanks for that info! Fixed up some areas to address your comments. I'm having difficulty finding where to add more tests since this is at the workflow level. I could potentially add one more test for the "resume" functionality? Or maybe create some authorization tests with OICDUserModel being passed to the `test_client`?

What are your thoughts?

orchestrator/workflow.py
Sparrow1029: Unfortunately linting tests are failing. I think in order to keep the user-facing interface with a default of `None`, the signature has to be `| None = None` to make mypy happy

test/unit_tests/cli/test_generate_code.py
Mark90: I think you might be responding to Peter's comment from before you added the testcase?

Even though there is "only" 1 added test, it covers the essence of the functionality so it's fine with me!

pboers1988 (approved):

Mark90:
Note: this PR only implements RBAC for the `@workflow` decorator. Implementation for `@inputstep` will be done in a new ticket.

PR #841, opened 2025-02-13
294 detect usage of basemodel

karimbenbourenane:
The `update` method for `SUBSCRIPTION_MODEL_REGISTRY` has been overridden to make sure that non-ProductBlockModel fields are not of type `pydantic.BaseModel`. Will throw a `TypeError` if directly inheriting from `pydantic.BaseModel`.

### DIFF
--- a/docs/reference-docs/app/scaling.md
+++ b/docs/reference-docs/app/scaling.md
@@ -108,6 +108,112 @@ The code above sets our local Celery instance (which initializes the workflows) 
 going to be used by the orchestrator-core. Without this code, the orchestrator-core would be only aware of a limited
 set of workflows that are part of orchestrator-core itself.
 
+### An example implementation
+
+When using Celery and Websockets you can use the following example and change it to your needs.
+
+```python
+"""This module contains functions and classes necessary for celery worker processes.
+
+When the orchestrator-core's thread process executor is specified as "celery", the `OrchestratorCore` FastAPI
+application registers celery-specific task functions and `start_process` and `resume_process` now defer to the
+celery task queue.
+
+Celery's task queue enables features like nightly validations by providing a task queue and workers to execute
+workflows that are all started in parallel, which would crash a single-threaded orchestrator-core.
+
+The application flow looks like this when "celery" is the executor (and websockets are enabled):
+
+- FastAPI application validates form input, and places a task on celery queue (create new process).
+  - If websockets are enabled, a connection should exist already b/t the client and backend.
+- FastAPI application begins watching redis pubsub channel for process updates from celery.
+- Celery worker picks up task from queue and begins executing.
+- On each step completion, it publishes state information to redis pubsub channel.
+- FastAPI application grabs this information and publishes it to the client websocket connection.
+
+A celery worker container will start by calling this module instead of `main.py` like so:
+```sh
+celery -A esnetorch.celery_worker worker -E -l INFO -Q new_tasks,resume_tasks,new_workflows,resume_workflows
+```
+
+* `-A` points to this module where the worker class is defined
+* `-E` sends task-related events (capturable and monitorable)
+* `-l` is the short flag for --loglevel
+* `-Q` specifies the queues which the worker should watch for new tasks
+
+See https://workfloworchestrator.org/orchestrator-core/reference-docs/app/scaling for more information.
+"""
+
+from uuid import UUID
+
+from celery import Celery
+from celery.signals import worker_shutting_down
+from orchestrator.db import init_database
+from orchestrator.domain import SUBSCRIPTION_MODEL_REGISTRY
+from orchestrator.services.tasks import initialise_celery
+from orchestrator.types import BroadcastFunc
+from orchestrator.websocket import broadcast_process_update_to_websocket, init_websocket_manager
+from orchestrator.websocket.websocket_manager import WebSocketManager
+from orchestrator.workflows import ALL_WORKFLOWS
+from structlog import get_logger
+
+from esnetorch.settings import app_settings
+
+
+logger = get_logger(__name__)
+
+
+def process_broadcast_fn(process_id: UUID) -> None:
+    # Catch all exceptions as broadcasting failure is noncritical to workflow completion
+    try:
+        broadcast_process_update_to_websocket(process_id)
+    except Exception as e:
+        logger.exception(e)
+
+
+class OrchestratorWorker(Celery):
+    websocket_manager: WebSocketManager
+    process_broadcast_fn: BroadcastFunc
+
+    def on_init(self) -> None:
+        init_database(app_settings)
+
+        # Prepare the wrapped_websocket_manager
+        # Note: cannot prepare the redis connections here as broadcasting is async
+        self.websocket_manager = init_websocket_manager(app_settings)
+        self.process_broadcast_fn = process_broadcast_fn
+
+        # Load the products and load the workflows
+        import esnetorch.products  # noqa: F401  Side-effects
+        import esnetorch.workflows  # noqa: F401  Side-effects
+
+        logger.info(
+            "Loaded the workflows and products",
+            workflows=len(ALL_WORKFLOWS.values()),
+            products=len(SUBSCRIPTION_MODEL_REGISTRY.values()),
+        )
+
+    def close(self) -> None:
+        super().close()
+
+
+celery = OrchestratorWorker(
+    f"{app_settings.SERVICE_NAME}-worker", broker=str(app_settings.CACHE_URI), include=["orchestrator.services.tasks"]
+)
+
+if app_settings.TESTING:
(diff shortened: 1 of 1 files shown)


orchestrator/domain/base.py
```diff
@@ -1403,6 +1403,22 @@ def db_model(self) -> SubscriptionTable:
...
+            if type(value):
+                for field_name, field_type in value._non_product_block_fields_.items():
+                    if field_type is BaseModel or field_type.__mro__[1] is BaseModel:
+                        bad_field_names.add(field_name)
+        if bad_field_names:
+            raise TypeError(
+                f"SubscriptionModel fields {bad_field_names} should not have type {BaseModel} or inherit directly from {BaseModel}. "
+                f"If this field was meant to be a Product Block, inherit from {ProductBlockModel} instead."
+            )
+        super().update(m, **kwargs)
```
Mark90, line 1419: I like the approach here! 

While we only intend to use `update()` someone could also do single key-value assignments (through `__setitem__(self, key, value)`) so it would be nice to extract the validation and reuse it for both.

orchestrator/domain/base.py
```diff
@@ -1403,6 +1403,22 @@ def db_model(self) -> SubscriptionTable:
         return self._db_model
 
 
+class SubscriptionModelRegistry(dict[str, type[SubscriptionModel]]):
+    def update(self, m, /, **kwargs):
+        bad_field_names = set()
+        for key, value in (m | kwargs).items():
+            if type(value):
```
Mark90, line 1410: I think this always evaluates to true? Perhaps this can be changed to validate that the value is a `type[SubscriptionModel]`?

orchestrator/domain/base.py
```diff
@@ -1403,6 +1403,22 @@ def db_model(self) -> SubscriptionTable:
...
+                for field_name, field_type in value._non_product_block_fields_.items():
```
Mark90, line 1411: This should probably recurse into `_product_block_fields` and then for those also validate that there is no BaseModel in the `_non_product_block_fields`

karimbenbourenane: I thought that's what I was doing here. Can you elaborate further? I think what you are saying is that for each `field` in `_non_product_block_fields_` I need to check if there are additional `_non_product_block_fields_` attributes for each of those, and recurse in that way? Is that what you mean?

orchestrator/domain/base.py
karimbenbourenane: It should always evaluate to true, but user could do an update like:
```
SUBSCRIPTION_MODEL_REGISTRY.update({
  None: ExampleSubscription,
  "Example 1": ExampleSubscription,
  "Example 2": Example Subscription,
})
```

I did this to protect the user from trying to do a bad update, since None (I think) would be a valid key in this dictionary. What do you think?

orchestrator/domain/base.py
karimbenbourenane: If I'm wrong and `None` is not a valid key I will get rid of that unnecessary check.

orchestrator/domain/base.py
karimbenbourenane: If I'm wrong and None is not a valid key I will get rid of that unnecessary check.

orchestrator/domain/base.py
karimbenbourenane: Thanks for this comment, I'll abstract that out so that this logic is used in either case.

orchestrator/domain/base.py
Mark90: Ah I see. Well, the keys here resemble the product name, setting this to None would be an interesting design choice :) It's not something a user should do and IMO there's no need to check for it either

orchestrator/domain/base.py
Mark90: If I'm not mistaken, an example situation that wouldn't be detected is this

```python
class BadBlock(BaseModel):
  str_field: str
 

class GoodBlock(ProductBlockModel):
  str_field: str
  sub_block_bad: BadBlock  # <-- this is not detected


class RootBlock(ProductBlockModel):
  int_field: int
  sub_block_good: GoodBlock
  sub_block_bad: BadBlock  # <-- this is detected
  

class MyProduct(SubscriptionModel):
   root_block: RootBlock
```

So to be able to detect that `BadBlock` inherits from BaseModel, we'd need to recurse through `MyProduct._product_block_fields_` and then inspect `GoodBlock._non_product_block_fields_`

Mark90:
> Maybe we can have a zoom call at some point and I can go over what I've tried and someone with more metaprogramming experience than I can help me understand what I need to do to get the linter to accept this pattern I'm trying to implement.

I've also tried adding type annotations for this for the last hour and it's very annoying. I'm fine with a `# type: ignore` on this :)

orchestrator/domain/base.py
karimbenbourenane: Ok, will fix.

karimbenbourenane:
Indeed it is annoying, ill go ahead and try with the typeignore

orchestrator/domain/base.py
karimbenbourenane: Having another look I realized that the example I gave where a key was `None` wasn't the same as having a `None` value, which is what was being checked previously. 
But in any case we shouldn't need to check for that since, as mentioned, the implementer will construct the Subscription Model registry and if they do silly things like using `None` keys and None values it will be very clear that it's broken, so I think it makes sense to just remove the non-`None` check.
As for validating that the values are `type[SubscriptionModel]`, I think this would be getting taken care of by static type checking, no? I can't see a common easy way for someone to sneak a BaseModel value into this dictionary "by accident" without running into the type checking that's already being used via type hints. I didn't see strong precedent for including runtime type checking for methods and attributes that have type hints added so it seems right to not check it at run time here.
So for now I'm getting rid of the value validation and not do the runtime type checking of the value here. If you think the runtime type checking really needs to go in at this point, let me know and I'll add it in.

karimbenbourenane: I would like to commit my changes, but I'm currently stuck fighting with mypy. No matter how I type hint my update method, including adding the three override annotated  update methods from the super class, mypy will complain about with things like 
```
orchestrator/domain/base.py:1419:5: error: Overloaded function implementation does not accept all possible arguments of signature 1  [misc]
orchestrator/domain/base.py:1419:5: error: Overloaded function implementation does not accept all possible arguments of signature 2  [misc]
orchestrator/domain/base.py:1419:5: error: Overloaded function implementation does not accept all possible arguments of signature 3  [misc]
```
I'm going to going to fight with it a little bit more and see if I can make progress to at least get the pre-commit hooks to let me push my changes.

karimbenbourenane:
```python
class DomainModel(BaseModel):
    _product_block_fields_: ClassVar[dict[str, Any]]
    _non_product_block_fields_: ClassVar[dict[str, type]]
```
I'm trying to understand why `_product_block_fields_` is a `dict` with values of `Any` instead of `ProductBlockModel`. I'm guessing there's some circular dependency that can't be worked out since a `ProductBlockModel` inherits from `DomainModel`, but shouldn't it be possible to use strings for type hinting here? There's probably a good reason why its done this way but I'm interested to find out why.

orchestrator/domain/base.py
karimbenbourenane: I think my problem will be solved by abandoning `update` and doing just `__setitem__`. I'll give my reasoning in a comment response further down.

orchestrator/domain/base.py
karimbenbourenane: Ok, I think I've gotten a key insight here that took some time for me to discover. I thought, since the only way that the original SUBSCRIPTION_MODEL_REGISTRY dictionary was being interacted with by the user implementting `Orchestrator` was through use of the `update` method of class `dict`, that I would put the logic for this validation in the `update` method. Then there was mention about `__setitem__` and my first thought was "oh great now I have to go find all the different methods I need to override to make this robust". After learning a lot more about how the `dict` and `MutableMap` classes work it finally clicked that `__setitem__` fixes all my problems at once since it's used in direct assignment like `d[k] = v`, `update`, `setdefault`, dictionary unpacking, initialization with keyword arguments, and probably many others im forgetting.

orchestrator/domain/base.py
karimbenbourenane: I will use this more comprehensive test in the domain tests, thank you!

orchestrator/domain/base.py
Mark90: I think both need to be overridden because `update` does not reuse any of the internal `__setitem__` logic. But I think the main problem with `update` was the typing, which we agreed elsewhere can be ignored?

karimbenbourenane:
Note that it looks like this pull request got automatically closed due to a force push. A new one was created here https://github.com/workfloworchestrator/orchestrator-core/pull/885

PR #849, opened 2025-02-19
811 extend product endpoint to include all product blocks

### LINKED ISSUE #811: Extend `product` endpoint to include all product blocks
### Summary
For the UI Metadata product page we would like to show all product blocks belonging to a product (unique list/no duplicates), and not only the root product block.

For a product like shown below, all 3 product blocks should be returned.

Image

Current GQL call:

```
query MetadataProducts {
        products {
            page {
                productId
                name
                description
                tag
                createdAt
                productType
                status
                productBlocks {
                    name
                }
                fixedInputs {
                    name
                    value
                }
                endDate
            }
            pageInfo {
                endCursor
                hasNextPage
                hasPreviousPage
                startCursor
                totalItems
                sortFields
                filterFields
            }
        }
    }
```

misterbox:
Adds 'allPbNames' field to query the name of a product and it's nested, dependent product blocks.

Closes #811


Mark90:
Can you add a brief description to the PR and a "closes #<ticket_id>" line?

orchestrator/graphql/schemas/product.py
```diff
@@ -51,6 +51,29 @@ async def subscriptions(
...
         return await resolve_subscriptions(info, filter_by_with_related_subscriptions, sort_by, first, after)
 
+    @strawberry.field(description="Returns list of all nested productblock names")  # type: ignore
+    async def all_pb_names(self) -> list[str]:
+
+        model = get_original_model(self, ProductTable)
+        product_blocks: list[ProductBlockTable] = model.product_blocks
+        names: list[str] = [model.name]
+
+        def get_all_pb_names(product_blocks: list[ProductBlockTable]) -> list[str]:
```
Mark90, line 61: We often use generator functions to avoid creating intermediate lists, which especially adds up on recursive functions. 

For example:

```python
def get_results() -> list[int]:
    results = []
    for i in range(3):
        results.append(i)
    results.extend(range(5, 8))
    return results

get_results()
```

Can be rewritten to 

```python
from typing import Iterable

def get_results() -> Iterable[int]:
    for i in range(3):
        yield i
    yield from range(5, 8)

list(get_results()
```

orchestrator/graphql/schemas/product.py
```diff
@@ -51,6 +51,29 @@ async def subscriptions(
...
+            for product_block in product_blocks:
+                result.append(product_block.name)
+
+                if product_block.depends_on:
+                    result.extend(get_all_pb_names(product_block.depends_on))
+
+            return result
+
+        if product_blocks:
+            names.extend(get_all_pb_names(product_blocks))
```
Mark90, line 73: It's also fine to pass `model.product_blocks` directly here. And the `if product_blocks` check is not needed because if the list is empty, names will be extended with an empty list.

orchestrator/graphql/schemas/product.py
```diff
@@ -51,6 +51,29 @@ async def subscriptions(
...
+
+        return names
```
Mark90, line 75: Would it make sense to sort the resulting names to guarantee a consistent order?

Mark90 (approved):
Few small remarks, overall LGTM!

Mark90 (changes requested):
Changing my review to 'request changes' just to prevent accidental merging

orchestrator/graphql/schemas/product.py
misterbox: Neat, I didn't know about generator functions. I believe I have resolved this.

orchestrator/graphql/schemas/product.py
misterbox: Done

orchestrator/graphql/schemas/product.py
misterbox: Done

misterbox:
> Can you add a brief description to the PR and a "closes #<ticket_id>" line?

Done!

Mark90 (approved):
LGTM!

orchestrator/graphql/schemas/product.py
```diff
@@ -51,6 +51,24 @@ async def subscriptions(
         filter_by_with_related_subscriptions = (filter_by or []) + [GraphqlFilter(field="product", value=self.name)]
         return await resolve_subscriptions(info, filter_by_with_related_subscriptions, sort_by, first, after)
 
+    @strawberry.field(description="Returns list of all nested productblock names")  # type: ignore
+    async def all_pb_names(self) -> list[str]:
+
+        model = get_original_model(self, ProductTable)
+        names: list[str] = [model.name]
```
tjeerddie, line 58: do we need to add `model.name` here? 
I would expect the root product block name to also come from the `model.product_blocks`

Mark90: Good point, I missed that, I think that's currently referring to the name of the product?

orchestrator/graphql/schemas/product.py
```diff
@@ -51,6 +51,24 @@ async def subscriptions(
...
+
+        def get_all_pb_names(product_blocks: list[ProductBlockTable]) -> Iterable[str]:
+            for product_block in product_blocks:
+                yield product_block.name
+
+                if product_block.depends_on:
+                    yield from get_all_pb_names(product_block.depends_on)
+
+        names.extend(get_all_pb_names(model.product_blocks))
```
mrijk, line 67: Not a big deal, but by reshuffling some code and an initial yield of model.name might get rid of this extend. (I've become oversensitive for mutable variables since I started programming in Lisp, Scheme and Clojure ;) )

orchestrator/graphql/schemas/product.py
```diff
@@ -51,6 +51,24 @@ async def subscriptions(
...
+        names.sort()
+
+        return names
```
mrijk, line 70: You could end up with 'sorted(get_all_pb_names)' instead

mrijk: Also, if you would extract the logic from the actual strawberry resolver, you might have a smaller test. I think you are now mostly testing strawberry instead of the actual coce

mrijk (approved):
Couple of remarks

orchestrator/graphql/schemas/product.py
mrijk: 'return sorted(get_all_pb_names(model.product_blocks))'

orchestrator/graphql/schemas/product.py
misterbox: My mistake. I misunderstood the original request. This should be fixed.

orchestrator/graphql/schemas/product.py
misterbox: Removed.

orchestrator/graphql/schemas/product.py
misterbox: Great idea. I'll look into this.

PR #856, opened 2025-02-24
Ensure all redis clients retry on connection timeouts/errors

Mark90:
This should reduce/resolve errors in the step `Cache Subscription and related subscriptions`.

Retrying is done using [EqualJitterBackoff](https://redis-py.readthedocs.io/en/stable/backoff.html#redis.backoff.EqualJitterBackoff) which is basically exponential backoff with some random deviation to prevent the thundering herd problem.  It was already used for the websocket broadcasting but it's now applied to all redis clients.

Env setting `REDIS_RETRY_COUNT` can be used to change the amount of retries or even disable it by setting to `0`.


pboers1988 (approved):
LGTM

misterbox (approved):

PR #877, opened 2025-03-06
[#724] Exclude subscriptions_search materialized view from Alembic au…

### LINKED ISSUE #724: Manage subscriptions_search view through SQLalchemy
`subscriptions_search` is a materialized view created manually through a migration. When running alembic's [revision](https://workfloworchestrator.org/orchestrator-core/reference-docs/cli/?h=autogenerate#orchestrator.cli.database.revision) it creates the following migration

```python
def upgrade() -> None:
    # ### commands auto generated by Alembic - please adjust! ###
    op.create_table(
        "subscriptions_search",
        sa.Column("subscription_id", sqlalchemy_utils.types.uuid.UUIDType(), nullable=False),
        sa.Column("tsv", sqlalchemy_utils.types.ts_vector.TSVectorType(), nullable=True),
        sa.ForeignKeyConstraint(
            ["subscription_id"],
            ["subscriptions.subscription_id"],
        ),
        sa.PrimaryKeyConstraint("subscription_id"),
    )
    op.create_index(
        op.f("ix_subscriptions_search_subscription_id"), "subscriptions_search", ["subscription_id"], unique=False
    )
    # ### end Alembic commands ###


def downgrade() -> None:
    # ### commands auto generated by Alembic - please adjust! ###
    op.drop_index(op.f("ix_subscriptions_search_subscription_id"), table_name="subscriptions_search")
    op.drop_table("subscriptions_search")
    # ### end Alembic commands ###

```

With the following log messages:
```
2024-07-29 13:14:47 [info     ] Detected added table 'subscriptions_search' [alembic.autogenerate.compare]
2024-07-29 13:14:47 [info     ] Detected added index ''ix_subscriptions_search_subscription_id'' on '('subscription_id',)' [alembic.autogenerate.compare]
```

This is not correct and should be fixed by preferably modifying `orchestrator.db.models.SubscriptionSearchView` so that alembic recognizes it as a materialized view. If that is not possible, the model should be excluded from alembic altogether.

torkashvand:
The `subscriptions_search` view is a manually created materialized view and should not be managed by Alembic. This PR adds an info flag to the view’s table definition and updates the `include_object` logic to skip any table marked as a materialized view. This prevents erroneous migration generation for `subscriptions_search`. 
Resolves #724.

### DIFF
--- a/orchestrator/migrations/env.py
+++ b/orchestrator/migrations/env.py
@@ -41,6 +41,14 @@ target_metadata = BaseModel.metadata
 # ... etc.
 
 
+def include_object(object, name, type_, reflected, compare_to):  # type: ignore
+    """Exclude tables with the 'materialized_view' flag in their info."""
+
+    if type_ == "table" and object.info.get("materialized_view", False):
+        return False
+    return True
+
+
 def run_migrations_offline() -> None:
     """Run migrations in 'offline' mode.
 
@@ -55,7 +63,11 @@ def run_migrations_offline() -> None:
     """
     url = config.get_main_option("sqlalchemy.url")
     context.configure(
-        url=url, target_metadata=target_metadata, literal_binds=True, dialect_opts={"paramstyle": "named"}
+        url=url,
+        target_metadata=target_metadata,
+        literal_binds=True,
+        dialect_opts={"paramstyle": "named"},
+        include_object=include_object,
     )
 
     with context.begin_transaction():
@@ -90,6 +102,7 @@ def run_migrations_online() -> None:
         target_metadata=target_metadata,
         process_revision_directives=process_revision_directives,
         compare_type=True,
+        include_object=include_object,
     )
 
     try:
--- a/orchestrator/migrations/env.py
+++ b/orchestrator/migrations/env.py
@@ -42,7 +42,7 @@ target_metadata = BaseModel.metadata
 
 
 def include_object(object, name, type_, reflected, compare_to):  # type: ignore
-    """Exclude tables with the 'materialized_view' flag in their info."""
+    """Determines if an object should be included."""
 
     if type_ == "table" and object.info.get("materialized_view", False):
         return False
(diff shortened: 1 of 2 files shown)


torkashvand:
Hey @Mark90,

Since users of orchestrator-core might have their own `env.py` as for example we have, and it takes precedence, they should import `include_object` from orchestrator-core to properly exclude materialized views and feature changes.

Our example is:

```python
from alembic import context
from orchestrator.migrations.env import include_object

context.configure(include_object=include_object)
```
This avoids subscriptions_search being detected as a new table. Let me know what you think! 😊

Mark90 (approved):
LGTM!

torkashvand:
OK

> > Hey @Mark90,
> > Since users of orchestrator-core might have their own `env.py` as for example we have, and it takes precedence, they should import `include_object` from orchestrator-core to properly exclude materialized views and future changes.
> 
> That's a good point. A section in the docs about how to reuse this alembic configuration would indeed be nice.

I will update the doc in another PR and let you review it!

orchestrator/migrations/env.py
```diff
@@ -41,6 +41,14 @@
 # ... etc.
 
 
+def include_object(object, name, type_, reflected, compare_to):  # type: ignore
+    """Exclude tables with the 'materialized_view' flag in their info."""
+
+    if type_ == "table" and object.info.get("materialized_view", False):
+        return False
+    return True
```
mrijk, line 49: Personally I would revert the logic (and call it exclude_object) and make this is one-liner, but if you find this more readable that's also fine.

torkashvand: Thanks for your feedback!

I believe the function name should remain `include_object` since it aligns with Alembic's function parameters. However, I will update the function docstring to clarify its purpose and definitely you are right.

Regarding the one-liner suggestion, I followed Alembic's code style and examples, which treat this function as a touchpoint for introducing explicit conditions—not just for `materialized_view` in this case but also for potential future conditions.

Issue #570, opened 2024-03-18
[Feature]: Add option in orchestrator-core to pass UUID's in the state


pboers1988:
### What should we build?

Currently we use str (or UUIDstr) to pass UUID's. This leads to endless conversions between the input form step (which mostly uses UUID's) and the following steps where we have to convert UUIDStr to UUID again.

Proposal: add possibility in Orchestrator core (see injectargs) to pass an UUID and list[UUID]. 

Tasks

- [ ] Update orchestrator-core
- [ ] Simplify workflows (can be done gradually) to take advantage of passing UUID's

hanstrompert:
Check if this works for UUID's. Add warning if UUIDstr is being used.

PR #878, opened 2025-03-07
Refactor inject_args for Direct UUID Support

torkashvand:
- Allow passing UUID, list[UUID], and Optional[UUID], Optional list[UUID] directly.
- Add tests for UUID, list, and optional UUID handling.

Resolves: [#878](https://github.com/workfloworchestrator/orchestrator-core/issues/570)

Read the first commit for review then later ones

### DIFF
--- a/orchestrator/utils/state.py
+++ b/orchestrator/utils/state.py
@@ -150,8 +150,14 @@ def _build_arguments(func: StepFunc | InputStepFunc, state: State) -> list:  # n
 
     Raises:
          KeyError: if requested argument is not in the state, or cannot be reconstructed as an initial domain model.
+         ValueError: if requested argument cannot be converted to the expected type.
 
     """
+
+    def _convert_to_uuid(v: Any) -> UUID:
+        """Converts the value to a UUID instance if it is not already one."""
+        return v if isinstance(v, UUID) else UUID(v)
+
     sig = inspect.signature(func)
     arguments: list[Any] = []
     if sig.parameters:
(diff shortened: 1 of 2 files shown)


orchestrator/utils/state.py
```diff
@@ -150,8 +150,14 @@ def _build_arguments(func: StepFunc | InputStepFunc, state: State) -> list:  # n
...
 
     """
+
+    def _convert_to_uuid(v: Any) -> UUID:
+        """Converts the value to a UUID instance if it is not already one."""
+        return v if isinstance(v, UUID) else UUID(v)
+
     sig = inspect.signature(func)
     arguments: list[Any] = []
     if sig.parameters:
```
Mark90, line 163: Maybe you can add an early exit here to reduce indentation?

```suggestion
    if not sig.parameters:
        return []
```

Mark90 (approved):

orchestrator/utils/state.py
```diff
@@ -150,65 +150,85 @@ def _build_arguments(func: StepFunc | InputStepFunc, state: State) -> list:  # n
...
+        if is_subscription_model_type:
+            subscription_id = _get_sub_id(state.get(name))
+            if subscription_id:
+                sub_mod = param.annotation.from_subscription(subscription_id)
+                arguments.append(sub_mod)
+            else:
+                logger.error("Could not find key in state.", key=name, state=state)
+                raise KeyError(f"Could not find key '{name}' in state.")
+        elif is_list_type(param.annotation, SubscriptionModel):
+            subscription_ids = map(_get_sub_id, state.get(name, []))
```
mrijk, line 194: I think in general we use list comprehension instead of 'map'.

orchestrator/utils/state.py
```diff
@@ -150,65 +150,85 @@ def _build_arguments(func: StepFunc | InputStepFunc, state: State) -> list:  # n
...
+        try:
+            is_subscription_model_type = issubclass(param.annotation, SubscriptionModel)
+        except Exception:
+            is_subscription_model_type = False
+
+        if is_subscription_model_type:
+            subscription_id = _get_sub_id(state.get(name))
+            if subscription_id:
+                sub_mod = param.annotation.from_subscription(subscription_id)
+                arguments.append(sub_mod)
```
mrijk, line 189: Not for this PR, but someday we should get rid of this endless sequence of appends and use 'yield' for cleaner code. (less noise imo).

mrijk (approved):
LGTM

PR #883, opened 2025-03-12
Add callback progress update endpoint

### LINKED ISSUE #882: [Feature]: Update step progress during execution
### Contact Details

_No response_

## Description
Currently, step progress updates in the UI are only reflected at the end of execution. For long-running steps, such as executing Terraform or Ansible playbooks, operators would benefit from real-time progress updates to track which resources are being deployed or modified. Without this, users may assume the system is unresponsive and refresh the page unnecessarily.

## Use Case
- Steps that execute Terraform runs can take significant time, and the output is valuable for tracking deployment status.
- Showing logs or progress in real-time would improve user experience and transparency.

## Proposed Solution
- Introduce a mechanism for updating step progress during execution.
- Extend the existing callback functionality to support intermediate state updates.
- A possible approach (suggested by Adam Kirchberger) is an additional nested endpoint (`/progress`) within the callback job, allowing it to POST updates to the awaiting callback step state.

## Related Discussions
- [Discord](https://discord.com/channels/1295834294270558280/1346788334814629939/1346788334814629939)

### Relevant pseudo code

```shell

```

adamkirchberger:
Add a new endpoint which can be used by callback jobs to provide progress updates during execution.

The progress updates are stored in the process state in a `callback_progress` key, this key is removed once the process is resumed to avoid duplicated job results in the `callback_result` and `callback_progress` keys and also to reduce the size of the final process state object. This means that the callback result is what is persisted and should contain any logs or resulting error messages needed for troubleshooting, and the progress updates should only be used for providing progress during the callback execution.

Closes #882


pboers1988 (approved):
LGTM

pboers1988:
@adamkirchberger It would be amazing if you could add some docs https://github.com/workfloworchestrator/orchestrator-core/blob/fc2803d4c6bffb413f7b0d965b45e7320f0d3110/docs/reference-docs/workflows/callbacks.md#L4

test/unit_tests/workflows/test_async_workflow.py
```diff
@@ -165,3 +166,118 @@ def test_wf_with_multiple_callback_steps(test_client):
...
+        callback_route1 = state["callback_route"]
+        response = test_client.post(callback_route1, json={"ext_data": "12345"})
+        assert response.status_code == 200
+
+        # Check process status
+        response = test_client.get(f"api/processes/{process_id}")
+        response_data = response.json()
+        assert response_data["last_status"] == "awaiting_callback"
+        state = response_data["current_state"]
+        assert state["dr_ext_data"] == "12345"
```
kvklink, line 254: Just a small question, where does the `dr_` prefix come from? I would have expected the key to be simply `ext_data`.

kvklink (approved):

orchestrator/api/api_v1/endpoints/processes.py
```diff
@@ -197,6 +197,41 @@ def continue_awaiting_process_endpoint(
...
+    if token != token_from_state:
+        raise_status(HTTPStatus.NOT_FOUND, "Invalid token")
+
+    progress_key = state.get(DEFAULT_CALLBACK_PROGRESS_KEY, "callback_progress")
+    state = {**state, progress_key: data} | {"__remove_keys": [progress_key]}
+
+    current_step = process.steps[-1]
+    current_step.state = state
+    db.session.add(current_step)
+    db.session.commit()
```
Mark90, line 230: Nice!

It would be great if we can move these highlighted lines into a service function inside `services/processes.py` near the similar `continue_awaiting_process()`. Maybe also a shared function for validating the token and updating current step

test/unit_tests/workflows/test_async_workflow.py
adamkirchberger: The `dr_` prefix is set in the `validate_state` function. This hasn't been changed.

 https://github.com/workfloworchestrator/orchestrator-core/blob/5c686bb6ff4cdb422a9e2b93b17bd14733b7e1a5/test/unit_tests/workflows/test_async_workflow.py#L53

orchestrator/api/api_v1/endpoints/processes.py
adamkirchberger: Great idea, I've moved the logic from the new callback progress endpoint into service functions and refactored the existing `continue_awaiting_process` function to use the same service functions where logic was repeated.

pboers1988 (approved):
LGTM

Issue #294, opened 2023-06-27
Detect usage of `BaseModel`


kvklink:
Detect and somehow deter people from having their objects inherit from the pydantic `BaseModel` instead of `ProductBlockModel`.
When inheriting from `BaseModel`, SQLAlchemy will not store and retrieve them consistently, which results in `ValueError`s.

PR #885, opened 2025-03-15
294 detect usage of basemodel

karimbenbourenane:
Also prevent using BaseModel itself in places where a SubscriptionModel or ProductBlockModel is needed to be used. Raises a TypeError during validation in the __setitem__ method of the
SubsubscriptionModelRegistry, which is a new specialized class to include the validation at the point of entry into its mapping.

### DIFF
--- a/orchestrator/domain/base.py
+++ b/orchestrator/domain/base.py
@@ -21,7 +21,6 @@ from typing import (
     Callable,
     ClassVar,
     Iterable,
-    Literal,
     Mapping,
     Optional,
     TypeVar,
@@ -36,7 +35,6 @@ import structlog
 from more_itertools import bucket, first, flatten, one, only
 from pydantic import BaseModel, ConfigDict, Field, ValidationError
 from pydantic.fields import PrivateAttr
-from pydantic.main import IncEx
 from sqlalchemy import select
 from sqlalchemy.orm import selectinload
 
@@ -90,35 +88,6 @@ class DomainModel(BaseModel):
 
     Contains all common Product block/Subscription instance code
     """
-
-    def model_dump_json(
-        self,
-        *,
-        indent: int | None = None,
-        include: IncEx | None = None,
-        exclude: IncEx | None = None,
-        context: Any | None = None,
-        by_alias: bool = False,
-        exclude_unset: bool = False,
-        exclude_defaults: bool = False,
-        exclude_none: bool = False,
-        round_trip: bool = False,
-        warnings: bool | Literal["none", "warn", "error"] = True,
-        serialize_as_any: bool = False,
-    ) -> str:
-        return super().model_dump_json(
-            indent=indent,
-            include=include,
-            exclude=exclude,
-            context=context,
-            by_alias=by_alias,
-            exclude_unset=exclude_unset,
-            exclude_defaults=exclude_defaults,
-            exclude_none=exclude_none,
-            round_trip=round_trip,
-            warnings=warnings,
-            serialize_as_any=serialize_as_any,
-        )
 
     model_config = ConfigDict(validate_assignment=True, validate_default=True)
 
--- a/orchestrator/domain/base.py
+++ b/orchestrator/domain/base.py
@@ -1462,13 +1462,11 @@ class SubscriptionModelRegistry(dict[str, type[SubscriptionModel]]):
                     self[key] = value
             elif isinstance(m, Iterable):
                 for index, item in enumerate(m):
-                    if not isinstance(item, tuple):
-                        raise TypeError(f"cannot convert dictionary update sequence " f"element #{index} to a sequence")
-                    if length := len(item) != 2:
-                        raise TypeError(
-                            f"dictionary update sequence element #{index} has " f"length {length}; 2 is required"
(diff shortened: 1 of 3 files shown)


karimbenbourenane:
I recognize tests will need to pass. I'll switch from draft once I complete that.

karimbenbourenane:
I've completed the tests. All tests pass locally so if these pass the automated tests this issue is ready to be re-reviewed.

orchestrator/domain/base.py
```diff
@@ -87,6 +91,35 @@ class DomainModel(BaseModel):
...
+        context: Any | None = None,
+        by_alias: bool = False,
+        exclude_unset: bool = False,
+        exclude_defaults: bool = False,
+        exclude_none: bool = False,
+        round_trip: bool = False,
+        warnings: bool | Literal["none", "warn", "error"] = True,
+        serialize_as_any: bool = False,
+    ) -> str:
+        return super().model_dump_json(
```
Mark90, line 109: I could be wrong but this doesn't seem to change anything on the defaults in the overridden model_dump_json ?

test/unit_tests/domain/test_base.py
```diff
@@ -55,6 +54,120 @@ def test_product_block_metadata(test_product_block_one, test_product_one, test_p
...
+        "common_subscription": CommonSubscription,
+        "uncommon_subscription": UncommonSubscription,
+    }
+
+    # These should register successfully
+    for name, model in valid_subscriptions.items():
+        try:
+            SUBSCRIPTION_MODEL_REGISTRY.update({name: model})
+        except TypeError:
+            pytest.fail(f"Subscription {model.__name__} registered as " f"{name} should be valid but raised TypeError.")
```
Mark90, line 168: Great tests! 👍 

It should probably clean up the `SUBSCRIPTION_MODEL_REGISTRY` afterwards though, otherwise it may affect other tests. The easiest/failsafe way would be probably a fixture like this

```
@pytest.fixture
def clean_registry():
    with mock.patch.dict(SUBSCRIPTION_MODEL_REGISTRY):
        yield
```

Mark90: This is probably why the tests are failing, seeing this error log

```
2025-03-17 09:07:07 [error    ] Couldn't resolve product with fixed inputs for a domain model, due to a product_name mismatch. [orchestrator.utils.fixed_inputs] func=check_db_fixed_input_config product_name=common_subscription
2025-03-17 09:07:07 [error    ] Couldn't resolve product with fixed inputs for a domain model, due to a product_name mismatch. [orchestrator.utils.fixed_inputs] func=check_db_fixed_input_config product_name=uncommon_subscription
```

Mark90 (approved):
Few small nitpicks, LGTM!

test/unit_tests/domain/test_base.py
karimbenbourenane: > Great tests! 👍
> 
> It should probably clean up the `SUBSCRIPTION_MODEL_REGISTRY` afterwards though, otherwise it may affect other tests. The easiest/failsafe way would be probably a fixture like this
> 
> ```
> @pytest.fixture
> def clean_registry():
>     with mock.patch.dict(SUBSCRIPTION_MODEL_REGISTRY):
>         yield
> ```

Where should this fixture go, since I'm still very much an arch-core novice? There are many files where fixtures are added so I'm not sure if it goes in one of those, or I can just put it in the file this test exists in immediately before the `test_subscription_model_registry` function. Please advise.

orchestrator/domain/base.py
karimbenbourenane: I have absolutely no clue how this even got generated. I didn't write this at all and didn't even notice it. Thanks for catching it, but im scratching my head at what the hell my IDE did. Maybe I accidentally hit some keyboard shortcut that activated a generator of some sort? I'd like to figure out what caused it because I don't want that happening again, LOL.

test/unit_tests/domain/test_base.py
Mark90: Yes in this file is fine

orchestrator/domain/base.py
```diff
@@ -1402,6 +1439,71 @@ def db_model(self) -> SubscriptionTable:
...
+                    self[key] = value
+            elif isinstance(m, Iterable):
+                for index, item in enumerate(m):
+                    if not isinstance(item, tuple):
+                        raise TypeError(f"cannot convert dictionary update sequence " f"element #{index} to a sequence")
+                    if length := len(item) != 2:
+                        raise TypeError(
+                            f"dictionary update sequence element #{index} has " f"length {length}; 2 is required"
+                        )
+                    self[item[0]] = item[1]
```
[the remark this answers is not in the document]
karimbenbourenane: Simplified as requested.

karimbenbourenane:
@Mark90 All tests passed, and I made all the changes you asked for. Let me know if I missed anything. Thanks for the review!

pboers1988 (approved):
Good Work :) more difficult than it seemed beforehand.

Mark90:
This PR resolved #294

PR #891, opened 2025-03-20
Add GEANT as a partner in codes they have changed

torkashvand:


torkashvand:
Hi @hanstrompert 

I used the following Git command to find our contributions:
```shell
git log --author="torkashvand.dev@gmail.com" --name-only --pretty=format: | sort -u
```
We did the same command for Karel contributions.

Do you know whether we should use `2019-2020` or `2019-2025`? If the latter is correct, would it be a good idea to update all files with the new years?

hanstrompert:
> Do you know whether we should use `2019-2020` or `2019-2025`? If the latter is correct, would it be a good idea to update all files with the new years?

Usually we only update the year so it covers the last update to that file. So you should have a look at the files where you added GEANT as co-author, if you do not know the exact date of the last contribution, it cannot hurt to just change it to ...-2025. We do not update the year when the file is not changed, although I guess that also wouldn't hurt. It is very well possible that not all copyright notices carry the correct year(s) ...

pboers1988 (approved):

PR #907, opened 2025-03-31
Add InputStateTable to orchestrator/db/__init__.py

RichieB2B:
Unless there was a good reason not to do this?


pboers1988 (approved):
LGTM, forgot to do this.

PR #913, opened 2025-04-08
Add process_id to workflow @step logging

RichieB2B:
When we log from a workflow `@step` we want to know which `process_id` the logs belong to. We have created our local version of the `@step` decorator for this but I think it makes sense to add this to the main WFO project.


pboers1988 (approved):
LGTM

PR #917, opened 2025-04-10
Add workflow_name and process_id to @retrystep structlog context variables

RichieB2B:
Like #913 but for `@retrystep`


Mark90:
Thanks for improving this!

Issue #844, opened 2025-02-14
Create a WFO metrics endpoint/exporter


srichmond:
# Background

The idea is to have a place were certain performance metrics are exposed so that they can be scraped and pushed to a TSDB. 

### Instant performance metrics: 
- number of current workflow active
- number of current workflows active failed
- number of current tasks active
- number of current tasks active failed
- number of completed workflows (I am kind of unsure about this one)
- duration of completed workflows (same here)

### long-term metrics: 
- number of products
- number of subscriptions per state

These are the first things that pop up in my mind, I guess it could make sense to get some interesting stats about database usage but probably there are other ways to do it 

# Implementation

I guess a scraping endpoint that can be pointed at by a prometheus would work, but I also know that many of us use influxdb so this approach might be suboptimal. The idea came from the metric server in Proxomox where you can choose between graphite and influxdb. 

I do not think this has to be something configurable via GUI.  

# Tasks

- [x] Add /metrics endpoint to Orchestrator
- [x] Setup in such a way that the core-orchestrator provides default metrics.
- [x] Figure out how to represent 'dynamic' metrics. For example the number of product types can vary. Every metric represents one single datapoint for one product.
- [ ] Optional: provide metadata: how long does it take to collect a certain metric.
- [ ] ~Do we need async support? (https://github.com/hynek/prometheus-async)~
- [x] Unit tests
- [x] Test on large production-like dataset

hanstrompert:
Probably better to use counters instead of gauges to be less interval dependent.

pboers1988:
These metrics should be exposed as counters/booleans or enum-states
Make the endpoint extensible and fast.

- total task seconds per task
- total workflow seconds per workflow
- total no of tasks per task name
- total no of workflows per workflow name
- Reporting workflow/task state
- no of subscriptions per product type, per product, per lifecycle state
- Status of the workflow engine
- No of subscription in/out of sync per lifecycle state

hanstrompert:
Return Prometheus style output, see https://prometheus.io/docs/concepts/data_model/#notation and the link to best common practices.

mrijk:
Curious if an event-based approach (for example Kafka or similar) would be useful

mrijk:
Subscription stats:

Number of subscriptions per product:
```sql
SELECT 
    p.product_id, 
    p.name,
    COUNT(s.subscription_id) AS subscription_count
FROM products AS p
[...]

PR #920, opened 2025-04-16
Add metrics endpoint

misterbox:
Adds a `/api/metrics` endpoint to provide various datapoints to be used by Prometheus.

Resolves [#844]


kvklink:
Added documentation on how to extend the collectors with home-grown ones, this now also closes #921

orchestrator/metrics/processes.py
```diff
@@ -0,0 +1,160 @@
...
+            process_counts.add_metric(
+                [
+                    row.last_status,
+                    str(row.created_by),
+                    str(row.is_task),
+                    row.product_name,
+                    row.workflow_name,
+                    row.customer_id,
+                    row.workflow_target,
+                ],
```
Mark90, line 143: This list can be created in a variable and reused below

Mark90 (approved):
Very nice work :) 

1 small nitpick, LGTM otherwise!

orchestrator/metrics/processes.py
kvklink: Should be better now

test/unit_tests/metrics/test_subscription_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+    subscription.save()
+    db.session.commit()
+    expected_metric = (
+        f'wfo_subscriptions_count{{customer_id="{CUSTOMER_ID}",insync="False",lifecycle_state="active",'
+        f'product_name="Product 1"}} 1.0'
+    )
+    response = test_client.get("/api/metrics")
+    assert expected_metric in response.text
+
+    subscription.insync = True
```
mrijk, line 52: Personally I like unit tests that test exactly one scenario. Is there a reason for not having a separate test for this second part?

kvklink: My reasoning for the unit tests was to check whether it works when taking a subscription out of sync, and also when taking it back into sync. I can split them up into separate tests, I agree with the sentiment of "one test one scenario"

test/unit_tests/metrics/test_engine_metrics.py
```diff
@@ -0,0 +1,18 @@
...
+    test_client.put("/api/settings/status", json={"global_lock": True})
+    response = test_client.get("api/metrics")
+    expected_metric_lines = [
+        "# HELP wfo_engine_status Current workflow engine status.",
+        "# TYPE wfo_engine_status gauge",
+        'wfo_engine_status{wfo_engine_status="PAUSED"} 1.0',
+        'wfo_engine_status{wfo_engine_status="PAUSING"} 0.0',
+        'wfo_engine_status{wfo_engine_status="RUNNING"} 0.0',
+    ]
+    assert all(line in response.text for line in expected_metric_lines)
```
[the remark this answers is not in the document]
kvklink: The help text im indifferent about, but in this specific case I would argue that the name of the metric should also be included. If this were to accidentally change, they would show up in someone's Prometheus as two separate metrics

orchestrator/metrics/subscriptions.py
```diff
@@ -0,0 +1,93 @@
...
+            unit="count",
+            documentation="Number of subscriptions per product, lifecycle state, customer, and in sync state.",
+        )
+
+        for row in _get_subscriptions():
+            subscriptions.add_metric(
+                [row.product_name, row.lifecycle_state, row.customer_id, str(row.insync)], row.subscription_count
+            )
+
+        return [subscriptions]
```
mrijk, line 93: Didn't check in detail, but an implementation using 'yield' might be slightly cleaner

mrijk (approved):
LGTM. Couple of minor remarks, mostly about the unit tests.

Igoranze (approved):

orchestrator/metrics/subscriptions.py
kvklink: Gave this a try, but that causes _all_ unit tests to error on:
```sqlalchemy.exc.OperationalError: (psycopg2.errors.ObjectInUse) database "orchestrator-core-test-gw[0..9]" is being accessed by other users```
So I think I will leave these as `return` statements

Mark90: Yep, these objects are tied to the sqlalchemy session so I would also use return

test/unit_tests/metrics/test_subscription_metrics.py
Mark90: For me this is a legitimate test scenario because the metric is a gauge, so it's nice to see that it will actually change if the "source" changes

PR #930, opened 2025-04-30
Adding Target.VALIDATION to Workflows, including isTask bool

saket-geant:
Validate Subscriptions workflows do not explicitly check for "validation" workflow type

Resolves - [804](https://github.com/orgs/workfloworchestrator/projects/7/views/1?pane=issue&itemId=100133943&issue=workfloworchestrator%7Corchestrator-core%7C804)

EDIT:
Added a VALIDATE target type for Workflows to make distinguish explicitly between SYSTEM and VALIDATE workflows. Added the isTask bool to WorkflowTable to help with validation if an item is a task or not. 

This is a breaking change, so make sure you read the documentation migration 4.0.MD file.


orchestrator/db/models.py
```diff
@@ -418,6 +418,9 @@ def select() -> Select:
             select(WorkflowTable).options(undefer(WorkflowTable.deleted_at)).filter(WorkflowTable.deleted_at.is_(None))
         )
 
+    def is_task(self) -> bool:
+        return self.target == Target.VALIDATE
```
Mark90, line 422: Both Validation and System workflows are tasks :) 

```suggestion
    def is_task(self) -> bool:
        return self.target in (Target.VALIDATE, Target.SYSTEM)
```

docs/architecture/application/workflow.md
```diff
@@ -204,14 +204,14 @@ Validate workflows run integrity checks on an existing subscription. Checking th
...
-        target="SYSTEM",
+        target="VALIDATE",
         description="Validate Node Enrollment before production",
         tag="NodeEnrollment",
         search_phrase="Node Enrollment%",
     )
 ```
 
-It uses a `target` of `SYSTEM` - similar to how tasks are defined. That target is more of a free form sort of thing. Same thing with the `name` - that's the name of the actual workflow, and the `tag` is shared by of this set of workflows.
+It uses a `target` of `VALIDATE` - similar to how tasks are defined. That target is more of a free form sort of thing. Same thing with the `name` - that's the name of the actual workflow, and the `tag` is shared by of this set of workflows.
```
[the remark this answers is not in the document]
Igoranze: Thanks for the clarification

docs/migration-guide/4.0.md
```diff
@@ -0,0 +1,56 @@
+# 4.0 Migration Guide
+
+In this document we describe the steps that should be taken to migrate from `orchestrator-core` v3 to v4.
+
+## About 4.0
+
+In this release, the [workflow target validate](../architecture/application/workflow.md#validate) `SYSTEM` has been
+removed for validate workflows and a new target `VALIDATE` has been added. The `SYSTEM` target was used to target
```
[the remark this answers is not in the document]
Igoranze: Used your suggestion

orchestrator/db/models.py
Igoranze: Got it!

orchestrator/schedules/validate_subscriptions.py
```diff
@@ -35,6 +36,10 @@
 
 @scheduler(name="Subscriptions Validator", time_unit="day", at="00:10")
 def validate_subscriptions() -> None:
+    if app_settings.TARGET != Target.VALIDATE:
+        logger.info("Skipping subscription validation as the target is not VALIDATE.")
+        return
```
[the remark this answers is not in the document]
Igoranze: I understand that only target="VALIDATE" should be run on the cronjob

pboers1988 (approved):
LGTM

tjeerddie (approved):

orchestrator/services/processes.py
```diff
@@ -108,7 +107,7 @@ def _db_create_process(stat: ProcessStat) -> None:
         workflow_id=wf_table.workflow_id,
         last_status=ProcessStatus.CREATED,
         created_by=stat.current_user,
-        is_task=stat.workflow.target == Target.SYSTEM,
+        is_task=stat.workflow.is_task,
```
Mark90: As discussed, better to change this to `wf_table.is_task` to use the value from the `WorkflowTable` DB model. 
And remove the default/computed `is_task` value from the `Workflow(Protocol)`.
So that there is exactly 1 definition of whether a workflow is a task

orchestrator/workflow.py
```diff
@@ -212,6 +214,11 @@ def wrapping_function() -> NoReturn:
     wrapping_function.target = target
     wrapping_function.steps = steps
 
+    wf_table = get_workflow_by_name(wrapping_function.name)
```
Mark90, line 217: As discussed, it's safer to perform this DB query in the "Start process" code than here, to avoid a stale `.is_task` value in memory

docs/migration-guide/4.0.md
```diff
@@ -0,0 +1,60 @@
+# 4.0 Migration Guide
+
+In this document we describe the steps that should be taken to migrate from `orchestrator-core` v3 to v4.
+
+## About 4.0
+
+In this release a new workflow target `VALIDATE` has been added
+for [validation workflows](../architecture/application/workflow.md#validate). Previously, the `SYSTEM` target was used
+to target `VALIDATE` has been added. The `SYSTEM` target was used to target
+validation workflows, thus some workflows were expected to be run in the context of a system. This was not the case for
```
Mark90, line 10: There's some duplication in this text

orchestrator/workflows/tasks/validate_products.py
```diff
@@ -109,7 +109,7 @@ def check_that_products_have_create_modify_and_terminate_workflows() -> State:
         workflows = {
             c.target
             for c in product.workflows
-            if c.target in ["CREATE", "TERMINATE", "MODIFY", "SYSTEM"] and c.name != "modify_note"
+            if c.target in ["CREATE", "TERMINATE", "MODIFY", "SYSTEM", "VALIDATE"] and c.name != "modify_note"
         }
         if len(workflows) < 4:
```
mrijk, line 114: I would expect that this (terrible) magic number would also need an update

Mark90: Good point, seems that this refers to the length of the list so we could create a variable for that list.

Also `SYSTEM` should probably be removed because it's no longer the target for product validation workflows

Mark90 (approved):

Issue #933, opened 2025-05-01
[Bug]: Callback progress updating incorrect state key


adamkirchberger:
### Contact Details

adam.kirchberger@nomios.co.uk

### What happened?

Sending callback progress works fine for the first progress update, however subsequent updates are not correctly updating the progress key and instead creating a new key with the value of the initial update.

**Progress update with string**
```
{
    "callback_progress": "update 1",
    "update 1": "update 2"
}
```

**Progress update with dict**

Subsequent update fails entirely as dict is unhashable and can't be set as the key in the state.

**Investigation**

The problem line seems to be
https://github.com/workfloworchestrator/orchestrator-core/blob/f0677d1f58549a78263c6f021b537f31ceea1ef4/orchestrator/services/processes.py#L685

### Version

3.2.1

### What python version are you seeing the problem on?

Python 3.11

### Relevant log output

```shell
state = pstat.state.unwrap()
        # progress_key = state.get("__callback_progress_key") or DEFAULT_CALLBACK_PROGRESS_KEY
        progress_key = state.get(DEFAULT_CALLBACK_PROGRESS_KEY, "callback_progress")
>       state = {**state, progress_key: data} | {"__remove_keys": [progress_key]}
E       TypeError: unhashable type: 'dict'

orchestrator/services/processes.py:687: TypeError
```

PR #934, opened 2025-05-01
fix: subsequent callback progress key error

adamkirchberger:
Fixes https://github.com/workfloworchestrator/orchestrator-core/issues/933

The progress key was being incorrectly set to the value of the previous progress value. This only happens on subsequent progress updates and unfortunately the unit tests did not test for subsequent updates which would have caught this error.

Updated the progress key logic to fix this, and updated the tests as well.

### DIFF
--- a/orchestrator/services/processes.py
+++ b/orchestrator/services/processes.py
@@ -682,7 +682,7 @@ def update_awaiting_process_progress(
     ensure_correct_callback_token(pstat, token=token)
 
     state = pstat.state.unwrap()
-    progress_key = state.get(DEFAULT_CALLBACK_PROGRESS_KEY, "callback_progress")
+    progress_key = state.get("__callback_progress_key", DEFAULT_CALLBACK_PROGRESS_KEY)
     state = {**state, progress_key: data} | {"__remove_keys": [progress_key]}
 
     replace_current_step_state(process, new_state=state)
--- a/orchestrator/services/processes.py
+++ b/orchestrator/services/processes.py
@@ -682,7 +682,7 @@ def update_awaiting_process_progress(
     ensure_correct_callback_token(pstat, token=token)
 
     state = pstat.state.unwrap()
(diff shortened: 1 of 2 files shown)


orchestrator/services/processes.py
```diff
@@ -682,7 +682,7 @@ def update_awaiting_process_progress(
     ensure_correct_callback_token(pstat, token=token)
 
     state = pstat.state.unwrap()
-    progress_key = state.get(DEFAULT_CALLBACK_PROGRESS_KEY, "callback_progress")
+    progress_key = state.get("__callback_progress_key", DEFAULT_CALLBACK_PROGRESS_KEY)
```
Mark90, line 685: This was probably inspired by the `callback_step()` which has a kwarg `result_key: str` that is stored in  the state as `"__callback_result_key": result_key`. I *think* this was added to be able to specify a different state key in case the `callback_result` key is needed for something else, but I'm not entirely sure.

Anyway, as far as I can tell there is currently no way to set `__callback_progress_key` in the state. Is this something you'd ever see becoming useful? If it isn't, I'd also be fine with simplifying this to:

```suggestion
    progress_key = DEFAULT_CALLBACK_PROGRESS_KEY
```

Mark90 (approved):

kvklink (approved):
LGTM

orchestrator/services/processes.py
kvklink: We already weren't using the `__callback_result_key` functionality in the first place, so I have no strong preference here

adamkirchberger: The initial reason for using this approach was mainly to replicate how it was done in the other callback step, which would give the workflow developer the ability to change the progress key. The developer would be able to set this by setting the `__callback_progress_key` in a previous step if I'm not mistaken.

But I agree, I don't see a real need for this so lets simplify it.

PR #939, opened 2025-05-08
Add role based access control to running workflows (#931)

### LINKED ISSUE #931: [Feature]: Add Role based access control to running workflows - inputstep decorator
### Contact Details

_No response_

### What should we build?

Add the `authorize_callback` parameter to the `@inputstep` decorator.

See #782 for details.

### Relevant pseudo code

```shell

```

eenblam:
This PR adds role-based access control to running, resuming and retrying workflows.

Documentation has been added to the `Auth(n|z)` page of the reference documentation.

As noted in the docs, this functionality is still in beta as the UI is not yet adapted to handle authorization rejections, nor does it hide/disable UI elements for starting/resuming workflows when a user isn't allowed to do so. Follow-up stories to implement these things are on the WFO Partner Code Sprint.

Version bumped to 4.1.0rc2.

Closes #931

### DIFF
--- a/orchestrator/api/api_v1/endpoints/processes.py
+++ b/orchestrator/api/api_v1/endpoints/processes.py
@@ -101,10 +101,10 @@ def get_auth_callbacks(pstat: ProcessStat) -> tuple[Authorizer | None, Authorize
     # Default to workflow start callbacks
     auth_resume = pstat.workflow.authorize_callback
     # Retry defaults to the workflow start callback if not otherwise specified
-    auth_retry = pstat.workflow.retry_auth_callback if pstat.workflow.retry_auth_callback else auth_resume
+    auth_retry = pstat.workflow.retry_auth_callback if pstat.workflow.retry_auth_callback is not None else auth_resume
     # Iterate over previous steps to look for policy changes
     remaining_steps = pstat.log
-    past_steps = pstat.workflow.steps[:-len(remaining_steps)]
+    past_steps = pstat.workflow.steps[: -len(remaining_steps)]
     # Review last steps and current step
     for step in past_steps + [pstat.log[0]]:
         if step.resume_auth_callback and step.retry_auth_callback is None:
@@ -112,16 +112,16 @@ def get_auth_callbacks(pstat: ProcessStat) -> tuple[Authorizer | None, Authorize
             auth_resume = step.resume_auth_callback
             auth_retry = step.resume_auth_callback
             continue
-        elif step.resume_auth_callback and step.retry_auth_callback:
+        if step.resume_auth_callback and step.retry_auth_callback:
             auth_resume = step.resume_auth_callback
             auth_retry = step.retry_auth_callback
             continue
-        elif step.resume_auth_callback is None and step.retry_auth_callback:
+        if step.resume_auth_callback is None and step.retry_auth_callback:
             # Only update retry
             auth_retry = step.retry_auth_callback
             continue
-        else: # Both None
-            continue
+        # Both None
+        continue
     return auth_resume, auth_retry
 
 
@@ -209,12 +209,11 @@ def resume_process_endpoint(
     pstat = load_process(process)
     auth_resume, auth_retry = get_auth_callbacks(pstat)
     if process.last_status == ProcessStatus.SUSPENDED:
-        if not auth_resume(user_model):
+        if auth_resume is not None and not auth_resume(user_model):
             raise_status(HTTPStatus.FORBIDDEN, "User is not authorized to resume step")
     elif process.last_status == ProcessStatus.FAILED:
-        if not auth_retry(user_model):
+        if auth_retry is not None and not auth_retry(user_model):
             raise_status(HTTPStatus.FORBIDDEN, "User is not authorized to retry step")
-
 
     broadcast_invalidate_status_counts()
     broadcast_func = api_broadcast_process_data(request)
--- a/orchestrator/api/api_v1/endpoints/processes.py
+++ b/orchestrator/api/api_v1/endpoints/processes.py
@@ -25,7 +25,7 @@ from fastapi.param_functions import Body, Depends, Header
 from fastapi.routing import APIRouter
 from fastapi.websockets import WebSocket
 from fastapi_etag.dependency import CacheHit
-from more_itertools import chunked
+from more_itertools import chunked, last
 from sentry_sdk.tracing import trace
 from sqlalchemy import CompoundSelect, Select, select
 from sqlalchemy.orm import defer, joinedload
@@ -110,23 +110,13 @@ def get_auth_callbacks(steps: StepList, workflow: Workflow) -> tuple[Authorizer 
     # auth_retry defaults to the workflow start callback if not otherwise specified.
     # A workflow SHOULD have both callbacks set to not-None. This enforces the correct default regardless.
     auth_retry = workflow.retry_auth_callback or auth_resume  # type: ignore[unreachable, truthy-function]
-    # Iterate over previous steps to look for policy changes
-    for step in steps:
-        if step.resume_auth_callback and step.retry_auth_callback is None:
-            # Set both to authorize_callback
-            auth_resume = step.resume_auth_callback
-            auth_retry = step.resume_auth_callback
-            continue
-        if step.resume_auth_callback and step.retry_auth_callback:
-            auth_resume = step.resume_auth_callback
-            auth_retry = step.retry_auth_callback
-            continue
-        if step.resume_auth_callback is None and step.retry_auth_callback:
-            # Only update retry
-            auth_retry = step.retry_auth_callback
-            continue
-        # Both None
-        continue
+
+    # Choose the most recently established value for resume.
+    auth_resume = last(filter(None, (step.resume_auth_callback for step in steps)), auth_resume)
+    # Choose the most recently established value for retry, unless there is a more recent value for resume.
+    auth_retry = last(
+        filter(None, (step.retry_auth_callback or step.resume_auth_callback for step in steps)), auth_retry
+    )
     return auth_resume, auth_retry
 
 
--- a/orchestrator/api/api_v1/endpoints/processes.py
+++ b/orchestrator/api/api_v1/endpoints/processes.py
@@ -64,7 +64,7 @@ from orchestrator.websocket import (
     broadcast_process_update_to_websocket,
     websocket_manager,
 )
-from orchestrator.workflow import ProcessStat, ProcessStatus
+from orchestrator.workflow import ProcessStat, ProcessStatus, StepList, Workflow
 from pydantic_forms.types import JSON, State
 
 router = APIRouter()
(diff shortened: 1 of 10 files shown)


orchestrator/workflow.py
```diff
@@ -166,14 +168,39 @@ def __repr__(self) -> str:
...
+            user_model = cast(OIDCUserModel, user_model)
+        else:
+            logger.error("authorize_user_from_state: no user model")
+
+        if authorize_callback is not None:
+            logger.error("authorize_user_from_state: callback found")
+            authorize_callback(user_model)
+            if not authorize_callback(user_model):
+                logger.error("authorize_user_from_state: FORBIDDEN")
+                #TODO not sure that step name is available here, but could put it on state?
```
pboers1988, line 186: It might be available in the `StateInputStepFunc` object

orchestrator/services/processes.py
```diff
@@ -515,8 +551,17 @@ def thread_resume_process(
         user_inputs = [{}]
 
     pstat = load_process(process)
-    if not pstat.workflow.authorize_callback(user_model):
-        raise_status(HTTPStatus.FORBIDDEN, error_message_unauthorized(str(process.workflow_name)))
+
+    auth_resume, auth_retry = get_auth_callbacks(pstat)
+    #TODO if retrying a WAITING step, can I detect that the system is retrying automatically?
+    # Possibly user=None means no auth and retry button, user=SYSTEM means resume_all.
+    # What about retry when logged in?
```
eenblam, line 558: This has proven tricky to experiment with. The example orchestrator is currently non-operational, and I've had trouble loading these changes into our own orchestrator instance. Guidance would be appreciated on how I can distinguish a workflow being retried manually vs. resumed from WAITING.

orchestrator/workflow.py
```diff
@@ -90,7 +96,8 @@ class Workflow(Protocol):
     __qualname__: str
     name: str
     description: str
-    authorize_callback: Callable[[OIDCUserModel | None], bool]
+    authorize_callback: Authorizer
+    retry_auth_callback: Authorizer
```
eenblam, line 100: I can see different arguments for what to name these. For now, I've chosen to not change Alex's naming of `authorize_callback` in case it's already in use downstream, but I'm open to feedback on how best to name the following callbacks. Each is a function implementing a RBAC policy.

* `authorize_callback`: called when a user starts a workflow.
* `resume_auth_callback`: called when a user resumes a workflow paused by an `inputstep`
* `retry_auth_callback`:  called when a failed step is retried. Arg to both `@workflow` and `@inputstep`.

orchestrator/services/processes.py
eenblam: I was able to resolve this by lifting this check up to `resume_process_endpoint`. This "fixes" the issue by restricting the scope of what's being considered. There's still the matter of adding RBAC to `resume_all_processess_endpoint`, but that might be a bit trickier to do and needs specification.

I'll make sure that's addressed in a ticket somewhere.

orchestrator/api/api_v1/endpoints/processes.py
```diff
@@ -86,6 +87,44 @@ def check_global_lock() -> None:
...
+    callbacks will be lost if they're only available in the process state.
+
+    Priority:
+    - RESUME callback is explicit RESUME callback, else previous START/RESUME callback
+    - RETRY callback is explicit RETRY, else explicit RESUME, else previous RETRY
+    """
+    # Default to workflow start callbacks
+    auth_resume = pstat.workflow.authorize_callback
+    # Retry defaults to the workflow start callback if not otherwise specified
+    auth_retry = pstat.workflow.retry_auth_callback if pstat.workflow.retry_auth_callback is not None else auth_resume
```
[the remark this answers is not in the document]
eenblam: Thanks Mark! Updated, along with some caveats added in comments for context.

orchestrator/api/api_v1/endpoints/processes.py
```diff
@@ -86,6 +87,44 @@ def check_global_lock() -> None:
...
+    # Iterate over previous steps to look for policy changes
+    remaining_steps = pstat.log
```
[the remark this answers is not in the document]
eenblam: Ha, it makes sense in retrospect that it's a "log" of the past steps, but I admit I was surprised when I first realized what this was.

orchestrator/services/processes.py
```diff
@@ -467,9 +467,7 @@ def thread_start_process(
     user_model: OIDCUserModel | None = None,
     broadcast_func: BroadcastFunc | None = None,
 ) -> UUID:
-    pstat = create_process(workflow_key, user_inputs=user_inputs, user=user)
-    if not pstat.workflow.authorize_callback(user_model):
-        raise_status(HTTPStatus.FORBIDDEN, error_message_unauthorized(workflow_key))
```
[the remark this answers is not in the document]
eenblam: Exactly! I favored the one in `create_process` since it's also called by celery.

orchestrator/api/api_v1/endpoints/processes.py
```diff
@@ -86,6 +87,39 @@ def check_global_lock() -> None:
...
+    - RETRY callback is explicit RETRY, else explicit RESUME, else previous RETRY
+    """
+    # Default to workflow start callbacks
+    auth_resume = workflow.authorize_callback
+    # auth_retry defaults to the workflow start callback if not otherwise specified.
+    # A workflow SHOULD have both callbacks set to not-None. This enforces the correct default regardless.
+    auth_retry = workflow.retry_auth_callback or auth_resume  # type: ignore[unreachable, truthy-function]
+
+    # Choose the most recently established value for resume.
+    auth_resume = last(filter(None, (step.resume_auth_callback for step in steps)), auth_resume)
```
mrijk, line 115: Not a big deal, but in general we use list comprehension instead of filter

orchestrator/workflow.py
```diff
@@ -270,9 +287,16 @@ def wrapper(state: State) -> Process:
     return decorator
 
 
-def inputstep(name: str, assignee: Assignee) -> Callable[[InputStepFunc], Step]:
+def inputstep(
+    name: str,
+    assignee: Assignee,
+    resume_auth_callback: Authorizer | None = None,
```
mrijk, line 293: Maybe it's an option to use default callbacks instead of None, so you can always call the function and don't have to check if they are provided.

mrijk (approved):
LGTM

orchestrator/api/api_v1/endpoints/processes.py
```diff
@@ -86,6 +87,44 @@ def check_global_lock() -> None:
...
+        if step.resume_auth_callback and step.retry_auth_callback:
+            auth_resume = step.resume_auth_callback
+            auth_retry = step.retry_auth_callback
+            continue
+        if step.resume_auth_callback is None and step.retry_auth_callback:
+            # Only update retry
+            auth_retry = step.retry_auth_callback
+            continue
+        # Both None
+        continue
```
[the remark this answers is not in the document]
Mark90: Medically: you may want to get that gopher checked :P 

Yeah I can see how this implementation naturally followed from describing the procedure as code. When not knowing that procedure and simply looking at the code, the functional approach is less effort to "grok" - but that is subjective of course. 

Thanks for reworking the code - and fixing the bug in my example. It's always welcome to challenge our suggestions of course :)

orchestrator/workflow.py
Mark90: I don't think this would work with the design, which hinges on being able to detect that `resume_auth_callback` is not provided - in which case the "normal" auth_callback is used

test/unit_tests/api/test_processes.py
```diff
@@ -593,3 +595,116 @@ def unauthorized_workflow():
...
+        ((A, None, None, D), (A, D)),
+        ((None, B, None, D), (None, D)),
+        ((A, B, None, D), (A, D)),
+        ((None, None, C, D), (C, D)),  # 4
+        ((A, None, C, D), (C, D)),
+        ((None, B, C, D), (C, D)),
+        ((A, B, C, D), (C, D)),
+    ]
+    for case in cases:
+        auth, retry, step_resume_auth, step_retry_auth = case[0]
```
Mark90, line 697: You could use `@pytest.mark.parametrize()` to run each case as a different input scenario for this unittest. Benefit is that you don't need a for loop, you can see whether it's just 1 scenario failing or 3 (or all). There's several other tests in the codebase that utilize this that you can use as a reference.

If there's test preparations that are relatively "expensive", they could be moved into a module-level fixture (`@pytest.fixture(scope="module")`)

orchestrator/workflow.py
eenblam: I think that's a very good general approach that may not fit here without additional refactoring.

The callbacks for `workflow` ultimately use a default callback of `allow`, but for `inputstep` the behavior of "no callback given" is "look for a callback from the preceding `workflow` or `inputstep`." For this to work, we need a sentinel value on the resulting `Step` that can be examined at runtime.

That sentinel value doesn't have to be `None`, but I don't think it's ideal for it to be of type `Authorizer` or even callable if it shouldn't be called in the first place.

test/unit_tests/api/test_processes.py
eenblam: Thanks Mark! I added this in, and it immediately improved my testing experience.

orchestrator/api/api_v1/endpoints/processes.py
```diff
@@ -163,6 +201,15 @@ def resume_process_endpoint(
...
         raise_status(HTTPStatus.CONFLICT, "Resuming a resumed workflow is not possible")
 
+    pstat = load_process(process)
+    auth_resume, auth_retry = get_auth_callbacks(get_current_steps(pstat), pstat.workflow)
+    if process.last_status == ProcessStatus.SUSPENDED:
+        if auth_resume is not None and not auth_resume(user_model):
+            raise_status(HTTPStatus.FORBIDDEN, "User is not authorized to resume step")
+    elif process.last_status == ProcessStatus.FAILED:
+        if auth_retry is not None and not auth_retry(user_model):
+            raise_status(HTTPStatus.FORBIDDEN, "User is not authorized to retry step")
```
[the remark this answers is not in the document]
Mark90: Based on implementation of the endpoint `resume_all_processess_endpoint` these are the process statuses eligible for a resume

```
                [
                    ProcessStatus.FAILED,
                    ProcessStatus.WAITING,
                    ProcessStatus.API_UNAVAILABLE,
                    ProcessStatus.INCONSISTENT_DATA,
                ]
```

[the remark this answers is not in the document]
eenblam: Good catch. Added `can_be_resumed(status: ProcessStatus)`.

orchestrator/api/api_v1/endpoints/processes.py
Mark90: Agree, let's keep `filter()`

Issue #804, opened 2025-01-22
[Bug]: Validate Subscriptions workflows do not explicitly check for "validation" workflow types


Igoranze:
### What happened?

From the `validate_subscriptions` and `task_validate_product_type` the workflows are run based on the `Target.SYSTEM`. However, there is no explicit check on the workflows to verify these are only `validate` workflows. Thus if for some reason there are other workflows, the code will run them.

Should we add a validate check on the workflows explicit or not?

### Version

2.9.2rc2

### What python version are you seeing the problem on?

Python 3.12

### Relevant log output

```shell

```

pboers1988:
Good suggestion, a first step in this would be to add TARGET.VALIDATE to the orchestrator-core so the validate_subscriptions schedule. It would also be useful to add an `is_task` boolean to the workflows table.

kvklink:
FYI: In GÉANT we encountered this situation as well, and we worked around it by enforcing that [any validation workflow name starts with `validate_`](https://gitlab.software.geant.org/goat/gap/geant-service-orchestrator/-/blob/develop/gso/schedules/validate_subscriptions.py?ref_type=heads#L35)

saket-geant:
- [x] Add Target.VALIDATE
- [x] Add WorkflowTable.is_task computed property on DB model + GQL schema
- [x] Change all workflow.target == SYSTEM evaluations to workflow.is_task
- [x] Change validate_subscriptions to only execute on Target.VALIDATE
- [ ] Documentation + Example data migration script to update existing workflow rows (including warning that validation workflow names may be implementation specific)
- [ ] Major version bump

Initial [MR](https://github.com/workfloworchestrator/orchestrator-core/pull/930/files) on core

Mark90:
> Change all workflow.target == SYSTEM evaluations to workflow.is_task

Hi Saket, perhaps you checked this by accident, or there was a misunderstanding.
Most of the code still uses `.target == SYSTEM` evaluations which need to be changed.
I've unchecked the box for now

PR #941, opened 2025-05-15
804 Add Validate to backed response model

Igoranze:
https://github.com/workfloworchestrator/orchestrator-core/issues/804

Minor update on response model

### DIFF
--- a/test/unit_tests/api/test_subscriptions.py
+++ b/test/unit_tests/api/test_subscriptions.py
@@ -497,7 +497,7 @@ def test_insync_invalid_tagged(seed, test_client):
         "modify": [],
         "terminate": [],
         "system": [],
-        "validate": []
+        "validate": [],
     }
 
 


orchestrator/schemas/workflow.py
```diff
@@ -60,6 +60,7 @@ class SubscriptionWorkflowListsSchema(OrchestratorBaseModel):
     modify: list[WorkflowListItemSchema]
     terminate: list[WorkflowListItemSchema]
     system: list[WorkflowListItemSchema]
+    validate_: list[WorkflowListItemSchema] = Field(default_factory=list, alias="validate")
```
mrijk, line 63: why the underscore after 'validate'?

Mark90: To avoid name collision with `BaseModel.validate()`

Igoranze: Thanks Mark, that is correct!

PR #958, opened 2025-05-28
Make is_task backwards compatible

Igoranze:
See: 
https://github.com/orgs/workfloworchestrator/projects/7/views/1?pane=issue&itemId=112782644&issue=workfloworchestrator%7Corchestrator-core%7C957

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.0.1rc1
+current_version = 4.0.1
 commit = False
 tag = False
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(rc(?P<build>\d+))?
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.0.0
+current_version = 4.0.0rc1
 commit = False
 tag = False
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(rc(?P<build>\d+))?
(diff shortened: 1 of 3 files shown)


orchestrator/migrations/helpers.py
```diff
@@ -144,27 +159,41 @@ def create_workflow(conn: sa.engine.Connection, workflow: dict) -> None:
...
-            """
-        ),
-        workflow,
-    )
+                         CROSS JOIN new_workflow AS nw
+                WHERE p.product_type = :product_type
+                ON CONFLICT DO NOTHING \
+                """
+    else:
+        # Remove is_task from workflow dict and insert SQL
```
Mark90, line 178: Nice, maybe a small comment that this can be removed in the next major release? (5.x or greater)

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.0.0
+current_version = 4.0.0rc1
```
Mark90, line 2: This should be 4.0.1

orchestrator/migrations/helpers.py
Igoranze: Not sure if this can ever be removed, since if you start a new project from the codebase, you will always encounter the issue about is_task does not exist on fresh DB

pboers1988 (approved):
LGTM

orchestrator/migrations/helpers.py
RichieB2B: Keeping the helpers backwards compatible seems to be the right approach, especially since squashing alembic migrations is not a trivial excersize. I'm also interested how @Mark90 was doing migrations of existing WFO installs after the 4.0.0 upgrade. We ran into issues with `pytest` that we fixed by running `upgrade schema@head` before `upgrade data@head` instead of `upgrade heads`.

PR #960, opened 2025-05-30
957 Check for table

Igoranze:


Mark90 (approved):

orchestrator/migrations/helpers.py
```diff
@@ -30,11 +30,16 @@ def has_table_column(table_name: str, column_name: str, conn: sa.engine.Connecti
...
+    :return: True if the column exists, False otherwise
     """
     inspector = sa.inspect(conn.engine)
-    columns = inspector.get_columns(table_name)
-    return any(col["name"] == column_name for col in columns)
+    try:
+        columns = inspector.get_columns(table_name)
+        return any(col["name"] == column_name for col in columns)
+    except sa.exc.NoSuchTableError:
+        # On some migrations the table might not exist yet, so we catch the exception
```
mrijk, line 40: just curious how this can happen. Doesn't this mean that there is an ordering problem in the migrations?

Igoranze: Added more details to the function. Basically, when we introduces target.VALIDATE we needed backwards compatibility with the previous migration helpers. This adds a check to see if a table does or does nog exist, and thus the column in that table.

PR #970, opened 2025-06-20
Move from flit/pip to uv

Mark90:
# Summary

This PR introduces [uv](https://docs.astral.sh/uv/) for creating and maintaining your development environment when working on the orchestrator-core.

Initial setup and basic usage has been documented in `docs/contributing/development.md` which are published here: https://workfloworchestrator.org/orchestrator-core/contributing/development/. Please follow these steps to get started with uv.

# Differences to flit/pip

Highlighting the key differences through a few practical examples.

## Updating local environment

This is what you want to do before getting started on a new feature or bugfix.

Before:

```shell
rm -rf .venv
python -m venv .venv
source .venv/bin/activate
pip install -U pip
pip install flit
flit install --deps develop
``` 

After:

```shell
uv sync --all-groups --all-extras
```

Side bonus: uv downloads and installs the correct Python version. No more messing around with pyenv! Though if you want to, [you still can](https://docs.astral.sh/uv/reference/cli/#uv-python).

## Upgrading dependencies

Before:

* Recreating the virtualenv would implicitly "upgrade" dependencies to the latest version

After:

* Appending `--upgrade` to the `uv sync` command will explicitly upgrade any dependencies, taking version constraints from pyproject.toml into account

The benefit of dependencies not being upgraded automatically is a consistent environment for all developers, and CI not breaking overnight.

## Speed..

On my machine (M4 MBP), recreating the .venv as mentioned above (with pip reading all packages from cache) takes 35 seconds.

With uv, creating the .venv from scratch (with uv also reading packages from it's cache to make the comparison fair) takes 2 seconds. However, if the .venv already exists, the time drops below 0.1 seconds.

This may not seem that important, but this adds up especially in CI pipelines. And it just feels nice 😎 

# Migration approach

I backed up the current `pyproject.toml` and generated a new one with:

```
uv init --build-backend flit --library --python 3.11 --bare

# --build-backend flit: use flit-core as build backend
# --library: intended to be built and distributed as a python package
# --bare: only create pyproject.toml
# --python 3.11: minimum supported version
```

Then manually added the metadata from the old `pyproject.toml`, updating to newer conventions where necessary.

After that I added the dependencies, optional dependencies (dev/docs) and dependency groups (celery) using `uv add`.


pboers1988 (approved):
Lgtm

tjeerddie (approved):

Issue #898, opened 2025-03-27
[Bug]: Improvement of the idempotency of the celery workers


pboers1988:
### Contact Details

peter.boers@surf.nl

### What happened?

When running the orchestrator in celery mode, the start stopping of tasks is not idempotent. There are some cases where tasks can be stuck in a "resumed" state and never picked up. Furthermore the running_processes counter in the database often does not reflect the accurate state of affairs.

### Steps to reproduce

1. submit a large amount of workflows onto a queue
2. kill celery
3. A large number of tasks will remain in the `created` state.

It is also possible to do this when retrying a large number of workflows, they will then become stuck in `resumed` state. Or killing celery whilst executing long running tasks and start/stopping the engine

### Sugested solution

As we heve implemented #832, we can now make the pick-up of tasks idempotent. Please design and implement a method so celery, at startup re-fills it's queue correctly.

### Version

3.1.2rc4 and lower

### What python version are you seeing the problem on?

All

### Relevant log output

```shell

```

Igoranze:
Forgot to mention, but you will also need a redis instance up and running, i'm using this docker compose file:

```
version: '3'

services:
  redis:
    image: redis
    ports:
      - "127.0.0.1:6379:6379"

volumes:
  redis-ssl-data:
```

kvklink:
We are experiencing the issue with workflows getting stuck in a 'created' state as well when we fire off our nightly subscription validation. Funny enough, it's always exactly four workflows that are stuck.

<details> 
<summary>Trace of how it fails, if relevant.</summary>

```
2025-06-19 08:50:35 [error    ] Unknown workflow failure       [orchestrator.services.processes] process_id=UUID('7a485fdd-5aa5-426c-bde4-f0eac9145b82') environment=prod
Traceback (most recent call last): environment=prod
  File "/app/orchestrator/services/processes.py", line 302, in safe_logstep environment=prod
    return _db_log_step(stat, step, process_state, broadcast_func=broadcast_func) environment=prod
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ environment=prod
  File "/app/orchestrator/services/processes.py", line 269, in _db_log_step environment=prod
    p = _update_process(stat.process_id, step, process_state) environment=prod
[...]

PR #979, opened 2025-06-27
Fix test test_retry_authorization

tjeerddie:
the test raises:
```
ValueError: Failed to write failure step to process: process with PID f4e1a612-e1f0-4432-b109-16de73637dcd not found
```

because tests have a scoped DB session that doesn't actually commits to the DB, so when the test was resumed with a intentional exception, it also rolls back the created process, resulting in the process not being found after rollback and trying to save the step as failed.

I fixed it by creating the process outside the test with a fixture.

fixes: https://github.com/workfloworchestrator/orchestrator-core/pull/939#issuecomment-2942235944
Related: https://github.com/workfloworchestrator/orchestrator-core/issues/898

### DIFF
--- a/test/unit_tests/api/test_processes.py
+++ b/test/unit_tests/api/test_processes.py
@@ -597,52 +597,8 @@ def test_unauthorized_to_run_process(test_client):
         assert HTTPStatus.FORBIDDEN == response.status_code
 
 
-def test_inputstep_authorization(test_client):
-    def disallow(_: OIDCUserModel | None = None) -> bool:
-        return False
-
-    def allow(_: OIDCUserModel | None = None) -> bool:
-        return True
-
-    class ConfirmForm(FormPage):
-        confirm: bool
-
-    @inputstep("unauthorized_resume", assignee=Assignee.SYSTEM, resume_auth_callback=disallow)
-    def unauthorized_resume(state):
-        user_input = yield ConfirmForm
-        return user_input.model_dump()
-
-    @inputstep("authorized_resume", assignee=Assignee.SYSTEM, resume_auth_callback=allow)
-    def authorized_resume(state):
-        user_input = yield ConfirmForm
-        return user_input.model_dump()
-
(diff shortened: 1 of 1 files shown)


test/unit_tests/api/test_processes.py
```diff
@@ -607,81 +618,69 @@ def allow(_: OIDCUserModel | None = None) -> bool:
...
-        # No auth succeeds
-        response = test_client.put(f"/api/processes/{process_id}/resume", json=[{"confirm": True}])
-        assert HTTPStatus.NO_CONTENT == response.status_code
-        # Authorized succeeds
-        response = test_client.put(f"/api/processes/{process_id}/resume", json=[{"confirm": True}])
-        assert HTTPStatus.NO_CONTENT == response.status_code
-        # Unauthorized fails
-        response = test_client.put(f"/api/processes/{process_id}/resume", json=[{"confirm": True}])
-        assert HTTPStatus.FORBIDDEN == response.status_code
+        return init >> authorized_resume >> next_step >> done
```
Mark90, line 632: I don't see the `unauthorized_resume` step from before?

tjeerddie: added back

test/unit_tests/api/test_processes.py
```diff
@@ -607,81 +618,87 @@ def allow(_: OIDCUserModel | None = None) -> bool:
...
+
+    return process_on_resume
+
+
+def test_authorized_resume_input_step(test_client, process_on_resume):
+    response = test_client.put(f"/api/processes/{process_on_resume}/resume", json=[{"confirm": True}])
+    assert HTTPStatus.NO_CONTENT == response.status_code
+
+
+def xtest_unauthorized_resume_input_step_retry(test_client, process_on_retry):
```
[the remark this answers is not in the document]
tjeerddie: nope
```suggestion
def test_unauthorized_resume_input_step_retry(test_client, process_on_retry):
```

PR #984, opened 2025-07-02
update task_resume_workflows to also resume processes in CREATED/RESUMED status

tjeerddie:
- improve start_process to include logic that happens in both executors.
- Change process status updates to be done within a transaction to lock the process in DB in executor functions to prevent race conditions:
  - change `_celery_set_process_status_resumed ` to a transaction to lock process in the DB, preventing more then one worker to pick up the process.
    - also use can_be_resumed to check that only correct statuses can be changed.
  - add _set_process_status_running to `thread_start_process` and `thread_resume_process` to update process status to `RUNNING` within a transaction and make sure that its status isn't `RUNNING` (picked up by another worker).
- move threadpool executor functions to its own file. 
  - celery: `orchestrator/services/celery.py` -> `orchestrator/services/executors/celery.py`
  - thread: `orchestrator/services/processes.py` -> `orchestrator/services/executors/threadpool.py`
    - `thread_start_process`
    - `thread_resume_process`
    - `thread_validate_workflow`
    - `THREADPOOL_EXECUTION_CONTEXT`
- move can_be_resumed from api/processes.py to services/processes.py.
- retrieve input state in thread_start_process and thread_resume_process to be able to restart the processes when the process gets stuck in `CREATED` or `RESUMED` right after initial form or form step.
- improve workflow removed error messages.

updated testing:
- add unit tests for threadpool and celery executor functions.
- add unit tests for resume_process.
- improve unit tests for start_process.
- add 2 tests that validate the happy flow of start_process and resume_process.
- change api resume workflow with incorrect status test to check all incorrect statuses.
- update task_resume_workflow tests.

Related: https://github.com/workfloworchestrator/orchestrator-core/issues/898


kvklink (approved):

orchestrator/services/executors/celery.py
```diff
@@ -80,7 +67,7 @@ def _celery_resume_process(
     user_inputs: list[State] | None = None,
     user: str | None = None,
     **kwargs: Any,
-) -> UUID:
+) -> bool:
```
Mark90, line 70: Now that we're revisiting things; this function calls:
`store_input_state(pstat.process_id, user_inputs, "user_input")`

Could you move this to `resume_process()` in the `processes.py` file? I think it should be done there after the `post_form` validation, with a `if user_inputs:` check to prevent inserting empty input.

This would make it consistent with how it's done for new processes (`start_process()` calls `create_process()` which then does `store_input_state(process_id, state | initial_state, "initial_state")`).

orchestrator/workflows/tasks/resume_workflows.py
```diff
@@ -24,19 +22,37 @@
...
+
 @step("Find waiting workflows")
 def find_waiting_workflows() -> State:
-    waiting_processes = db.session.scalars(
-        select(ProcessTable).filter(ProcessTable.last_status == ProcessStatus.WAITING)
-    ).all()
-    waiting_process_ids = [str(process.process_id) for process in waiting_processes]
-    return {"number_of_waiting_processes": len(waiting_process_ids), "waiting_process_ids": waiting_process_ids}
+    created_process_ids = get_process_ids_by_process_statuses([ProcessStatus.CREATED])
+    resumed_process_ids = get_process_ids_by_process_statuses([ProcessStatus.RESUMED])
```
Mark90, line 34: I'm not completely sure if this is the way to go.

Imagine this scenario:
- There are separate queues for 1) new workflows (status CREATED) 2) resumed workflows (status RESUMED)
- Each queue has 50 workflows queued
- For each queue there are 5 workers churning through queued workflows which they have put on status RUNNING
- When `task_resume_workflows` runs it finds 45 processes on CREATED and 45 on RESUMED so it calls `resume_process()` for each of them

Unless I'm missing something, this would re-queue 90 processes that don't need to be re-queued.

In `thread_start_process()` you already changed it to set the workflow to `RUNNING` before executing any step. This means that chance of the celery worker going AWOL before setting it to RUNNING has become much smaller, so that is already much better. If a worker dies when a process is running, it can be resumed by an engine restart.

So, I don't think we should resume created/resumed workflows in this task.
What do you think?

orchestrator/services/processes.py
```diff
@@ -493,57 +480,12 @@ def start_process(
...
-
+    pstat = create_process(workflow_key, user_inputs=user_inputs, user=user)
+    if not pstat.workflow.authorize_callback(user_model):
+        raise_status(HTTPStatus.FORBIDDEN, error_message_unauthorized(pstat.workflow.name))
 
-def thread_validate_workflow(validation_workflow: str, json: list[State] | None) -> UUID:
-    return thread_start_process(validation_workflow, user_inputs=json)
-
-
-THREADPOOL_EXECUTION_CONTEXT: dict[str, Callable] = {
```
Mark90, line 542: Maybe the PR should mention that these dicts were moved. And that the signature of the functions has changed. Just in case someone is using them directly

orchestrator/services/tasks.py
```diff
@@ -118,17 +95,15 @@ def resume_process(process_id: UUID, user_inputs: list[State] | None, user: str)
 
     celery_task = partial(celery.task, log=local_logger, serializer="orchestrator-json")
 
-    @celery_task(name=NEW_TASK)  # type: ignore
-    def new_task(process_id, workflow_key: str, user: str) -> UUID | None:
+    @celery_task(name=NEW_TASK, time_limit=1)  # type: ignore
```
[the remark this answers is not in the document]
tjeerddie: woops
```suggestion
    @celery_task(name=NEW_TASK)  # type: ignore
```

orchestrator/workflows/tasks/resume_workflows.py
tjeerddie: I think this is fixed by limiting which statusses can be changed to `RESUMED` by the `_celery_resume_process`. it should only change it from `SUSPENDED`, `WAITING`, `FAILED`, `AWAITING_CALLBACK`, `API_UNAVAILABLE` and `INCONSISTENT_DATA`. so it can never incorrectly change the status, removing the race condition.

orchestrator/services/tasks.py
```diff
@@ -118,17 +95,15 @@ def resume_process(process_id: UUID, user_inputs: list[State] | None, user: str)
...
+    def new_task(process_id: UUID, workflow_key: str, user: str) -> UUID | None:
         local_logger.info("Start task", process_id=process_id, workflow_key=workflow_key)
-        state = retrieve_input_state(process_id, "initial_state").input_state
-        return start_process(process_id, workflow_key, state=state, user=user)
+        return start_process(process_id, user=user)
```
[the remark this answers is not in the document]
tjeerddie: user input is added in `thread_start_process` and `thread_resume_process`

orchestrator/workflows/tasks/resume_workflows.py
tjeerddie: also changed process status updates to be done within a transaction to lock the process in DB in executor functions to prevent race conditions

orchestrator/services/executors/threadpool.py
```diff
@@ -0,0 +1,113 @@
...
+
+    result = db.session.execute(stmt)
+    locked_process = result.scalar_one_or_none()
+
+    if locked_process and locked_process.last_status is not ProcessStatus.RUNNING:
+        locked_process.last_status = ProcessStatus.RUNNING
+        db.session.commit()
+    else:
+        db.session.rollback()
+        raise Exception("Process is already running")
```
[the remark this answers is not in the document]
tjeerddie: yea, this makes sense, same in celery

orchestrator/migrations/versions/schema/2025-07-28_850dccac3b02_update_description_of_resume_workflows_.py
```diff
@@ -0,0 +1,35 @@
...
+down_revision = "93fc5834c7e5"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(
+        """
+        UPDATE tasks
+        SET description = 'Resume all workflows that are stuck on tasks with the status ''waiting'', `created` or `resumed`'
```
Mark90, line 22: Single quotes or backticks? :)

orchestrator/workflows/tasks/resume_workflows.py
```diff
@@ -46,14 +68,44 @@ def resume_found_workflows(waiting_process_ids: list[UUIDstr]) -> State:
             processes.resume_process(process)
             resumed_process_ids.append(process_id)
         except Exception:
-            logger.exception()
+            logger.warning()
```
Mark90, line 71:
```suggestion
            logger.warning("Could not resume process", process_id=process_id, error=str(exc))
```

and change `        except Exception as exc:` above

orchestrator/workflows/tasks/resume_workflows.py
```diff
@@ -46,14 +68,44 @@ def resume_found_workflows(waiting_process_ids: list[UUIDstr]) -> State:
...
+        try:
+            process = db.session.get(ProcessTable, process_id)
+            if not process:
+                continue
+            # Workaround the commit disable function
+            db.session.info["disabled"] = False
+            processes.restart_process(process)
+            started_process_ids.append(process_id)
+        except Exception:
+            logger.warning()
```
Mark90, line 95:
```suggestion
            logger.warning("Could not restart process", process_id=process_id, error=str(exc))
```

and change `        except Exception as exc:` above

Mark90 (approved):
LGTM, 3 small nitpicks (sorry) and looks like the branch is out of date.

Great work! 😄

Issue #955, opened 2025-05-26
[Feature]: Add Role based access control to running workflows - update subscription workflows endpoint


Mark90:
## General Requirements

* Ability to assign privileges per group
* Ability to check privileges halfway a workflow
* Make it possible to check authorization at each input type step
  like suspend, callback input steps (iow when a user/proces changes the workflow into running state)
* Dropped requirement: ability to change the minimal rights in the UI. This should be a handled by the auth implementation of the application

## Suggested solution

* Modify the endpoint `/api/subscriptions/workflows/<id>` to return `is_allowed` based on the 
* Use similar logic as #821 

Example:

```json
{
    "create": [
        {
            "name": "create_ip_lir_prefix",
            "description": "Create IP Lir Prefix",
            "reason": "subscription.no_modify_invalid_status",
            "usable_when": [],
            "status": "active",
            "action": "modified",
            "is_allowed": true     // <--- new field
        }
    ],
```

### Relevant pseudo code

```shell

```

Issue #956, opened 2025-05-28
[Feature]: Add Role based access control to running workflows - Process Detail Retry


Mark90:
### What should we build?

(issue yet to be refined)

Think about a way to disable the Retry button if the authenticated user is not authorized for it.

![Image](https://github.com/user-attachments/assets/09757ec8-2d55-4f1a-aeed-542182a68447)

### Relevant pseudo code

```shell

```

Mark90:
This may be a bit trickier to implement given that the API will need to figure at which step the workflow failed, and then evaluate the auth/retry_auth handler of the corresponding inputstep - or that of the entire workflow.

Mark90:
Implemented in #990 (not sure why github didn't auto-close this)

PR #990, opened 2025-07-08
Add RBAC info to process and subscription detail endpoints

kvklink:
This PR will allow to frontend to evaluate user permissions in order to disable certain input fields.

Related: https://github.com/workfloworchestrator/orchestrator-ui-library/pull/2058

Closes: #955, #956

### DIFF
--- a/orchestrator/api/api_v1/endpoints/subscriptions.py
+++ b/orchestrator/api/api_v1/endpoints/subscriptions.py
@@ -47,10 +47,12 @@ from orchestrator.services.subscriptions import (
     subscription_workflows,
 )
 from orchestrator.settings import app_settings
+from orchestrator.targets import Target
 from orchestrator.types import SubscriptionLifecycle
 from orchestrator.utils.deprecation_logger import deprecated_endpoint
 from orchestrator.utils.get_subscription_dict import get_subscription_dict
 from orchestrator.websocket import sync_invalidate_subscription_cache
+from orchestrator.workflows import get_workflow
 
 router = APIRouter()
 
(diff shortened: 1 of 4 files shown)


orchestrator/api/api_v1/endpoints/subscriptions.py
```diff
@@ -181,7 +185,14 @@ def subscription_workflows_by_id(subscription_id: UUID) -> dict[str, list[dict[s
     if not subscription:
         raise_status(HTTPStatus.NOT_FOUND)
 
-    return subscription_workflows(subscription)
+    subscription_workflows_dict = subscription_workflows(subscription)
+    for workflow_target in Target.values():
+        for workflow_dict in subscription_workflows_dict[workflow_target.lower()]:
+            workflow = get_workflow(workflow_dict["name"])
+            if workflow.authorize_callback and not workflow.authorize_callback(current_user):  # type: ignore
+                workflow_dict["reason"] = "subscription.insufficient_workflow_permissions"
```
Mark90, line 193: Would be nice to put this in a function `authorized_subscription_workflows()` or something (in that same services/subscriptions.py module) which takes the OIDCUserModel as parameter and returns this as a result. Keeps the endpoint code clean and makes it a little less work to reuse in the future (which I can imagine we will)

Also some comments on the code itself;
- If the inner for-loop is made to iterate over `subscription_workflows_dict.values()` then I think you can get rid of the outer for-loop
- `workflow_dict["reason"]` can already be set, for example if the subscription is out of sync. In this case we don't want to override this

orchestrator/graphql/types.py
```diff
@@ -132,6 +132,12 @@ class GraphqlFilter:
 }
 
 
+@strawberry.type(description="User permissions on a specific process")
+class FormUserPermissions:
```
Mark90, line 136: Nit: the strawberry graphql types are usually suffixed with `Type` :) 
```suggestion
class FormUserPermissionsType:
```

Mark90 (changes requested):
Looks good, 1 comment and 1 nitpick!

tjeerddie (approved):

Issue #888, opened 2025-03-17
[Bug]: Migrate domain models does not set subscription_instance_relations.domain_model_attr


Mark90:
https://github.com/workfloworchestrator/orchestrator-core/blob/271ac32d2e122034bb71aedce402495fcc766843/orchestrator/cli/domain_gen_helpers/product_block_helpers.py#L204

After adding a new product block, the [migrate-domain-models](https://workfloworchestrator.org/orchestrator-core/reference-docs/cli/#migrate-domain-models) command can be used to compare domain models to the database and generate migrations to bring the database in sync.

One of these migrations creates `subscription_instance_relations` records to relate existing subscription instances to the new subscription instances. However, the generated migrations currently don't set the `domain_model_attr` column. For example, the SQL query looks like this:

```sql
INSERT INTO subscription_instance_relations (in_use_by_id, depends_on_id, order_id) 
VALUES (
 '0c40000c-3e70-42cd-a6ab-2008468fd4eb', 
 (SELECT subscription_instances.subscription_instance_id FROM subscription_instances WHERE subscription_instances.product_block_id IN (SELECT product_blocks.product_block_id FROM product_blocks WHERE product_blocks.name IN ('test block')) AND subscription_instances.subscription_id = '019db62f-1d8c-499e-b172-048c1f9ed126'  LIMIT 1), 
 0
)
```

This is automatically corrected by the next time `SubscriptionModel.save()` is called (e.g. during a validation) in which fills the `domain_model_attr`. But until that save has been run, the database is not in sync. Moreover, when a product block refers to another product block more than once in different fields, then this cannot be reliably loaded deserialized by the domain model. For example:

```python
class ProductBlock1(ProductBlockModel):
  pb2_a: ProductBlock2
  pb2_b: ProductBlock2
```

The solution is to change the generator so that the aforementioned generated sql query specify the domain model attr:

```sql
INSERT INTO subscription_instance_relations (in_use_by_id, depends_on_id, order_id, domain_model_attr) 
VALUES (
 '0c40000c-3e70-42cd-a6ab-2008468fd4eb', 
 (SELECT subscription_instances.subscription_instance_id FROM subscription_instances WHERE subscription_instances.product_block_id IN (SELECT product_blocks.product_block_id FROM product_blocks WHERE product_blocks.name IN ('test block')) AND subscription_instances.subscription_id = '019db62f-1d8c-499e-b172-048c1f9ed126'  LIMIT 1), 
 0, 
[...]

PR #993, opened 2025-07-10
Change migrate domain model to save product block instances with a domain_model_attr

tjeerddie:
It will be a lot more work to support the behaviour of saving a block relation onto another block multiple times, since we would need to create a instance for each prop, example of a block related multiple times:
```
Class BlockOne(ProductBlockModel):
    example_field: str

Class BlockTwo(ProductBlockModel):
    block: BlockOne
    another_block: BlockOne

# it is possible to create the relation as a list (Recommended)
Class BlockTwo(ProductBlockModel):
    blocks: list[BlockOne]
```
I also wouldn't recommend creating the relation multiple times and if it is needed, you can use the migrator to create the migration that creates an instance for the first property found or create the migration yourself.

- now checks if a block has been related multiple times and will show a detailed warning that it's not supported by the migrator. you are given the option to continue with the first attribute. example:
  ```
  WARNING: Relating a Product Block multiple times is not supported by this migrator!
  You will need to create your own migration to create a Product Block Instance for each attribute that is related
  Product Block 'test block' has been related multiple times to 'ProductBlockOneForTest'
  Attributes the block ('test block') has been related with: new_block_4, new_block_2, new_block_1, new_block_3
  The relation will only be added to the first attribute ('new_block_4') want to continue?
  1) yes
  2) no
  ?
  ```

closes: https://github.com/workfloworchestrator/orchestrator-core/issues/888


orchestrator/cli/domain_gen_helpers/resource_type_helpers.py
```diff
@@ -266,6 +266,10 @@ def _find_new_relations(block_name: str, relations: dict[str, set[str]]) -> set[
     return set(flatten((list(v) for k, v in relations.items() if block_name in k)))
 
 
+def _find_new_block_relations(block_name: str, relations: dict[str, list[BlockRelationDict]]) -> set[str]:
+    return set(flatten([r["name"] for r in v] for k, v in relations.items() if block_name in k))
```
Mark90, line 270: Why `block_name in k`?

tjeerddie: used the old logic above this fn, but you are right, no reason for this and `relations.get(block_name, set())` can be used instead

PR #1004, opened 2025-07-17
Update error_state_to_dict to handle headers with None

tjeerddie:
- update unit tests for error_state_to_dict, with newer versions of openapi generator, the ApiException can have headers with None instead of an empty dict, this breaks the error_state_to_dict.


pboers1988 (approved):
LGTM

PR #1028, opened 2025-08-12
Llm integration POC

timfdev:
`pydantic-ai` and `ag-ui-protocol` 

need `pydantic >= 2.10` and `>=2.11.2` respectively, this breaks some of the unit tests


orchestrator/search/retrieval/builder.py
```diff
@@ -0,0 +1,76 @@
+from sqlalchemy import Select, Table, and_, select
+from sqlalchemy_utils.types.ltree import Ltree
+
+from orchestrator.db.models import AiSearchIndex
+from orchestrator.search.filters import FilterSet, LtreeFilter
+from orchestrator.search.schemas.parameters import BaseSearchParameters
+
+
+class QueryBuilder:
```
luc-tielen, line 9: A bit nitpicky, but: this technically doesn't need to be a class, can be done with plain functions (you're not really using `self`).

orchestrator/search/agent/prompts.py
```diff
@@ -0,0 +1,62 @@
...
+    try:
+        schema_dict = get_structured_filter_schema()
+        if schema_dict:
+            schema_info = "\n".join([f"    {path}: {field_type}" for path, field_type in schema_dict.items()])
+        else:
+            schema_info = "    No filterable fields available"
+    except Exception as e:
+        logger.warning(f"Failed to load schema for prompt: {e}")
+        schema_info = "    Schema temporarily unavailable"
+    logger.error(f"Generated schema for agent prompt:\n{schema_info}")
```
tjeerddie, line 26: I don't think this is suppose to be an error log?

tjeerddie (approved):

orchestrator/search/retrieval/retriever.py
```diff
@@ -0,0 +1,447 @@
...
+                )
+                .label("sem_rank"),
+                func.dense_rank()
+                .over(order_by=[entity_scores.c.avg_fuzzy_score.desc().nulls_last(), entity_scores.c.entity_id.asc()])
+                .label("fuzzy_rank"),
+            ).select_from(
+                entity_scores.join(entity_highlights, entity_scores.c.entity_id == entity_highlights.c.entity_id)
+            )
+        ).cte("ranked_results")
+
```
Mark90, line 383: Could we split this function up in one for the DB interaction part which produces an output, and another function that performs the below computations based on the former's output? And preferably also some unittests for the latter

orchestrator/search/retrieval/retriever.py
```diff
@@ -0,0 +1,447 @@
+from abc import ABC, abstractmethod
```
Mark90, line 1: Maybe split up into a package with a module for each retriever type, it's a lot of scrolling now :)

orchestrator/search/retrieval/retriever.py
```diff
@@ -0,0 +1,447 @@
...
+        Returns:
+        -------
+        Select
+            A new `Select` statement with ranking expressions applied.
+        """
+        ...
+
+    def _quantize_score_for_pagination(self, score_value: float) -> BindParameter[Decimal]:
+        """Convert score value to properly quantized Decimal parameter for pagination."""
+        pas_dec = Decimal(str(score_value)).quantize(Decimal("0.000000000001"))
```
Mark90, line 107: Should this change along with the SCORE_PRECISION if that ever changes?

If so maybe do something like `f'{1 / 10**precision:.{precision}f}`

orchestrator/search/retrieval/engine.py
```diff
@@ -0,0 +1,152 @@
...
+    if not db_rows:
+        return SearchResponse(results=[], metadata=metadata)
+
+    user_query = search_params.query
+
+    results = []
+    for row in db_rows:
+        matching_field = None
+
+        if user_query and row.get("highlight_text") and row.get("highlight_path"):
```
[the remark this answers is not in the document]
Mark90: And I think the `if` statement  can use `Retriever.HIGHLIGHT_TEXT_LABEL` instead of the string value?

orchestrator/search/retrieval/utils.py
```diff
@@ -0,0 +1,106 @@
...
+    all_matches = []
+    words = [w.strip() for w in term.split() if w.strip()]
+
+    for word in words:
+        word_boundary_pattern = rf"\b{re.escape(word)}\b"
+        matches = list(re.finditer(word_boundary_pattern, text, re.IGNORECASE))
+
+        if not matches:
+            substring_pattern = re.escape(word)
+            matches = list(re.finditer(substring_pattern, text, re.IGNORECASE))
```
Mark90, line 32: If a resulting text has both word _and_ substring matches, wouldn't we want to highlight the substring matches as well?

orchestrator/search/schemas/results.py
```diff
@@ -0,0 +1,63 @@
...
+
+    model_config = ConfigDict(
+        extra="forbid",
+        use_enum_values=True,
+    )
+
+
+class TypeDefinition(BaseModel):
+    operators: list[FilterOp]
+    valueSchema: dict[FilterOp, ValueSchema]
```
Mark90, line 61: Is camelCase needed here?

Mark90 (commented):
Nice, that's a lot of work 🔥 

Overall structure of the code is good, that's why I was able to leave a lot of questions and small remarks. I mean this as a good thing :)

orchestrator/api/api_v1/endpoints/search.py
```diff
@@ -0,0 +1,277 @@
...
+    try:
+        pagination_params = await process_pagination_cursor(cursor, search_params)
+    except InvalidCursorError:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="Invalid pagination cursor")
+
+    search_response = await execute_search(
+        search_params=search_params,
+        db_session=db.session,
+        pagination_params=pagination_params,
+    )
```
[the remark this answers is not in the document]
luc-tielen: This is planned for later.

orchestrator/search/schemas/results.py
timfdev: Sharp one, no I think this is left over from something I tried with pydantic aliases to use camelCase in the response, but it was hard to keep that consistent for deep nested data.

Issue #1026, opened 2025-08-11
[Feature]: Use a Modify workflow as a Reconcile workflow


Mark90:
### What should we build?

As an Orchestrator User I want to be able to correct information in external systems with 1 click by running a Reconcile workflow.

A "Reconcile workflow" is nothing more than calling the Modify workflow that will reapply current configuration from the database. (e.g. running the Modify workflow with only empty input forms)

**Implementation details**

- The reconcile workflow will only run steps that update external systems. It does **not** include the:
  - Input forms
  - Subscription updates in Core DB
- The definition of those steps should be reused by both the modify and reconcile workflow
- The reconcile workflow still requires the subscription to be in sync before it can be started
- Only allow 1 Reconcile workflow per product
  - See how this is implemented for Create workflows with `ProductTable.create_subscription_workflow_key`

**Intended usage in Orchestrator implementations**

Below is an example for the modify workflow of one specific production.
These steps should be repeated for each product that need (and support) a reconcile workflow.

Example based on example-orchestrator [modify_l2vpn](https://github.com/workfloworchestrator/example-orchestrator/blob/master/workflows/l2vpn/modify_l2vpn.py):

- Add migration to add a row in Workflows table with `target=RECONCILE` and `name="reconcile_l2vpn"`
- Add `LazyWorkflowInstance` mapping the module `modify_l2vpn` to `reconcile_l2vpn`:
```py
LazyWorkflowInstance("workflows.l2vpn.modify_l2vpn", "reconcile_l2vpn")
```

The `modify_l2vpn.py` should be modified as well.
It currently looks like this:

```py
@modify_workflow("Modify l2vpn", initial_input_form=initial_input_form_generator)
def modify_l2vpn() -> StepList:
    return (
        begin
        >> set_status(SubscriptionLifecycle.PROVISIONING)
        >> update_subscription
        >> update_subscription_description
        >> set_status(SubscriptionLifecycle.ACTIVE)
        >> update_l2vpn_in_nrm
    )
```

The steps responsible for updating external systems should be pulled out to a reusable variable:

```py
update_l2vpn_in_external_systems = (
        begin 
        >> update_l2vpn_in_nrm
)

@modify_workflow("Modify l2vpn", initial_input_form=initial_input_form_generator)
def modify_l2vpn() -> StepList:
    return (
        begin
        >> set_status(SubscriptionLifecycle.PROVISIONING)
        >> update_subscription
        >> update_subscription_description
[...]

PR #1048, opened 2025-08-28
New reconcile workflow decorator for subscription synchronization with external systems

tvdven:
## Add reconcile workflow decorator for subscription synchronization

This PR implements a new reconcile_workflow decorator that enables synchronization of subscription configurations with external systems. 

The reconcile workflow reuses the modify workflow but removes the need for user input forms, making it ideal for automated synchronization tasks.

## Summary
- Implement reconcile_workflow decorator that wraps modify_workflow functionality without requiring user input via forms
- Update subscription workflows to include reconcile workflows in the available workflow list
- Add unit tests for new reconcile workflow covering basic usage

Issue: #1026


kvklink (approved):
LGTM, but I think it would be nice to also have some user documentation for this feature

Mark90 (approved):

PR #1050, opened 2025-08-29
Fix blocked scheduler being stuck without any jobs

tjeerddie:
- add helper function to get pauzed scheduler to use for managing jobs in the db and correclty shut down afterwards.


orchestrator/schedules/scheduler.py
```diff
@@ -36,6 +37,17 @@ def scheduler_dispose_db_connections() -> None:
     jobstores["default"].engine.dispose()
 
 
+@contextmanager
+def get_pauzed_scheduler() -> Generator[BackgroundScheduler, Any, None]:
```
Mark90, line 41: "pauzed" is part Dutch, part English :) I suggest to write it as 'paused'

Mark90 (approved):
Small comment, LGTM

Issue #1056, opened 2025-09-05
[Bug]: apscheduler breaks when multiple tasks run at the same time


tjeerddie:
### What happened?

The `run` command should executes all scheduled jobs in a single process but currently this breaks with multiple  tasks that run at the same time. this is because tasks use the same DB session/transaction.

error when making 2 tasks run at the same time: `sqlalchemy.exc.InvalidRequestError: This session is in 'prepared' state; no further SQL can be emitted within this transaction.`

### Version

4.4.0rc2

PR #1057, opened 2025-09-08
Fix scheduler breaking when multiple tasks are scheduled at the same time

tjeerddie:
- Change scheduler to run with a single process
- remove new blocking scheduler for scheduler run command since it brings to many issues currently, blocking the thread with a while loop

Closes: https://github.com/workfloworchestrator/orchestrator-core/issues/1056


Igoranze (approved):

orchestrator/schedules/scheduler.py
```diff
@@ -29,8 +30,15 @@
 from orchestrator.utils.helpers import camel_to_snake, to_camel
 
 jobstores = {"default": SQLAlchemyJobStore(url=str(app_settings.DATABASE_URI))}
+executors = {
+    "default": ThreadPoolExecutor(1),
+    "processpool": ProcessPoolExecutor(1),
```
Mark90, line 35: Is this used anywhere?

Mark90 (approved):

orchestrator/schedules/scheduler.py
tjeerddie: yea processpool isn't necessary, thought I already removed it.
```suggestion
```

PR #1062, opened 2025-09-12
Await get_current_user in workflow is_allowed graphql field

tjeerddie:
get_current_user has type: `Awaitable[OIDCUserModel | None]`

- Also await get_current_user in process user_permissions graphql field


pboers1988 (approved):
LGTM

Issue #1003, opened 2025-07-17
[Bug]: Corelink create workflow step functions bypass lifecycle state validation and does not raise a ValueError


tvdven:
### Expected Behavior

- When a workflow step function parameter is annotated with a product lifecycle state
- The system should check that the subscription object matches that specific state
- If the subscription is in a different state, a ValueError should be raised

### Actual Behavior

- The workflow step function executes without validating the annotated product lifecycle state
- No ValueError is raised when the subscription object is in an incorrect state
- The workflow continues processing despite the state mismatch

### Steps to Reproduce

1. Add a new step function within Corelink create workflow
2. Annotate the function parameter of  with a specific product lifecycle state (e.g., `Sn8CorelinkProvisioning`)
3. Pass a subscription object that is NOT in that state
3. Execute the workflow

### Version

4.1.0

### What python version are you seeing the problem on?

Python 3.13

tvdven:
@pboers1988 can you review / add more context where necessary?

PR #1092, opened 2025-09-26
Option to perform subcription model lifecycle validation checks for product type in workflow steps

tvdven:
## Summary

The Lifecycle Validation Mode setting is used to validate in workflow steps that a subscription model has been instantiated with the correct product type class for its lifecycle status

E.g. a subscription model with a lifecycle status of `PROVISIONING` should be instantiated with a product type class that has a lifecycle status of `PROVISIONING`.

* Added lifecycle validation check
* Updated settings.py so that users can decide in which mode they want to run their app
* Wrote unit tests
* Added documentation

Issue: #1003

### DIFF
--- a/docs/reference-docs/app/settings-overview.md
+++ b/docs/reference-docs/app/settings-overview.md
@@ -77,6 +77,27 @@ my_settings = MySettings()
 expose_settings("my_settings", my_settings)
 ```
 
+## Lifecycle Validation Mode
+
+You can run the application with three different lifecycle validation modes:
+
+```Python
+class LifecycleValidationMode(strEnum):
+    STRICT = "strict"
+    LOOSE = "loose" # default
+    IGNORED = "ignored"
+```
+
+The Lifecycle Validation Mode can be set using the `LIFECYCLE_VALIDATION_MODE` environment variable. The default value is `loose`.
+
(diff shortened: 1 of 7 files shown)


docs/reference-docs/app/settings-overview.md
```diff
@@ -77,6 +77,27 @@ my_settings = MySettings()
...
+    IGNORED = "ignored"
+```
+
+The Lifecycle Validation Mode can be set using the `LIFECYCLE_VALIDATION_MODE` environment variable. The default value is `loose`.
+
+The Lifecycle Validation Mode setting is used to validate that a **subscription model has been instantiated with the correct product type class for its lifecycle status**. E.g. a subscription model with a lifecycle status of `PROVISIONING` should be instantiated with a product type class that has a lifecycle status of `PROVISIONING`.
+
+The different modes are explained below:
+
+- `strict`: In this mode, the application will enforce strict checks on the lifecycle status of subscription models. If any issues are found, the application will raise an error and stop running.
```
Mark90, line 97: Maybe we should include a warning, something like:

> A workflow failing on this error will not be recoverable, so it is advised to test this in development first.

orchestrator/domain/base.py
```diff
@@ -1390,7 +1385,7 @@ def from_subscription(cls: type[S], subscription_id: UUID | UUIDstr) -> S:
 
         status = SubscriptionLifecycle(subscription.status)
 
-        if not cls.__base_type__:
+        if not cls.__base_type__:  # provisioning of active
```
Mark90, line 1388: Leftover comment

```suggestion
        if not cls.__base_type__:
```

pboers1988 (approved):
Nice

PR #1100, opened 2025-10-01
make_async w run_in_threadpool wrapper for resolvers

Mark90:


orchestrator/graphql/resolvers/process.py
```diff
@@ -64,7 +65,8 @@ async def resolve_process(info: OrchestratorInfo, process_id: UUID) -> ProcessTy
     return None
 
 
-async def resolve_processes(
+@make_async
+def resolve_processes(
```
Mark90: Finally got around to testing this today, and it took me a while to realize why I didn't see any improvement.

Turns out that in our SURF orchestrator we set `ENGINE_ARGUMENTS["pool_size"] = 15` in the app initialization. (the default in the core is 60) This is low enough for the queries not to be the bottleneck - at least in my test environment with not that much data.

After increasing the pool size to 40, I got up to 30 rps (using the `ProcessDetail` query that the UI makes) with the main branch, and 40 concurrent requests on this `628-async-gql-option-2` branch.

I see we haven't documented tweaking the engine pool size anywhere, will make an issue for that. 

Also, I could easily overload the WFO API with too many requests, leading to these delayed errors:

> sqlalchemy.exc.TimeoutError: QueuePool limit of size 40 overflow 10 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/20/3o7r)

I think we should lower this timeout and make the REST/GQL API return an appropriate 503 response. This is not really related to this PR, I'll also make an issue for that.

Anyway, it's looking good so far, will test the other resolvers as well. :)

Mark90:
@Sparrow1029 it's looking good! Seeing as this is a temporary PR I made myself just to review, feel free recreate one yourself so I can approve it

PR #1118, opened 2025-10-09
Agent export

### LINKED ISSUE #1060: Let agent export results
Let agent export results (via GraphQL export)

### LINKED ISSUE #1083: Agent should not stream back records as part of the tool call response
The agent now streams the found records along with other results to the frontend. I think streaming the data itself is very inefficient. Maybe we should send the uuid's only and let the frontend retrieve those records. 

Or maybe the query itself but then  we would have to send the query embedding as well if needed otherwise it changes in a subsequent call..

### LINKED ISSUE #1099: [Feature]: Save queries and their embeddings
### Contact Details

_No response_

### What should we build?

Currently embeddings are encoded in the cursor which works but is not ideal as the payload is very big and we lose this cursor between different  sources. 

For example when an agent constructs a query, injects context and returns an answer , we might want to return this query, so the frontend can do a separate call to show these results to avoid streaming all the data over the agent route (#1083)

Additionally we can unlock new potential where we may show recent queries to users, save embedding API calls, do analysis on used queries etc...

### Relevant pseudo code

```shell

```

timfdev:
This PR closes #1060, closes #1083, closes #1099.

* Separation between searching and data fetching, it makes much more sense to return only search relevant data, full entity data fetching can be on demand through a graphql call. This massively improves search speed.

*  Saving search queries. Saving the query (pydantic model) with embedding for: pagination, exports, caching, showing agent results in frontend, analysis/query suggestions etc. The removed embedding from the encoded cursor reduces payload size by (~)95% when using 1536 dimensional vectors.

* Saving agent runs, can be used to reconstruct agent runs for analysis, persisting chats etc. preserving and collecting of all executed queries in a run.

* agent is now a router on the same level as other API routes , preventing naming conflicts, proper dependency injection

* Usage of StateDeltaEvents, which is more efficient because it only sends updated parts of the state [link]( https://docs.ag-ui.com/concepts/state)

* Refactored and simplified cursor pagination

* Agent now has a new export tool that sets a download url in the state, copilotkit then renders this as a button
* Agent also has a new tool to fetch detailed data if needed, because search results now contain lightweight data like id,score only. For now this is the same format as the export tool, we can change this later, possibly only allowing detailed fetching for single entities and letting the agent choose what data it wants.

### DIFF
--- a/orchestrator/cli/database.py
+++ b/orchestrator/cli/database.py
@@ -23,6 +23,8 @@ from structlog import get_logger
 
 import orchestrator.workflows
 from orchestrator.cli.domain_gen_helpers.types import ModelUpdates
+from orchestrator.cli.generate import create_writer, get_template_environment
+from orchestrator.cli.generator.generator.migration import create_data_head_if_not_exists
 from orchestrator.cli.helpers.print_helpers import COLOR, str_fmt
 from orchestrator.cli.migrate_domain_models import create_domain_models_migration_sql
 from orchestrator.cli.migrate_tasks import create_tasks_migration_wizard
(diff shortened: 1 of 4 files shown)


pboers1988 (approved):
LGTM

PR #1140, opened 2025-10-23
Revise scaling and scheduling docs

### LINKED ISSUE #535: [Feature]: Add medium priority documentation for Runtimes
### Contact Details

srichmond@es.net

### What should we build?

Please create reference documentation for the following topics on https://workfloworchestrator.org/orchestrator-core/ using the "how to write docs" information located in https://docs.google.com/document/d/10Fz98moy-pMyyk__PTR8xg4aeHuIv5VegJa99ZpzvkI/edit

- [ ] Synchronous
- [ ] Asynchronous
- [ ] Celery
- [ ] Scheduler

### Relevant pseudo code

_No response_

eenblam:
Requesting initial review - this has a few unresolved points. See my comments.

There was also discussion in the original ticket (#535) about possibly relocating  or combining these docs. I'm open to moving things, but so far I've instead tried to just better cross-link things to avoid breaking existing paths.

Closes #535.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.6.0rc3
+current_version = 4.6.0rc4
 commit = False
 tag = False
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(rc(?P<build>\d+))?
--- a/orchestrator/__init__.py
+++ b/orchestrator/__init__.py
@@ -13,7 +13,7 @@
 
 """This is the orchestrator workflow engine."""
 
-__version__ = "4.6.0rc3"
+__version__ = "4.6.0rc4"
 
 
 from structlog import get_logger
--- a/orchestrator/db/models.py
+++ b/orchestrator/db/models.py
@@ -711,7 +711,7 @@ class SearchQueryTable(BaseModel):
     parameters = mapped_column(pg.JSONB, nullable=False)
 
     # Query embedding for semantic search (pgvector)
-    query_embedding = mapped_column(Vector(1536), nullable=True)
+    query_embedding = mapped_column(Vector(llm_settings.EMBEDDING_DIMENSION), nullable=True)
 
     executed_at = mapped_column(UtcTimestamp, server_default=text("current_timestamp()"), nullable=False)
 
--- a/orchestrator/llm_settings.py
+++ b/orchestrator/llm_settings.py
@@ -18,7 +18,7 @@ from structlog import get_logger
 
 logger = get_logger(__name__)
 
-EMBEDDING_DIMENSION_MIN = 200
+EMBEDDING_DIMENSION_MIN = 100
 EMBEDDING_DIMENSION_MAX = 2000
 EMBEDDING_DIMENSION_DEFAULT = 1536
 


docs/reference-docs/app/scaling.md
```diff
@@ -42,6 +43,7 @@ The orchestrator-worker has additional dependencies which can be installed with
 pip install orchestrator-core[celery]
 ```
 
+TODO link to notes on how many instances to run of each
```
eenblam: Do we have existing notes on this? I know SURF has advice here, just not sure what it is and if it's already written down someplace I can link to.

docs/architecture/application/tasks.md
```diff
@@ -2,22 +2,41 @@
 
 This document covers the moving parts needed to schedule jobs in the orchestrator.
 
+In short, the scheduler is simply an additional process that adds Tasks to the queue.
+A Task is just a workflow that isn't tied to a specific product.
+Tasks are created in the same way as workflows, but with the `"system"` target, i.e.
+
+```python
+@workflow("Some task", target=Target.SYSTEM)`
```
Mark90:
```suggestion
@workflow("Some task", target=Target.SYSTEM)
```

docs/architecture/application/tasks.md
```diff
@@ -68,26 +91,38 @@ def upgrade() -> None:
     pass
 ```
 
-This just needs to add an entry in the workflows table. No relations with other tables like how the workflow id gets a relation in the products table and etc.
+Note that, unlike a normal workflow:
 
-### Running the task in the UI
+* `is_task` is set to `True`.
+* No relations with other tables are created. (For example, a normal workflow id gets a relation in the products table, etc.)
```
Mark90: Since you modified this :) I suggest we change this section to redirect users to "Registering Workflow"  https://workfloworchestrator.org/orchestrator-core/getting-started/workflows/#register-workflows which explains to either use the generator or the migration helper function. That page would need to be modified slightly to add that there's also a `create_task` helper in `orchestrator.migrations.helpers`

docs/architecture/application/tasks.md
```diff
@@ -142,7 +178,8 @@ USER www-data:www-data
 CMD /usr/src/app/bin/server
 # Comment out the previous command and uncomment the
 # following lines to build a version that runs the
-# backend and scheduer in the same container.
+# backend and scheduler in the same container.
+# (These are all generated when initializing your repository.)
 # COPY ./bin/server ./bin/server
 # COPY ./bin/scheduling ./bin/scheduling
 # COPY ./bin/wrapper ./bin/wrapper
```
Mark90: We can also remove this because I don't know which `bin/wrapper` this refers to

docs/reference-docs/app/scaling.md
```diff
@@ -191,8 +197,8 @@ class OrchestratorWorker(Celery):
         self.process_broadcast_fn = process_broadcast_fn
```
Mark90: While we're updating this; in our OrchestratorWorker the on_init starts off like this:

```
    def on_init(self) -> None:
        start_debugger()
```

Which is imported with `from nwastdlib.debugging import start_debugger`. This allows for interactive debugging.

docs/reference-docs/app/scaling.md
```diff
@@ -203,12 +209,18 @@ class OrchestratorWorker(Celery):
...
+backend = f"rpc://{AppSettings().CACHE_URI}/0"
 
+# TODO How is the default celery backend set up?
 celery = OrchestratorWorker(
-    f"{app_settings.SERVICE_NAME}-worker", broker=str(app_settings.CACHE_URI), include=["orchestrator.services.tasks"]
+    f"{app_settings.SERVICE_NAME}-worker", broker=broker, include=["orchestrator.services.tasks"]
 )
 
 if app_settings.TESTING:
+    #TODO Should this not be the `backend` from above, i.e. rpc://CACHE_URI/0?
```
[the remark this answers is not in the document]
eenblam: Right, I'm specifically looking for clarity on why we included the `rpc://` option in the code of the previous section.

docs/architecture/application/tasks.md
```diff
@@ -114,42 +148,70 @@ For detailed configuration options, see the [APScheduler scheduling docs].
...
-# CMD /usr/src/app/bin/wrapper
+# Dockerfile for scheduler image
+FROM your-orchestrator
+ENTRYPOINT ["python", "main.py"]
+CMD ["scheduler", "run"]
+```
+
+For the second option: suppose you start your app with a script, `bin/server`, that handles your migrations, kicks off uvicorn, etc.
+You can then replace your backend's Docker entrypoint with a script like this, `bin/wrapper`:
+
```
Mark90: Great!

docs/architecture/application/tasks.md
```diff
@@ -2,22 +2,44 @@
...
+```python
+@workflow("Some task", target=Target.SYSTEM)
+def some_task() -> StepList:
+    return init >> foo >> done
+```
+
+Such a workflow will be flagged as a task in the database, and will not have a relation defined connecting it to a specific product.
+
+Note that `@workflow` is a lower-level call than, say, `@create_workflow`.
+So we instead of `return begin >> foo`, we need to use `return init >> foo >> done` to instantiate a `StepList`.
```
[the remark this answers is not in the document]
eenblam: Thanks, good catch!

docs/reference-docs/app/scaling.md
eenblam: Thanks Peter! I replaced the section on autoscaling as you described.

docs/architecture/application/tasks.md
eenblam: This rearrangement seems like a larger conversation, so marking this resolved with respect to the current MR.

PR #1155, opened 2025-10-30
Fix/refactor query model

timfdev:


pboers1988:
Looks good, but should we start adding some tests for the query module?

PR #1162, opened 2025-11-04
Fix errors to be able to fetch soft deleted workflows that only exist in the DB

tjeerddie:
- Change get_auth_callbacks to work with optional workflow param
- Convert values to boolean for retryAllowed and resumeAllowed for graphql process user_permissions


test/unit_tests/graphql/test_process.py
```diff
@@ -39,4 +44,28 @@ def test_process(test_client, mocked_processes):
...
+def test_process_is_allowed_with_historic_workflow_only_left_in_db(
+    mock_get_workflow, test_client, mocked_processes, test_workflow_soft_deleted
+):
+    mock_get_workflow.return_value = None
+    process_id = mocked_processes[0]
+    test_query = build_simple_query(process_id)
+
+    response = test_client.post("/api/graphql", content=test_query, headers={"Content-Type": "application/json"})
+    assert response.status_code == HTTPStatus.OK
+    assert response.json() == {
```
mrijk: Why not only test the relevant part of the response? I guess processId is not that interesting, just the userPermissions

mrijk (approved):

pboers1988 (approved):

PR #1169, opened 2025-11-10
Do not embed empty string values

### LINKED ISSUE #1139: [Bug]: Semantic retriever sends empty match fields
Semantic retriever occasionally sends MatchingField models that are empty, now that the search is not returning detailed data anymore we rely on this model to show a title/subtitle in the UI for a search result.

Maybe has something to do with candidate size? Needs some refinement.

timfdev:
closes #1139 

The index process was embedding empty string values. 

indexer.py:

```python
    def _prepare_text_for_embedding(field: ExtractedField) -> str:
        return f"{field.path}: {str(field.value)}"
```

Because the embedding consists out of format: `path:value`, we d still match the path on a search.

For example when searching for `some note`, we d get a search result with MatchedField:

```
            "matching_field": {
                "text": "",
                "path": "subscription.note",
                "highlight_indices": null
```

`and (path := row.get(Retriever.HIGHLIGHT_PATH_LABEL)) is not None` makes sure that we have a matching field even if if the text is an empty string.

This fix requires an index run with `--force-index`

### DIFF
--- a/.github/workflows/build-push-container.yml
+++ b/.github/workflows/build-push-container.yml
@@ -1,9 +1,13 @@
 name: Build and push container image
 
 on:
+  # Build and push the image in releases
   release:
     types: [prereleased, released]
-
+  # Build the image in pull requests
+  pull_request:
+    paths:
+      - 'Dockerfile'
 env:
   REGISTRY: ghcr.io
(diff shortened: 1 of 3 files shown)


pboers1988 (approved):
Lgtm

PR #1238, opened 2025-11-25
Integration tests and benchmark

### LINKED ISSUE #1145: [Feature]:  Benchmark truncated vector size
### Contact Details

_No response_

### What should we build?

For some models we truncate the vector size on the assumption that the first part contains the most important context. A benchmark, with some known cases might be handy to remove the assumption.

### Relevant pseudo code

```shell

```

### LINKED ISSUE #1146: [Feature]:  Facilitate E2E test with search scenario's
### Contact Details

_No response_

### What should we build?

Currently we benchmark changes in the search by testing it in real world applications. As the LLM stuff and the vector stuff are kind of fuzzy and different models behave different it's easy to introduce a regression without knowing it.

So we could have a fixed data set, like the pokemon index: define some queries for it and test that expectation about the results are met.

- move pokemon stuff to core? (50 pokemons -> fixed set)
- define some searches and what results there should be
- write tests that can be run against a DB populated with the pokedex

Focus op:
- alle types werken (float, int, text, resource types + semantic)
- exact matches should work

timfdev:
closes #1145 , closes #1146

### DIFF
--- a/test/integration_tests/search/scripts/benchmark/display.py
+++ b/test/integration_tests/search/scripts/benchmark/display.py
@@ -72,6 +72,7 @@ def display_ranking_comparison(query_results, expected_ranking: list[str]) -> No
 
 def display_benchmark_results(results, queries, models) -> None:
     """Display benchmark results with simple metrics and visual comparisons."""
+    from statistics import median
 
     # Group results by model
     results_by_model = defaultdict(list)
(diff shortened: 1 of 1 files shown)


test/unit_tests/search/test_indexer.py
```diff
@@ -0,0 +1,306 @@
...
+
+@pytest.fixture
+def indexer(mock_config):
+    """Create an indexer instance for testing."""
+    indexer = Indexer(config=mock_config, dry_run=True, force_index=False, chunk_size=10)
+    indexer._entity_titles[ENTITY_ID] = ENTITY_TITLE
+    return indexer
+
+
+class TestIndexerContentHashing:
```
Mark90: Are these tests structured into classes purely to organize them per "topic"?

Issue #1084, opened 2025-09-24
[Bug]: Startup fails on new deployment with empty DB and Prometheus metrics enabled (ENABLE_PROMETHEUS_METRICS_ENDPOINT=True)


fililand:
### What happened?

Starting a brand new stack with empty DB and ENABLE_PROMETHEUS_METRICS_ENDPOINT=True, running 

```
python main.py db upgrade heads
```
results in a `ProgrammingError: relation "subscriptions" does not exist`. 
When ENABLE_PROMETHEUS_METRICS_ENDPOINT=False, the stack starts fine, and enabling metrics later does not trigger the error — even on an empty DB.

### Version

4.2.0

### What python version are you seeing the problem on?

Python 3.12

### Relevant log output

```shell
2025-09-24 09:58:23 [info     ] Database object configured, all methods referencing `db` should work. [orchestrator.db]
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1963, in _exec_single_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.12/site-packages/sqlalchemy/engine/default.py", line 943, in do_execute
    cursor.execute(statement, parameters)
  File "/usr/local/lib/python3.12/site-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
psycopg.errors.UndefinedTable: relation "subscriptions" does not exist
LINE 2: FROM subscriptions LEFT OUTER JOIN products ON products.prod...
             ^

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/garr/main.py", line 48, in <module>
    app = init_app(app_settings)
          ^^^^^^^^^^^^^^^^^^^^^^
  File "/garr/main.py", line 22, in init_app
    app = OrchestratorCore(base_settings=orchestrator_settings)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/orchestrator/app.py", line 149, in __init__
    initialize_default_metrics()
  File "/usr/local/lib/python3.12/site-packages/orchestrator/metrics/init.py", line 12, in initialize_default_metrics
    ORCHESTRATOR_METRICS_REGISTRY.register(SubscriptionCollector())
[...]

PR #1366, opened 2026-02-02
fix: update metrics initialisation to handle missing DB tables

Haak1981:
When setting param ENABLE_PROMETHEUS_METRICS_ENDPOINT to true, metrics retrieval is attempted before certain tables exist leading to a crash during app setup. This is mainly an issue in Dev or Test environments where databases are setup from scratch.

This PR changes the metrics initialization to handle this situation and print a helpful error, without crashing the app.

### DIFF
--- a/orchestrator/metrics/subscriptions.py
+++ b/orchestrator/metrics/subscriptions.py
@@ -42,6 +42,8 @@ def _get_subscriptions() -> list[SubscriptionTableQueryResult]:
     ;
     ```
     """
+
+    result: list[SubscriptionTableQueryResult] | None = None
     with handle_missing_tables():
         subscription_count = func.count(SubscriptionTable.subscription_id).label("subscription_count")
         result = (
@@ -62,7 +64,7 @@ def _get_subscriptions() -> list[SubscriptionTableQueryResult]:
             .order_by(desc(subscription_count))
         ).all()
 
-    return result if "result" in locals() else []
+    return result or []
 
 
 class SubscriptionCollector(Collector):


Haak1981:
> Do you see the same behaviour as in #1084? If so, maybe #1143 provides a fix.

Yep seeing the exact same behaviour and indeed #1143 seems to provide a work-around. I still think the pull request adds value seeing the retrieval of metrics shouldn't cause the app to crash.

orchestrator/metrics/processes.py
```diff
@@ -60,37 +61,41 @@ def _get_processes() -> list[ProcessTableQueryResult]:
...
+                ProductTable.name,
+                WorkflowTable.name,
+                SubscriptionTable.customer_id,
+                WorkflowTable.target,
+            )
+            .order_by(desc(process_count))
+        ).all()
+    except ProgrammingError:
+        # Database tables don't exist yet (e.g., during initial migration)
+        return []
```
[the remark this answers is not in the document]
Haak1981: will do, just to be sure use 'psycopg' or 'psycopg2'. In the pyproject.toml I only see "psycopg2-binary==2.9.11"?

orchestrator/metrics/subscriptions.py
```diff
@@ -41,24 +42,27 @@ def _get_subscriptions() -> list[SubscriptionTableQueryResult]:
...
+            .group_by(
+                SubscriptionTable.status,
+                SubscriptionTable.customer_id,
+                SubscriptionTable.insync,
+                ProductTable.name,
+            )
+            .order_by(desc(subscription_count))
+        ).all()
+
+    return result if "result" in locals() else []
```
Mark90: If you define the result variable before the context manager, you can avoid having to check `locals()`, something like: 

```py

result: list[SubscriptionTableQueryResult] | None = None

with handle_missing_tables():
    ....

return result or []
```

You should be able to run mypy locally quite easily with these instructions https://workfloworchestrator.org/orchestrator-core/contributing/development/ to save yourself some waiting time. Feel free to reach out on Discord if that is not the case :)

Haak1981: processed comments

Issue #1347, opened 2026-01-19
[Feature]: Change scheduler API to not return jobs scheduler through decorator


Mark90:
Schedules created using the decorator will not be shown in the UI. As such, we should also not return them in the API (added in #1104) to avoid confusion. Note: it should still be possible to see these schedules through the `show-schedule` CLI command.

PR #1391, opened 2026-02-10
#1347 Show all tasks from CLI, but remove them from resolver

Igoranze:
part of https://github.com/workfloworchestrator/orchestrator-core/issues/1347

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -190,6 +190,7 @@ nav:
                 - v4.x: guides/upgrading/4.0.md
                 - v4.7: guides/upgrading/4.7.md
                 - v4.8: guides/upgrading/4.8.md
+                # - v5.0: guides/upgrading/5.0.md  # TODO: uncomment for final 5.0 release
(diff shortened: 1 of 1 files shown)


orchestrator/cli/scheduler.py
```diff
@@ -172,3 +174,32 @@ def load_initial_schedule() -> None:
 
         typer.echo(f"Initial Schedule: {schedule}")
         add_unique_scheduled_task_to_queue(APSchedulerJobCreate(**schedule))  # type: ignore
+
+
+@app.command()
+def list_all_ap_scheduled_tasks() -> None:
+    """Returns a list of all scheduled tasks in the APScheduler instance.
+
+    This also includes tasks that are scheduled through the decorator and not through the API, which can be useful for debugging.
```
Mark90: Is this any different from `show_schedule`?

Igoranze: No, removed this one in favour of the `show_schedule`

PR #1404, opened 2026-02-13
Change secret env variables to use pydantic secret type

### LINKED ISSUE #1007: [Feature]: Change secret env variables to use pydantic `Secret` generic typing
### Contact Details

_No response_

### What should we build?

pydantic `Secret` generic type masks the property value immediately and makes the value only available with `.get_secret_value()`.

This would make the masking specifically for exposing the env variables unnecessary since they are masked from the start and makes less likely for something to accidentally not be masked unless not typed with `Secret[{type}]`.

https://docs.pydantic.dev/2.8/examples/secrets/#serialize-secretstr-and-secretbytes-as-plain-text

jbgann:
In this PR, the `AppSettings` variables `CACHE_URI`, `DATABASE_URI`, and `WEBSOCKET_BROADCASTER_URL` are updated to be Pydantic Secret subtypes, so that masking of these sensitive values automatically occurs any place they are not dereferenced with  `.get_secret_value()`.

Accordingly, ad-hoc masking is removed, and the necessity of accessing the value of these properly in dependent code is documented in the upgrade guide.

Closes #1007

### DIFF
--- a/docs/guides/upgrading/5.0.md
+++ b/docs/guides/upgrading/5.0.md
@@ -4,7 +4,7 @@ In this document we describe important changes to be aware of when upgrading fro
 
 If you are not yet using v4.x, please consult the 4.0 upgrade guide first.
 
-If you are using scheduled tasks and/or workflow authorization callbacks, please consult the 4.7 and 4.8 upgrades first as well. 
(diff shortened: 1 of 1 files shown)


Mark90 (approved):
Looks good, good docs!

PR #1479, opened 2026-03-18
Removed old OrchSecretStr

Igoranze:
Removed old OrchSecretStr. In favor of the Pydantic SecretStr

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -24,11 +24,16 @@ pre-commit run --all-files             # format, lint, type check.
 - Formatter: black + ruff
 
 ## Code Rules
+- Code shoule be as `pure` as possible with no side effects.
 - Minimize loops keep it as functional as possible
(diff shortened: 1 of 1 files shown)


orchestrator/settings.py
```diff
@@ -73,7 +73,7 @@ class AppSettings(BaseSettings):
     MAIL_PORT: int = 25
     MAIL_STARTTLS: bool = False
     CACHE_URI: SecretRedisDsn = "redis://localhost:6379/0"  # type: ignore
-    CACHE_HMAC_SECRET: OrchSecretStr | None = None  # HMAC signing key, used when pickling results in the cache
+    CACHE_HMAC_SECRET: SecretStr | None = None  # HMAC signing key, used when pickling results in the cache
```
Mark90: Think we can remove this? I don't see any usages of it @pboers1988

PR #1544, opened 2026-04-13
Deprecate TSV search, make LLM search the default

### LINKED ISSUE #1264: [Feature]: Deprecate the TSV search
### What should we build?

Deprecate the TSV and make the LLM enabled search the default.

- [ ] Implement the new UI
- [ ] Add deprecated warnings in the UI and backend
- [ ] Add the LLM migrations to the alembic tree
- [ ] Make indexing of all tables automatic when they are updated
- [ ] Handle a "no embedding model" situation gracefully and disable the vector embeddings without errors
- [ ]

pboers1988:
## Summary

Closes #1264

- Make `litellm` a core dependency (remove `search` pip extra)
- Add alembic migration for AI search tables (`ai_search_index`, `agent_runs`, `search_queries`, `graph_snapshots`) — previously created at runtime by `llm_migration.py`
- Remove `SEARCH_ENABLED` setting and all feature gates — search is now always enabled
- Add `DeprecationWarning` to TSV search query path (REST + GraphQL)
- Delete `llm_migration.py` (replaced by alembic migration)
- Remove `SEARCH_ENABLED` from tests and CI workflows
- Add section 7 to the 5.0 upgrade guide covering prerequisites (pgvector, extensions), migration, index initialization, and embedding resize

## Test plan

- [x] Verify `uv lock` resolves with litellm in core deps
- [x] Run alembic migration on a fresh database and verify all tables are created
- [x] Run alembic migration on an existing database (with tables from runtime migration) — should be idempotent
- [x] Verify TSV search endpoints emit deprecation warnings
- [x] Run `uv run pytest test/unit_tests -x`
- [x] Run search integration tests without `SEARCH_ENABLED=true` prefix
- [x] Verify CLI commands (`index`, `embedding`, `search`, `speedtest`) are registered without `SEARCH_ENABLED`


docs/guides/upgrading/5.0.md
```diff
@@ -276,3 +276,86 @@ VlanRangesType = strawberry.scalar(
...
+```
+
+This creates the AI search tables. For existing installations that already have these tables (from the runtime migration), the migration uses `CREATE TABLE IF NOT EXISTS` and is safe to run.
+
+**3. Configure your embedding provider**
+
+Set the following environment variables:
+
+```bash
+OPENAI_API_KEY=sk-...          # Or your embedding provider's API key
```
pboers1988: Also add the openai base url, and mention that its optional. Search works without it. Semantic mode will be disabled though

orchestrator/api/helpers.py
```diff
@@ -161,17 +172,25 @@ def _getattr(obj: object, attr: str) -> Any:
...
+        if isinstance(value, dict):
+            yield from _prefixed(key, get_dict_items(value))
+            yield (key, value)
+        elif isinstance(value, list):
+            yield from itertools.chain.from_iterable(
+                _list_item_entries(key, idx, item) for idx, item in enumerate(value)
+            )
+
+    def get_dict_items(d: dict) -> Iterable:
+        return itertools.chain.from_iterable(_dict_value_entries(k, v) for k, v in d.items())
```
kvklink, line 204: I think these could do with some code docs. Both for maintainability, and because I have only a very vague idea of what is going on here.

orchestrator/workflows/tasks/cleanup_tasks_log.py
```diff
@@ -50,23 +49,19 @@ def remove_tasks() -> State:
...
-
-        if len(deleted_process_id_list) > 0:
-            count = (
-                db.session.query(AiSearchIndex)
-                .filter(AiSearchIndex.entity_type == EntityType.PROCESS)
-                .filter(AiSearchIndex.entity_id.in_(deleted_process_id_list))
-                .delete(synchronize_session=False)
-            )
+    from orchestrator.db.models import AiSearchIndex
+    from orchestrator.search.core.types import EntityType
```
kvklink: Any reason not to move these to the top of the file?

orchestrator/llm_settings.py
kvklink, line 1: I think this file can get merged with the settings.py file next to it, since these settings are now not conditional anymore anyway

kvklink (changes requested):

orchestrator/api/helpers.py
pboers1988: This is a little bit Yak shaving. As it already existed, formatting just changed. Ill try and add some docs.

kvklink: Cool, learnt about a new metaphor now :) And I agree, it's pretty trivial, something like a docstring on `get_dict_items` like "Convert a dict into a flat list of keys" would suffice (which is what I think it does?)

kvklink (dismissed):
Leip Goeie Troep Maat

docs/guides/upgrading/5.0.md
```diff
@@ -276,3 +276,96 @@ VlanRangesType = strawberry.scalar(
...
+
+After the migration, populate the search index by running the CLI indexing commands:
+
+```bash
+python main.py orchestrator index subscriptions
+python main.py orchestrator index products
+python main.py orchestrator index processes
+python main.py orchestrator index workflows
+```
+
```
Mark90, line 352: Isn't there a task for this?

orchestrator/settings.py
```diff
@@ -112,6 +126,39 @@ class AppSettings(BaseSettings):
...
+    EMBEDDING_API_BASE: str | None = None
+    EMBEDDING_FALLBACK_MAX_TOKENS: int | None = 512
+    EMBEDDING_MAX_BATCH_SIZE: int | None = None
+
+    # General LiteLLM settings
+    LLM_MAX_RETRIES: int = 3
+    LLM_TIMEOUT: int = 30
+
+    # Toggle creation of extensions
+    LLM_FORCE_EXTENTION_MIGRATION: bool = False
```
Mark90:
```suggestion
    LLM_FORCE_EXTENSION_MIGRATION: bool = False
```

orchestrator/settings.py
```diff
@@ -112,6 +126,39 @@ class AppSettings(BaseSettings):
...
+
+    @field_validator("EMBEDDING_MODEL")
+    def validate_embedding_model_format(cls, v: str) -> str:
+        """Validate that embedding model is in 'vendor/model' format."""
+        if "/" not in v:
+            raise ValueError("EMBEDDING_MODEL must be in format 'vendor/model'")
+        return v
```
Mark90: Use `Field(pattern="...")` here?

Mark90 (approved):

docs/guides/upgrading/5.0.md
pboers1988: It's a one time thing. Indexing of single instance of these entities is handled in workflows or API endpoints.

jbgann (approved):

kvklink (approved):

PR #1549, opened 2026-04-14
Migrate from psycopg2 to psycopg3

estoel81:
## Summary

- Replace `psycopg2-binary` with `psycopg[binary]>=3.3.2` and update all `postgresql://` URIs to `postgresql+psycopg://` across CI, docs, settings, and test fixtures
- Fix idle-in-transaction connections caused by psycopg3's autobegin: add pool `checkin` rollback, `pool_recycle`, and wrap all bare DB reads on Celery workers in `transactional()`
- Make `transactional()` reentrant so nested calls (e.g. a step wrapping a Celery task entry point) don't roll back the outer transaction
- Serialize step state to a plain dict in `_db_log_step` before the outer flush, preventing re-evaluation of expensive `@computed_field` properties and ensuring the next step never receives live `SubscriptionModel` instances
- Add `OrchestratorCore.register_table()` to copy extra `column_property` attributes from a custom table subclass onto the base mapper
- Emit `DeprecationWarning` when `DATABASE_URI` uses the bare `postgresql://` scheme

## Test Plan

- [x] `uv run pytest test/unit_tests/db/test_database.py` — transactional reentrancy
- [x] `uv run pytest test/unit_tests/services/test_processes.py` — state serialization + idle-in-transaction
- [x] `uv run pytest test/unit_tests/test_subscription_table_registry.py` — register_table
- [x] Full unit test suite: `uv run pytest test/unit_tests/`


test/unit_tests/db/test_pool_events.py
```diff
@@ -0,0 +1,108 @@
...
+
+from unittest.mock import MagicMock
+
+import pytest
+from sqlalchemy import create_engine
+
+from orchestrator.db.database import _register_pool_events
+
+
+def _sqlite_engine():
```
pboers1988: Do we need these tests? Seems like we are testing sqlalchemy functionality

test/unit_tests/test_database_uri_validation.py
```diff
@@ -0,0 +1,113 @@
+# Copyright 2019-2025 SURF.
```
pboers1988: Lots of tests not sure we need it.

pboers1988 (approved):
LGTM minor nitpicks

pboers1988 (dismissed):
@estoel81 Also need some 5.0.0 upgrade docs.

docs/guides/upgrading/5.0.md
```diff
@@ -276,3 +278,121 @@ VlanRangesType = strawberry.scalar(
...
+For example, if your subclass defines a computed column:
+
+```python
+from sqlalchemy.orm import column_property, Mapped
+from sqlalchemy import select
+from orchestrator.db import SubscriptionTable
+
+class MySubscriptionTable(SubscriptionTable):
+    customer_name: Mapped[str] = column_property(
+        select(CustomerTable.fullname)
```
tjeerddie, line 385: maybe add a table example above this, the only reference of `CustomerTable` is somewhere in the docs about adding it, could also link to that place

orchestrator/services/processes.py
```diff
@@ -324,6 +325,11 @@ def _db_log_step(
     p = _update_process(stat.process_id, step, process_state)
     current_step = _get_current_step_to_update(stat, p, step, process_state)
 
+    # Serialize state to a plain dict now, before the outer transactional() flushes it.
+    # This ensures the next step never receives live SubscriptionModel instances and
+    # avoids re-evaluating expensive @computed_field properties on flush.
+    current_step.state = json_loads(json_dumps(current_step.state))
```
tjeerddie, line 331: maybe override the `pydanticmodel.model_dump`, so its in done in the model

orchestrator/services/tasks.py
```diff
@@ -69,9 +72,12 @@ def initialise_celery(celery: Celery) -> None:  # noqa: C901
 
     def start_process(process_id: UUID, user: str) -> UUID | None:
         try:
-            process = _get_process(process_id)
-            pstat = load_process(process)
-            ensure_correct_process_status(process_id, ProcessStatus.CREATED)
+            # Celery workers use the empty-scope session (no database_scope); wrap in
+            # transactional() to prevent psycopg3 autobegin leaving the connection idle-in-transaction.
+            with transactional(db, local_logger):
```
tjeerddie, line 77: could move the transactional inside the `_get_process`, `load_process` and `ensure_correct_process_status`. so any re-uses in other places don't need to be wrapped inside a transactional.

orchestrator/app.py
```diff
@@ -149,6 +162,21 @@ def __init__(
 
         self.include_router(api_router, prefix="/api")
 
+        # Validate DATABASE_URI dialect before initializing the database.
```
tjeerddie, line 165: I don't believe this comment is necessary

orchestrator/workflow.py
```diff
@@ -1531,13 +1531,14 @@ def _exec_steps(steps: StepList, starting_process: Process, dblogstep: StepLogFu
...
         # Execute step
         try:
-            engine_status = get_engine_settings_table()
-            if engine_status.global_lock:
-                # Exiting from thread workflow engine is Paused or Pausing
-                consolelogger.info(
-                    "Not executing Step as the workflow engine is Paused. Process will remain in state 'running'"
-                )
-                return process
+            with transactional(db, logger):
```
tjeerddie, line 1534: move the `transactional` inside the `get_engine_settings_table`?

orchestrator/workflow.py
```diff
@@ -1549,8 +1550,11 @@ def _exec_steps(steps: StepList, starting_process: Process, dblogstep: StepLogFu
         # Convert ErrorState to ErrorDict when Failed or Waiting before writing to the database
         # as bare exceptions are not JSON serializable
         result_to_log = step_result_process.on_failed(error_state_to_dict).on_waiting(error_state_to_dict)
-        result_to_log.on_success(mutationlogger).on_failed(errorlogger).on_waiting(errorlogger)
-        process = dblogstep(step, result_to_log)
+        # Wrap logging in transactional() so SELECTs triggered by mutationlogger and
+        # dblogstep run inside a managed transaction (psycopg3 autobegin prevention).
+        with transactional(db, logger):
+            result_to_log.on_success(mutationlogger).on_failed(errorlogger).on_waiting(errorlogger)
```
tjeerddie, line 1556: I don't think mutationlogger triggers any selects and move the transactional inside the dblogstep

test/unit_tests/db/test_database.py
```diff
@@ -135,3 +135,31 @@ def test_transactional_disables_commit_inside_block() -> None:
...
     assert db.session.info.get("disabled") is False
+
+
+def test_transactional_nested_does_not_commit_or_rollback() -> None:
+    """Nested transactional() must not commit or rollback; the outer call owns the transaction."""
+    db = _make_db(disabled=True)  # simulate already inside an outer transactional()
+    log = MagicMock()
+
+    with transactional(db, log):
+        pass
```
tjeerddie, line 146: wouldn't you need a db insert or db error before you can check that a commit or rollback isn't called.

test/unit_tests/metrics/test_dbutils.py
```diff
@@ -0,0 +1,53 @@
...
+    """No exception inside the block: context manager exits cleanly."""
+    result = []
+    with handle_missing_tables():
+        result.append("executed")
+    assert result == ["executed"]
+
+
+def test_handle_missing_tables_suppresses_undefined_table_error() -> None:
+    """ProgrammingError wrapping UndefinedTable is caught and not re-raised."""
+    orig = psycopg_errors.UndefinedTable.__new__(psycopg_errors.UndefinedTable)
```
tjeerddie, line 31: this looks weird, cant you directly make the class?
```suggestion
    orig = psycopg_errors.UndefinedTable()
```

test/unit_tests/test_subscription_table_registry.py
```diff
@@ -0,0 +1,113 @@
...
+    assert "extra_field" in base_mapper.column_attrs
+
+
+def test_register_table_does_not_overwrite_existing_columns():
+    """register_table should not overwrite columns already on the base class."""
+    base_mapper = sa_inspect(SubscriptionTable)
+    original_description = base_mapper.column_attrs["description"]
+
+    class CustomSubscriptionTable(SubscriptionTable):
+        pass
```
tjeerddie, line 52: you aren't overwriting the description here?

test/unit_tests/test_subscription_table_registry.py
```diff
@@ -0,0 +1,113 @@
...
+
+    OrchestratorCore.register_table(SubscriptionTable, CustomSubscriptionTable)
+
+    result = db.session.scalars(select(SubscriptionTable)).first()
+    assert result is not None
+    assert result.extra_field is not None
+
+
+@pytest.mark.usefixtures("_cleanup_extra_field")
+def test_register_table_columns_visible_in_inspect():
```
tjeerddie, line 104: this seems redundant since `test_register_table_copies_column_properties` already checks if the `extra_fields` is inside `mapper.column_attrs`

test/unit_tests/services/test_processes.py
```diff
@@ -135,6 +151,84 @@ def test_process_log_db_step_success(simple_workflow):
...
+        last_status=ProcessStatus.CREATED,
+        created_by=SYSTEM_USER,
+    )
+    db.session.add(p)
+    db.session.commit()
+
+    # Build a minimal in-memory workflow function. The workflow is never executed
+    # because we mock _run_process_async; we only need a non-removed_workflow value
+    # for the early branch in thread_start_process and for ProcessStat construction.
+    wf = make_workflow(_workflow_test_fn, "wf description", None, Target.SYSTEM, StepList())
```
tjeerddie, line 209: can't you use the simple_workflow?

test/unit_tests/services/test_processes.py
```diff
@@ -135,6 +151,84 @@ def test_process_log_db_step_success(simple_workflow):
...
+        state=Success({}),
+        log=StepList(),
+        current_user="user",
+    )
+
+    assert not db.session.in_transaction(), "precondition: clean session"
+
+    # Mock _run_process_async to skip actually running the workflow thread; we are
+    # only testing the pre-runwf section of thread_start_process.
+    with mock.patch("orchestrator.services.executors.threadpool._run_process_async"):
```
tjeerddie, line 223: could add the patch as decorator on the test function

test/unit_tests/services/test_tasks.py
```diff
@@ -1,10 +1,296 @@
...
+        get_celery_task("tasks.new_task")
+
+
+# ---------------------------------------------------------------------------
+# register_custom_serializer
+# ---------------------------------------------------------------------------
+
+
+def test_register_custom_serializer_registers_orchestrator_json():
+    with patch("orchestrator.services.tasks.registry") as mock_registry:
```
tjeerddie, line 109: use patch as decorator, also update this in the other tests

test/unit_tests/services/test_tasks.py
```diff
@@ -1,10 +1,296 @@
...
+
+    assert celery.conf.task_routes == {
+        NEW_TASK: {"queue": "new_tasks"},
+        NEW_WORKFLOW: {"queue": "new_workflows"},
+        RESUME_TASK: {"queue": "resume_tasks"},
+        RESUME_WORKFLOW: {"queue": "resume_workflows"},
+    }
+
+
+def test_initialise_celery_registers_four_named_tasks():
```
tjeerddie, line 141: redundant, you already check it in the previous test

test/unit_tests/services/test_tasks.py
```diff
@@ -1,10 +1,296 @@
...
+        result = celery_start_fn(process_id, "user")
+
+    assert result == process_id
+
+
+@pytest.mark.parametrize("failing_fn", ["_get_process", "load_process", "thread_start_process"])
+def test_start_process_returns_none_on_exception(celery_start_fn, failing_fn):
+    process_id = uuid4()
+
+    patches = {
```
tjeerddie, line 201: probably better to use variables instead of a dict or change the keys to just the function name.

test/unit_tests/services/test_tasks.py
```diff
@@ -1,10 +1,296 @@
...
+    ):
+        target = {"_get_process": m_get, "thread_resume_process": m_thread}[failing_fn]
+        target.side_effect = RuntimeError("boom")
+        result = celery_resume_fn(process_id, "user")
+
+    assert result is None
+
+
+# ---------------------------------------------------------------------------
+# CeleryJobWorkerStatus (existing fixture kept below)
```
tjeerddie, line 292: the `(existing fixture kept below)` is useless in the comment, could have more context about the tests below instead.

pboers1988 (dismissed):

estoel81:
### Code review

Found 3 issues:

1. **Shebang placed after copyright block — will not be recognized by the OS**

The PR prepends a new copyright header to `orchestrator/app.py`, moving the shebang from line 1 to line 13. A shebang is only recognized when it is the very first line of a file. The file also ends up with two copyright blocks (lines 1–12 and lines 22–33) with mismatched years (`2019-2026` vs `2019-2025`). The fix is to update the existing copyright block in place rather than prepending a new one.

https://github.com/workfloworchestrator/orchestrator-core/blob/f94abe88e51d6f714a8ccd69384d4cc316dcb6b1/orchestrator/app.py#L11-L16

2. **`execute_step` runs outside both `transactional()` guards — idle-in-transaction risk**

In `_exec_steps`, the first `transactional()` covers only the global-lock check (lines 1534–1542) and the second covers only logging (lines 1555–1559). The actual step body at line 1544 executes unguarded. With psycopg3's autobegin, any SELECT issued inside the step opens a transaction on the connection. If the step then raises and the logging `transactional()` also fails (e.g. DB unavailable), the connection is returned to the pool with an open transaction — the exact problem the PR set out to fix.

https://github.com/workfloworchestrator/orchestrator-core/blob/f94abe88e51d6f714a8ccd69384d4cc316dcb6b1/orchestrator/workflow.py#L1542-L1548

3. **`WrappedSession.commit()` warning fires on every step execution**

`_db_log_step` calls `db.session.commit()` directly (processes.py line 334). When called from within the logging `transactional()` block, `disabled` is `True`, so `WrappedSession.commit()` logs `"Step function tried to issue a commit. It should not!"` — on every single step. This warning was intended to catch rogue commits inside user-written step functions; it now fires unconditionally as misleading noise. The direct `db.session.commit()` call in `_db_log_step` should be removed and the outer `transactional()` relied upon to commit, or the guard in `WrappedSession` should distinguish framework commits from step-function commits.

https://github.com/workfloworchestrator/orchestrator-core/blob/f94abe88e51d6f714a8ccd69384d4cc316dcb6b1/orchestrator/db/database.py#L155-L163

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

orchestrator/services/processes.py
estoel81: The state dict is not a Pydantic model — it's a plain dict[str, Any] that happens to contain SubscriptionModel instances. Overriding model_dump on SubscriptionModel wouldn't help here.

orchestrator/services/tasks.py
estoel81: The three calls are intentionally grouped in one transaction. After a commit, SQLAlchemy expires ORM attributes (expire_on_commit=True), so splitting into per-function transactions would cause load_process to lazy-load process.steps and process.workflow outside any transaction — the exact idle-in-transaction problem we're fixing. Keeping them in one block avoids that.

orchestrator/workflow.py
estoel81: Not changed — the current transactional wrapper is fine. Can be a follow-up if needed.

orchestrator/workflow.py
estoel81: You're right — mutationlogger has no DB reads. The outer transactional() around it is redundant. We removed the dead db.session.commit() from _db_log_step (the WrappedSession warning fix), but left moving the transactional inside _db_log_step as a follow-up refactor for now.

test/unit_tests/services/test_processes.py
estoel81: No, simple_workflow is expired after db.session.commit() due to SQLAlchemy's expire-on-commit behavior. Accessing any attribute on it would trigger a lazy load SELECT, which with psycopg3 autobegin opens a transaction — breaking the precondition that the session must be transaction-free before calling thread_start_process. make_workflow creates a detached in-memory object with no session interaction, keeping the precondition clean.

docs/guides/upgrading/5.0.md
```diff
@@ -99,13 +108,14 @@ app_settings.DATABASE_URI = SecretPostgresDsn(db_uri)
 ```
 
 !!! warning
-    Previously custom settings whose names contained the string `secret` or `password` would automatically be masked.
-    This is no longer the case, and you should now use the type `SecretStr` for these values instead.
+Previously custom settings whose names contained the string `secret` or `password` would automatically be masked.
+This is no longer the case, and you should now use the type `SecretStr` for these values instead.
```
Mark90: Please don't let claude rewrite other things in the docs. In this particular case the representation changes because of the removed indentation

docs/guides/upgrading/5.0.md
```diff
@@ -452,3 +474,135 @@ def auth_callback(context: AuthContext | None) -> bool:
...
+    result = db.session.execute(...)
+```
+
+### 11. New `OrchestratorCore.register_table()` API
+
+#### What changed
+
+A new static method `OrchestratorCore.register_table()` has been added. It allows you to register custom
+`column_property` attributes from a `SubscriptionTable` subclass onto the base `SubscriptionTable` mapper. This makes
+custom computed columns available in queries and GraphQL without modifying the core table definition.
```
Mark90: SubscriptionTable is just one example, it applies to all built-in ORM classes, right?

orchestrator/services/tasks.py
```diff
@@ -69,9 +72,12 @@ def initialise_celery(celery: Celery) -> None:  # noqa: C901
...
+                process = _get_process(process_id)
+                pstat = load_process(process)
+                ensure_correct_process_status(process_id, ProcessStatus.CREATED)
             thread_start_process(pstat, user=user, broadcast_func=process_broadcast_fn)
```
Mark90, line 81: Shouldn't this be inside the transactional?

orchestrator/utils/json.py
```diff
@@ -130,6 +131,8 @@ def to_serializable(o: Any) -> Any:
         TypeError: in case no conversion was possible.
 
     """
+    if isinstance(o, Decimal):
+        return float(o)
```
Mark90: Good to add this but I would convert the decimal to a string. Floats are treacherous

```
x = Decimal("0.1")
y = float(x)
print(y)
print(Decimal.from_float(y))
```

outputs

```
0.1
0.1000000000000000055511151231257827021181583404541015625
```

test/unit_tests/test_subscription_table_registry.py
```diff
@@ -0,0 +1,100 @@
...
+    """register_table should only copy column_properties, not relationships."""
+    base_mapper = sa_inspect(SubscriptionTable)
+    original_relationships = set(base_mapper.relationships.keys())
+
+    class CustomSubscriptionTable(SubscriptionTable):
+        pass
+
+    OrchestratorCore.register_table(SubscriptionTable, CustomSubscriptionTable)
+
+    assert set(base_mapper.relationships.keys()) == original_relationships
```
Mark90: The CustomSubscriptionTable does not specify any new relationships that shouldn't be copied, so this doesn't seem to test anything?

Mark90 (changes requested):

docs/guides/upgrading/5.0.md
estoel81: Yes, this is generic. Will adjust the docs accordingly

docs/guides/upgrading/5.0.md
estoel81: This is not claude, but my code formatter after save. Will revert this change

orchestrator/services/tasks.py
estoel81: thread_start_process manages its own transactions. If you add this into this transactional it will conflict with _set_process_status_running's commit() and rollback() closing this transactional

orchestrator/utils/json.py
estoel81: Will change this

test/unit_tests/test_subscription_table_registry.py
estoel81: Will remove this

test/unit_tests/test_subscription_table_registry.py
tjeerddie: this doesn't appear to test what the name of the test describes `test_register_table_does_not_overwrite_existing_columns`

Mark90 (approved):
🎉

tjeerddie (approved):

PR #1573, opened 2026-04-15
1552 enhancement fix rest authorization handler and improve docs

### LINKED ISSUE #1552: [Bug]: Fix REST authorization handler and improve docs
Orchestrator-core currently is inconsistent in how it handles authorization callbacks for REST and GraphQL paths. There is also no relevant documentation on how users should use these.

https://github.com/workfloworchestrator/oauth2-lib/issues/75 will change corresponding docs in oauth2_lib.

NB: I mean API authorization callbacks, not workflow/step callbacks.

In particular:

* WFO's GraphQL API authorizations successfully return 403 on failed authorization, because oauth2_lib's strawberry.py raises this HTTP exception.
* WFO's REST API authorizations do not return 403 on failed authorization, unless an exception is raised inside the authorizer callback.

This change should:

* Ensure the API provided by oauth2_lib is respected, and raise in the REST API authorization handler
* Update the Auth(n|z) documentation 

The gist of the documentation required:

API authorization for both REST and GraphQL is handled by registering callbacks to be called by oauth2_lib's AuthManager. The callback should take arguments X, Y, Z, and then:

* Return `None` if authentication should be bypassed (i.e. the callback is running in your local development environment.)
* Return `True` for an authorized user.
* Raise a 403 HTTP exception (with custom detail) for an unauthorized user, if desired
* Return `False` will cause WFO to raise a default exception.

eenblam:
## Summary
Ensures we raise a 403 Forbidden whenever an API authorization callback returns False instead of raising its own exception.

Updates docs to clarify usage of API authorization callbacks.

## Related Issues
Fixes #1552 

## Checklist
    - N/A


slbyrnes1 (approved):

orchestrator/security.py
```diff
@@ -45,7 +47,12 @@ async def authenticate(
 
 
 async def authorize(request: Request, user: Annotated[OIDCUserModel | None, Depends(authenticate)]) -> bool | None:
-    return await request.app.auth_manager.authorization.authorize(request, user)
+    result = await request.app.auth_manager.authorization.authorize(request, user)
+    if result is False:  # None is different!
```
Mark90, line 64: Todo: would be good to mention this behavior change in the 5.0 upgrade guide

Mark90 (approved):

PR #1588, opened 2026-04-22
[Feature]: Convert orchestrator-core into a namespace package

### LINKED ISSUE #1471: [Feature]: Support namespace packaging
### Contact Details

_No response_

### What should we build?

Convert the package definition of the `orchestrator` module into something that supports [namespace packages](https://peps.python.org/pep-0420/)

> Namespace packages are a mechanism for splitting a single Python package across multiple directories on disk.

This will enable the development and packaging of extra modules outside of `orchestrator-core` like [`orchestrator-optical`](https://github.com/workfloworchestrator/orchestrator-optical).
With it, we can have:

1. `uv add orchestrator-core`
2. `uv add orchestrator-optical`
3. `from orchestrator.optical.products import OpticalProduct`
4. Profit

## Implementation
For the implementation of this feature, relevant docs [in this section of PEP 420](https://peps.python.org/pep-0420/#packaging-implications). It comes down to removing the `__init__.py` from the `orchestrator` main module. Flit should readily support implicit namespace packages [like here](https://github.com/pypa/flit/issues/370).

### Relevant pseudo code

```shell

```

kvklink:
## Content of this PR
Converts the `orchestrator-core` module from `orchestrator.stuff` into a namespace package in `orchestrator.core.stuff`. This includes a major refactor of all code and is definitely a breaking change for every wfo user. A migration script is included that updates all import statements for users which should get them 99% of the way there.

Build backend `flit` has been replaced with `uv-build`, which allows us to remove some now redundant files that overlap with `pyproject.toml`. Most config has been moved into there now which cleans up the root folder of the repo. A nice bonus.

Merging this PR will pave the way for other namespace packages in the wfo universe like publishing `orchestrator-optical` to use the module `orchestrator.optical`, etc.

### TL;DR
- Bump version number to `5.0.0rc2`.
- Move from `orchestrator.*` to `orchestrator.core.*`.
- Add migration script for updating import statements in user implementations.
- Move from `flit` to `uv-build`.
- Update `ruff` to the latest version.
- Remove `black` from the toolchain as it's eclipsed in functionality by `ruff`.
- Clean up `.gitignore`.
- Merge `setup.py`, `setup.cfg`, and `nitpick-style.toml` into `pyptoject.toml`.
- Remove `.stignore`.

## Related Issues
Fixes #1471

## Checklist


docs/guides/upgrading/5.0.md
```diff
@@ -33,8 +33,8 @@ decorator.
 
 ### 2. Secret settings for sensitive URIs
```
acidjunk, line 72: you mentioned a script that will help users to migrate. Should be in the migration manual?

kvklink: You're totally right, I forgot to mention it in the documentation. I will write something about it there.

orchestrator/core/devtools/scripts/migrate_50.py
```diff
@@ -0,0 +1,55 @@
...
+    """Replace all orchestrator module imports."""
+    text = f.read_text()
+    text_orig = text
+    # First, we replace all "from orchestrator." with "from orchestrator.core."
+    rgx = r"from orchestrator\.(.+)"
+    text = re.sub(rgx, r"from orchestrator.core.\1", text)
+
+    # Then, replace all "import orchestrator." with "import orchestrator.core."
+    rgx = r"import orchestrator\.(.+)"
+    text = re.sub(rgx, r"import orchestrator.core.\1", text)
```
Mark90: Would be nice to have the regex exclude `orchestrator.core` to make it idempotent, should be easy with a negative lookahead :)

docs/guides/upgrading/5.0.md
```diff
@@ -18,10 +18,11 @@ In this release, the following breaking changes are introduced:
 5. The `description` parameter in workflow decorators is now deprecated
 6. `strawberry-graphql` upgrade includes breaking changes
 7. LLM-powered search is now the default
-8. `orchestrator.security.authorize` and `authorize_websocket` now raise on `False`
+8. `orchestrator.core.security.authorize` and `authorize_websocket` now raise on `False`
```
Mark90: Since the namespacing is upgrade step 12 and users will most likely go through the steps in order, all the imports before step 12 should probably not have the `.core.`  infix :)

docs/architecture/application/workflow.md
```diff
@@ -50,7 +50,7 @@ There are two high-level kinds of workflows:
     - Not tied to a specific product and may not involve a subscription at all.
```
Mark90, line 50: We should document somewhere (not specifically in this file, but somewhere) that orchestrator-core adhers to the [namespace package ](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) pattern by ommitting the top-level `__init__.py`, and that "extension packages" should omit this as well (if they don't, it'll break all core imports so it is quite noticeable at least)

On a related note, we might want to provide some kind of "orchestrator extension package" template for people to refer to if they want to create/validate one (can be a follow-up ticket)

docs/guides/upgrading/5.0.md
```diff
@@ -637,3 +636,26 @@ If you are using Celery workers, register the table in your worker startup as we
...
 OrchestratorCore.register_table(SubscriptionTable, MySubscriptionTable)
 ```
+
+### 12. Namespace packaging
+
+#### What changed
+
+The whole `orchestrator` Python module has moved to a namespace package structure, which means that every submodule under
+`orchestrator.*` has been moved to `orchestrator.core.*`. This will allow for the development of other, optional
+modules that can be used together with the Orchestrator core.
```
Mark90: Link to documentation of extension modules (mentioned in earlier comment)

docs/guides/upgrading/5.0.md
```diff
@@ -637,3 +636,26 @@ If you are using Celery workers, register the table in your worker startup as we
...
+
+#### What you need to do
+
+All import statements that mention the Orchestrator core should be updated to reflect this change. There is a migration
+script available to help with this, which will try to do this automatically. It can be run as follows:
+`uv run orchestrator.core.devtools.scripts.migrate_50 <dir>` where `<dir>` points to your orchestrator implementation.
```
Mark90:
```suggestion
`uv run orchestrator.core.devtools.scripts.migrate_50 <dir>` where `<dir>` points to your orchestrator implementation, unit tests or any other directory with python files that import something from `orchestrator`.
```

(feel free to rephrase)

orchestrator/core/db/filters/product.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2019-2026 SURF, GÉANT.
```
Mark90, line 1: Interesting that the github diff sometimes gives up on detecting that files have been moved

Image

orchestrator/core/cli/database.py
```diff
@@ -21,23 +21,22 @@
...
+from orchestrator.core.cli.migration_helpers import create_migration_file
+from orchestrator.core.db import init_database
+from orchestrator.core.settings import app_settings
 
 logger = get_logger(__name__)
 
 app: typer.Typer = typer.Typer()
 
-orchestrator_module_location = os.path.dirname(orchestrator.__file__)
+orchestrator_module_location = os.path.join(os.path.dirname(__file__), os.pardir)
```
Mark90, line 39: This seems more resilient in case we ever move this file?


```suggestion
orchestrator_module_location = os.path.dirname(orchestrator.core.__file__)
```

orchestrator/core/cli/migration_helpers.py
```diff
@@ -8,11 +21,9 @@
 from alembic.util.exc import CommandError
 from structlog import get_logger
 
-import orchestrator
-
 logger = get_logger(__name__)
 
-orchestrator_module_location = os.path.dirname(orchestrator.__file__)
+orchestrator_module_location = os.path.join(os.path.dirname(__file__), os.pardir)
```
Mark90, line 26: See comment on similar line in other file

docs/getting-started/base.md
```diff
@@ -59,10 +59,10 @@ This provides the CLI entrypoint to your Orchestrator.
...
-from orchestrator.cli.main import app as core_cli
-from orchestrator.db import init_database
-from orchestrator.log_config import LOGGER_OVERRIDES
+from orchestrator.core import app_settings
+from orchestrator.core.cli.main import app as core_cli
+from orchestrator.core.db import init_database
+from orchestrator.core.log_config import LOGGER_OVERRIDES
 
 def init_cli_app() -> typer.Typer:
     initialise_logging(LOGGER_OVERRIDES)
```
Mark90, line 68: Since during the recent partner meeting / code sprint we decided that versioned documentation is not a necessity, I believe we **should** make all documentation code snippets have versioned tabs. 

One for "4.x" that contains the current code snippet with `orchestrator` imports, and one for "5.x" with the `orchestrator.core` imports. See  `### Task code` in docs/guides/tasks.md for an example on tabs.

This may be a bit of a hassle to add initially, but it makes it easier to keep the documentation's code snippets correct in the future because then we can either just update the 5.x tab or add a new one if needed. Without this I think users will get lost

orchestrator/core/__init__.py
```diff
@@ -13,9 +13,9 @@
 
 """This is the orchestrator workflow engine."""
 
-# ruff: noqa: I001  (import order matters — app_settings must resolve before app.py)
+from importlib.metadata import version
 
-__version__ = "5.0.0rc1"
+__version__: str = version("orchestrator-core")
```
Mark90, line 18: Neat

.bumpversion.cfg
```diff
@@ -1,12 +1,16 @@
...
-[bumpversion:file:orchestrator/__init__.py]
-search = __version__ = "{current_version}"
-replace = __version__ = "{new_version}"
+[bumpversion:file:pyproject.toml]
+search = version = "{current_version}"
+replace = version = "{new_version}"
+
+[bumpversion:file:uv.lock]
+search = version = "{current_version}"
+replace = version = "{new_version}"
```
Mark90: Why bump uv.lock?

pyproject.toml
```diff
@@ -247,3 +222,121 @@ index-url = "https://pypi.org/simple"
...
+    "dist",
+    "*.egg",
+    "__pycache__",
+    ".cache",
+    ".*",
+    "migrations",
+    "docs",
+    "test/unit_tests/cli/data/generate",
+]
+testpaths = ["test/unit_tests"]
```
Mark90: This should include orchestrator/core to run doctests

Mark90 (changes requested):
Nice change, lots of work done and also some very useful TLC for the project ❤️ 

Went through all changes but the test files and left several remarks.

orchestrator/core/devtools/scripts/migrate_50.py
kvklink: I can probably remove the tailing period outright since with this version `import orchestrator` wouldn't be updated either

docs/guides/upgrading/5.0.md
kvklink: Either that or step 12 moves up the chain, whatever we prefer

docs/architecture/application/workflow.md
kvklink: I can draft a repository template, I like that idea very much!

.bumpversion.cfg
kvklink: To prevent [this](https://gitlab.software.geant.org/goat/gap/geant-service-orchestrator/-/pipelines/106404) from happening 😋

pyproject.toml
kvklink: Ah.. That will be a tricky one. Since including that directory caused duplicate imports when pytest collects all test classes and sqlalchemy will complain about duplicate tables. Pytest would import both `orchestrator.core.db.models` and `core.db.models`

If you have any tips... ❤️

.bumpversion.cfg
Mark90: image


🙄

kvklink: Beautiful timing 🥲

pyproject.toml
Mark90: My virtual friend has 2 suggestions

---

### Option 1: `--import-mode=importlib` (cleanest)

Add to `pyproject.toml`:

```toml
[tool.pytest]
addopts = [
    "--import-mode=importlib",
    "--doctest-modules",
    ...
]
testpaths = ["orchestrator/core", "test/unit_tests"]
```

With `importlib` mode, pytest uses `importlib.util.spec_from_file_location` and
computes the module name from the file's path relative to the first `sys.path`
entry that contains it. Since the repo root is on `sys.path` (via the editable
install), `orchestrator/core/db/models.py` resolves to `orchestrator.core.db.models`
— the same name as the editable install — so it's the same module object, not a
duplicate.

**Risk:** `importlib` mode slightly changes conftest scoping rules. Worth running
the full suite once to confirm nothing breaks.

---

### Option 2: Separate pytest invocation for doctests

Keep `testpaths = ["test/unit_tests"]` as-is and add a second invocation in the
Makefile / CI:

```bash
uv run pytest --doctest-modules orchestrator/core   # doctests in source
uv run pytest test/unit_tests                        # unit tests
```

**Con:** Two commands instead of one. The `pre-commit` / CI setup needs updating.
Also note that `--doctest-modules` without `--import-mode=importlib` has the same
root problem, but since the two invocations run in separate processes the module
registry is reset between them, so the duplicate registration never occurs.

.bumpversion.cfg
Mark90: Hehe. Anyway, the lockfile of the orchestrator-core has no impact on the lockfile of your own orchestrator. I don't see a reason to bump the uv.lock file

kvklink: Ah! But that is not quite what I meant. What this should prevent is that when people start working on the orchestrator right after a version bump, their commit pre-hooks would complain about a lockfile being out of sync. ~Although I don't recall if we have a hook that checks for this. If this is not the case, I will indeed remove this bump config because it won't make a difference.~

We don't even have a check for this, so I will remove the config. It doesn't make a difference in that case anyway

orchestrator/core/devtools/scripts/migrate_50.py
kvklink: Wait, I misread this comment. What do you mean by _exclude_ `orchestrator.core`?

orchestrator/core/cli/database.py
kvklink:
```
$ python orchestrator/core/cli/main.py db history
2026-04-24 12:52:52 [info     ] Starting the orchestrator      version=5.0.0rc2
Traceback (most recent call last):
  File "/Users/karel/git/orchestrator-core/orchestrator/core/cli/main.py", line 16, in <module>
    from orchestrator.core.cli import (
    ...<3 lines>...
    )
  File "/Users/karel/git/orchestrator-core/orchestrator/core/__init__.py", line 29, in <module>
    from orchestrator.core.app import OrchestratorCore
  File "/Users/karel/git/orchestrator-core/orchestrator/core/app.py", line 44, in <module>
    from orchestrator.core.cli.main import app as cli_app
  File "/Users/karel/git/orchestrator-core/orchestrator/core/cli/main.py", line 16, in <module>
    from orchestrator.core.cli import (
    ...<3 lines>...
    )
  File "/Users/karel/git/orchestrator-core/orchestrator/core/cli/database.py", line 40, in <module>
    orchestrator_module_location = orchestrator.core.__file__
                                   ^^^^^^^^^^^^^^^^^
AttributeError: cannot access submodule 'core' of module 'orchestrator' (most likely due to a circular import) 

```

Wish I could, this was the least hacky implementation I was able to get to work ☹️

docs/guides/upgrading/5.0.md
Mark90: Yeah moving it up to step 1 is fine as well

orchestrator/core/devtools/scripts/migrate_50.py
Mark90: So as to not change `orchestrator.core` imports to `orchestrator.core.core`

docs/architecture/application/workflow.md
kvklink: The template repo is now referenced in the docs, and we have a private draft to be built after we have some more internal discussion on what should be in there. Resolving this thread for now

pyproject.toml
kvklink: It appears that [namespace packages](https://docs.pytest.org/en/stable/explanation/goodpractices.html#tests-as-part-of-application-code) are [a bit of](https://stackoverflow.com/questions/50174130/how-do-i-pytest-a-project-using-pep-420-namespace-packages) a [second class citizen](https://github.com/pytest-dev/pytest/issues/12287) in pytest.

Lucky for us, [there is a relatively new configuration option available](https://docs.pytest.org/en/stable/reference/reference.html#confval-consider_namespace_packages) that resolves the problem :)

Mark90: Don't want this to block the PR so fine to ignore for now, maybe we'll just add a `# TODO figure out how to add doctests` in the pyproject.toml for now

Mark90: > It appears that [namespace packages](https://docs.pytest.org/en/stable/explanation/goodpractices.html#tests-as-part-of-application-code) are [a bit of](https://stackoverflow.com/questions/50174130/how-do-i-pytest-a-project-using-pep-420-namespace-packages) a [second class citizen](https://github.com/pytest-dev/pytest/issues/12287) in pytest.
> 
> Lucky for us, [there is a relatively new configuration option available](https://docs.pytest.org/en/stable/reference/reference.html#confval-consider_namespace_packages) that resolves the problem :)

Oh, nice find!

docs/guides/scaling.md
```diff
@@ -133,101 +133,201 @@ When using the Celery executor, you'll need to do this again for the worker inst
...
+        worker_prefetch_multiplier=1,
+        worker_send_task_event=True,
+        task_send_sent_event=True,
+    )
+    ```
+
+=== "`orchestrator-core` < 5.0"
+
+      ```python
+    """This module contains functions and classes necessary for celery worker processes.
```
Mark90, line 228: Hmm, formatting of this is wonky for some reason 

Image

.github/copyright.yaml
```diff
@@ -0,0 +1,8 @@
+# This file contains the definition of what a copyright header should contain.
+# The $CURRENT_YEAR can't be added here programatically, so it should be updated manually eacht year.
```
Mark90:
```suggestion
# The $CURRENT_YEAR can't be added here programatically, so it should be updated manually each year.
```

docs/guides/scaling.md
kvklink: Yep, got the indentation wrong there. Will fix.

docs/guides/tasks.md
```diff
@@ -377,8 +403,8 @@ Example result for running `python main.py scheduler show-schedule`:
 [OrTrigger]: https://apscheduler.readthedocs.io/en/master/api.html#apscheduler.triggers.combining.OrTrigger
 [APScheduler scheduling docs]: https://apscheduler.readthedocs.io/en/master/userguide.html#scheduling-tasks
 [trigger docs]: https://apscheduler.readthedocs.io/en/master/api.html#triggers
-[registering-workflows]: ../../../getting-started/workflows#register-workflows
-[cli-docs]: ../../../reference-docs/cli/#orchestrator.cli.scheduler.show_schedule
+[registering-workflows]: ../getting-started/workflows.md#register-workflows
+[cli-docs]: ../reference-docs/cli.md/#orchestrator.core.cli.scheduler.show_schedule
```
Mark90, line 407: Wow. How do these 404s keep slipping through even with `--strict` mkdocs building.. nice spot

orchestrator/core/cli/generator/templates/create_data_head.j2
```diff
@@ -0,0 +1,36 @@
+# Copyright 2024-2026 SURF, GÉANT.
+# Licensed under the Apache License, Version 2.0 (the "License");
```
Mark90, line 2: Seems like the license headers will also end up in the generated files for users?  We'll probably want to exclude these templates from the license header check, if that's the case

orchestrator/core/api/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019-2020 SURF.
```
Mark90, line 1: Hela

orchestrator/core/websocket/managers/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019-2020 SURF.
+# Copyright 2026 GÉANT.
```
Mark90: Hela 2

kvklink: Yikes, I must have been a bit too enthusiastic with my `sed`

orchestrator/core/cli/generator/templates/create_data_head.j2
kvklink: The license headers will indeed end up in the generated files. However, strictly speaking they also _should_ be in there, since our copyright also applies to those generated files. The header should subsequently get updated like below by the end user.

```
# Copyright 2024-2026 SURF, GÉANT, ACME.
# Licensed under the Apache License, Version 2.0 (the "License");
```

kvklink: The other option is that I remove copyright headers from *all* test files, then we may be technically not 100% correct, but at least we'll consistent 🙂

tjeerddie (approved):
LGTM

orchestrator/core/cli/generator/templates/create_data_head.j2
Mark90: You're right. It's fine like this :)

Mark90 (approved):
LGTM

PR #1594, opened 2026-04-23
Update LLMSettings and Dockerfile

Mark90:
## Summary

- Add LLM setting `EMBEDDING_API_ENABLED` to explicitly enable the use of litellm to retrieve embeddings
  - `False` (default): to support completely local structured/fuzzy search with no embedding service needed
  - `True`: retrieve embeddings from external (local or remote) service for semantic/hybrid serach
- Add LLM setting `EMBEDDING_ENCODING_FORMAT`: defaults to `"float"` matching the embedding format
- Distinguish between different litellm exceptions
- Update `Dockerfile` to use uv for optimal usage in example-orchestrator, and add `.dockerignore`

## Related Issues

Fixes for https://github.com/workfloworchestrator/example-orchestrator/issues/80

## Checklist


pboers1988 (approved):

orchestrator/search/core/embedding.py
```diff
@@ -43,21 +43,40 @@ def get_embeddings_from_api_batch(cls, texts: list[str], dry_run: bool) -> list[
...
+
+        Callers must treat None as "embedding unavailable" and fall back to fuzzy/
+        structured search rather than passing it to a vector similarity query.
+        """
         if not text:
-            return []
+            return None
+
+        # TODO decide if this is the proper way/player to check if embedding service is enabled
+        #  This is needed to avoid expensive i/o roundtrips
```
Mark90: ^

orchestrator/search/core/embedding.py
```diff
@@ -38,36 +38,55 @@ def get_embeddings_from_api_batch(cls, texts: list[str], dry_run: bool) -> list[
...
+            logger.debug("Embedding API not enabled, search functionality restricted to fuzzy/structured search")
+            return None
+
         try:
             resp = await llm_aembedding(
                 model=llm_settings.EMBEDDING_MODEL,
                 input=[text.lower()],
                 api_key=llm_settings.EMBEDDING_API_KEY,
                 api_base=llm_settings.EMBEDDING_API_BASE,
+                encoding_format=llm_settings.EMBEDDING_ENCODING_FORMAT,
```
timfdev, line 85: might be change in litellm convention, needed I guess

docs/guides/upgrading/5.0.md
```diff
@@ -371,6 +371,7 @@ search will still function normally.
 To enable semantic search, set the following environment variables:
 
 ```bash
+EMBEDDING_API_ENABLED=True
```
timfdev, line 374: Another approach could be to request fuzzy from the client, it would be a nicer UX but arguably confusing as the client would need to know if there is an embedding api configured. 

Requires a fix that is needed regardless in `engine.py` because it tries to generate embeddings even if we request fuzzy search.

Relevant code snippets:
* _get_retriever_from_override
* Retriever.route
 * _execute_search:148:157
* Retriever classes

I think each retriever should have a plan method that determines if it needs an embedding, it can then replace _get_retriever_from_override, which is currently called AFTER generating an embedding.

I don't like an extra flag but it's probably the only solution here.

PR #1595, opened 2026-04-24
Don't generate embeddings if EMBEDDING_API_ENABLED is false

Mark90:
## Summary

When running any of the `python main.py index <x>` commands the command will raise errors as it tries to contact the Embedding API even if `EMBEDDING_API_ENABLED` is false, which it is in case of purely local search. This change adds a check to the EmbeddingIndexer (similar as done in the QueryEmbedder) to return empty embeddings in that case, along with an `INFO` message to inform users.

## Related Issues

https://github.com/workfloworchestrator/example-orchestrator/issues/80

## Checklist

### DIFF
--- a/orchestrator/core/search/core/embedding.py
+++ b/orchestrator/core/search/core/embedding.py
@@ -10,7 +10,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+from typing import Any
 
 import structlog
 from litellm import aembedding as llm_aembedding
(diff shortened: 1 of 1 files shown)


pboers1988 (approved):

orchestrator/core/search/core/embedding.py
```diff
@@ -24,13 +24,21 @@
...
     @classmethod
     def get_embeddings_from_api_batch(cls, texts: list[str], dry_run: bool) -> list[list[float]]:
         if not texts:
             return []
         if dry_run:
             logger.debug("Dry Run: returning empty embeddings")
-            return [[] for _ in texts]
+            return EmbeddingIndexer._empty_embeddings(texts)
+
+        if not llm_settings.EMBEDDING_API_ENABLED:
```
pboers1988, line 38: Can't we detect this from other settings? For example when the api key is not set.

Mark90: As far as I know both of the other 2 settings can be falsey (empty string / None) so this wasn't possible.

@timfdev is working on changing the routing so that embeddings are only retrieved for queries that need them. That could perhaps make this setting obsolete again.

Mark90: https://github.com/workfloworchestrator/orchestrator-core/pull/1596/changes

PR #1596, opened 2026-04-26
refactor routing so that embeddings are only generated when a retriev…

timfdev:
…er needs them, defined in Retriever baseclass

## Related Issues
Fixes # (issue number)

## Checklist
- [ ] My code follows the style guidelines of this project as described in [CONTRIBUTING](https://github.com/workfloworchestrator/orchestrator-core/blob/main/docs/contributing/guidelines.md).

### DIFF
--- a/orchestrator/core/search/retrieval/retrievers/base.py
+++ b/orchestrator/core/search/retrieval/retrievers/base.py
@@ -17,7 +17,7 @@ from decimal import Decimal
 import structlog
 from sqlalchemy import BindParameter, Numeric, Select, literal
 
-from orchestrator.core.search.core.types import EntityType, FieldType, SearchMetadata
+from orchestrator.core.search.core.types import EntityType, FieldType, RetrieverType, SearchMetadata
 from orchestrator.core.search.query.queries import ExportQuery, SelectQuery
 
 from ..pagination import PageCursor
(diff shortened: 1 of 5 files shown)


orchestrator/core/search/retrieval/retrievers/base.py
```diff
@@ -40,56 +40,139 @@ class Retriever(ABC):
...
+            retriever_cls = RrfHybridRetriever
+        elif query.vector_query:
+            retriever_cls = SemanticRetriever
+        elif query.fuzzy_term:
+            retriever_cls = FuzzyRetriever
+        else:
+            retriever_cls = StructuredRetriever
+
+        if query.entity_type == EntityType.PROCESS and retriever_cls in (FuzzyRetriever, RrfHybridRetriever):
+            return ProcessHybridRetriever
```
Mark90, line 75: I wonder why process is such a special case?

timfdev: Of the top of my head it's because processes contain too much information. For example they might have a list of a few hundred uuid's in one of its fields, which is also meaningless information. 

So we don't index this, instead we join on the process step table so we can still fuzzy search on jsonb keys in that table.

Mark90 (approved):

pboers1988 (approved):

pboers1988 (approved):

PR #1607, opened 2026-05-03
Cleanly separate unit and integration tests (#1277)

### LINKED ISSUE #1277: [Maintenance]: Integration tests
From discussion:

_Originally posted by @Mark90 in https://github.com/workfloworchestrator/orchestrator-core/pull/1214#discussion_r2601713118_
            

The orchestrator-core unit tests are a mix of unit and integration tests (=tests which require Postgres/Redis, or which setup a process like the scheduler, etc). To make development easier, it would be nice to logically separate them. 

Characteristics of each:

**Unit tests**
* No external services required (postgres/redis)
* Fast execution, should be possible to have a lot of unit tests
* CI: 
  * Runs these first
  * Test with matrix of
    * all supported python versions 
    * (optional) with different versions of libraries

**Integration tests**
* Require external services (postgres/redis)
* Execution may be slow, we should be selective in which things to test
* CI:
  * Runs these only after successful unittests
  * Test with matrix of
    * oldest/latest python version
    * different versions of postgres
    * different versions of redis 

It would also be very nice if we could use https://testcontainers.com/?language=python to no longer require developers to manually spin up / clean up / manage a docker compose stack to run the integration tests. (check support for Github CI and if users can still "bring their own database" for tests, i.e. when running in example-orchestrator)

Steps/tasks:
- Identify current unit tests that are integration tests
- Move them to the integration tests directory
  - Consolidate with existing celery tests which are a perfect example of integration tests
- Try to use testcontainers 
- Update CI definitions
- Document testing strategy in contribution docs
- Document the use of `orig()` for unittesting step functions in orchestrator implementations

torkashvand:
Closes [#1277](https://github.com/workfloworchestrator/orchestrator-core/issues/1277).

## Summary

This PR separates the current test suite into real unit tests and integration tests.

Before this change, running `test/unit_tests/` still required Postgres, Redis, Alembic migrations, the FastAPI app, APScheduler, and Celery-related setup. That made the “unit” test suite heavier than expected and harder to run locally.

After this change:

- `test/unit_tests/` contains only tests that run without Postgres, Redis, Docker, migrations, app boot, scheduler, or Celery.
- DB-, API-, workflow-, scheduler-, Celery-, and LLM/search-backed tests live under `test/integration_tests/`.
- Heavy fixtures moved out of the unit-test `conftest.py`.
- Integration tests can use existing `DATABASE_URI` / `CACHE_URI` values, or fall back to testcontainers when both are unset.
- CI now runs unit tests first, then runs the integration suites after the unit-test job passes.

## Test layout

| Area | Location | Description |
|---|---|---|
| Unit tests | `test/unit_tests/**` | Pure tests only. No external services, no DB, no Redis, no app boot. |
| Standard integration tests | `test/integration_tests/**` excluding Celery and LLM | REST/GraphQL, DB persistence, workflows, scheduler, metrics, migrations, search indexing. |
| Celery integration tests | `test/integration_tests/test_with_pytest_celery.py` | Celery worker/broker integration tests. |
| LLM/search integration tests | `test/integration_tests/search/llm/**` | LLM retrieval, embedding, and ranking tests. |
| Shared test helpers | `test/helpers.py` | Pure helpers shared by unit and integration tests. |

### Split unit and integration fixtures

The old `test/unit_tests/conftest.py` contained a lot of integration setup. It opened database connections, ran migrations, started the FastAPI app, initialized scheduler state, and provided product/subscription/workflow factories.

That setup now lives in the integration layer.

| File | Purpose |
|---|---|
| `test/unit_tests/conftest.py` | Lightweight unit-test fixtures only. |
| `test/integration_tests/conftest.py` | DB, app, scheduler, cache, workflow, product/subscription factories, and other integration fixtures. |
| `test/integration_tests/_session_setup.py` | Resolves Postgres and Redis before importing orchestrator modules. |
| `test/integration_tests/fixtures/services.py` | Handles `DATABASE_URI` / `CACHE_URI`, testcontainers fallback, and `app_settings` patching. |
| `test/integration_tests/fixtures/**` | Heavy product, subscription, workflow, and process factories. |

This means a misclassified test now fails clearly instead of silently pulling in the full integration stack.

### Moved integration tests out of `test/unit_tests`

Tests were audited and moved based on what they actually need.

Integration tests now include:

- FastAPI router tests
- GraphQL resolver and mutation tests
- database persistence tests
- workflow engine tests
- scheduler tests
- metrics tests
- websocket tests
- CLI migration tests
- DB-backed search/indexing tests
- Celery tests
- LLM/search tests

During the classification step, a few files were checked more closely and kept in `test/unit_tests/` because they are pure mock-based tests.

Several files were also split at function level so that pure tests stay in unit tests, while only the few integration-dependent tests moved.

Assertions were not weakened during these moves.

### Added integration service provisioning

Integration tests now resolve Postgres and Redis in this order:

1. Existing `DATABASE_URI` and `CACHE_URI` values.
2. Testcontainers fallback, when both variables are unset and Docker is available.
3. A clear error message if neither env vars nor Docker are available.

This preserves the existing contract for CI, `docker/pytest-support/docker-compose.yaml`, `example-orchestrator`, and downstream users.

Partial configuration is rejected. For example, setting only `DATABASE_URI` but not `CACHE_URI` fails early, so we do not accidentally mix a real database with an ephemeral Redis instance.

`testcontainers[postgres,redis]` is added only to the dev dependency group. Runtime dependencies are unchanged.

### Updated CI

CI now has one main entry point for this flow: `run-tests.yml`.

It runs:

1. `unit-tests`
2. `standard-integration-tests`
3. `celery-integration-tests`
4. `llm-search-integration-tests`

The three integration jobs all depend on `unit-tests`.

| Workflow | Change |
|---|---|
| `run-unit-tests.yml` | Reusable/manual only. No Postgres, no Redis. Runs Python 3.11–3.14. |
| `run-tests.ymll` | New main PR/push workflow. Runs unit tests first, then integration suites. |
| `run-pytest-celery.yml` | Reusable/manual only. Called from `run-tests.yml`. |
| `run-llm-integration-tests.yml` | Reusable/manual only. Called from `run-tests.yml`. |

The unit matrix runs once per PR.

The standard integration matrix covers a small set of Python/Postgres/Redis combinations, so CI covers the important supported versions without making the matrix too large.

### Updated docs

Updated documentation includes:

- how to run unit tests,
- how to run standard integration tests,
- how to run Celery tests,
- how to run LLM/search tests,
- how to provide Postgres/Redis manually,
- how to use the existing docker-compose stack,
- how testcontainers fallback works,
- where to put new tests,
- how to use `orig()` when unit-testing decorated workflow step functions.

Updated files:

- `docs/contributing/development.md`
- `docs/guides/testing.md`

### Unit tests

```bash
unset DATABASE_URI CACHE_URI

uv run pytest test/unit_tests orchestrator/core \
  --ignore=orchestrator/core/devtools \
  --ignore=examples \
  --ignore=docs
```

### Standard integration tests

With existing Postgres and Redis:

```bash
DATABASE_URI=... CACHE_URI=... \
uv run pytest test/integration_tests \
  --ignore=test/integration_tests/test_with_pytest_celery.py \
  --ignore=test/integration_tests/search/llm
```

Without `DATABASE_URI` / `CACHE_URI`, pytest will try to use testcontainers if Docker is available.

### Celery integration tests

```bash
DATABASE_URI=... CACHE_URI=... \
uv run pytest test/integration_tests/test_with_pytest_celery.py
```

### LLM/search integration tests

```bash
DATABASE_URI=... CACHE_URI=... \
uv run pytest test/integration_tests/search/llm
```

## What did not change

This PR intentionally does not change:

- production code under `orchestrator/`,
- runtime dependencies,
- the `DATABASE_URI` / `CACHE_URI` contract,
- `test/acceptance_tests/`,
- the existing docker-compose based test setup,
- downstream usage from `example-orchestrator`.


kvklink (approved):
Very nice cleanup of our test suite!

.github/workflows/run-codspeed-tests.yml
```diff
@@ -75,7 +75,7 @@ jobs:
       - uses: CodSpeedHQ/action@v4
         with:
           mode: simulation
-          run: CACHE_URI=redis://redis DATABASE_URI=postgresql+psycopg://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_HOST/$POSTGRES_DB uv run pytest test/unit_tests --codspeed
+          run: CACHE_URI=redis://redis DATABASE_URI=postgresql+psycopg://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_HOST/$POSTGRES_DB uv run pytest test/integration_tests/domain/test_base_performance.py --codspeed
```
Mark90: Several codspeed tests don't have a `pytest.mark.benchmark` but instead include the `benchmark` fixture and then use `@benchmark` on a closure function, so as to exclude setup/teardown logic from the measurement. You can safely target the root of the integration tests and add `--codspeed`, on my machine this runs the 13 expected tests

```suggestion
          run: CACHE_URI=redis://redis DATABASE_URI=postgresql+psycopg://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_HOST/$POSTGRES_DB uv run pytest test/integration_tests --codspeed
```

Mark90: It would actually be nicest if we can run `pytest test/integration_tests test/unit_tests --codspeed` because we'll also want to have benchmarks in unit tests. But that's not possible at the moment due to 

> ERROR test/integration_tests/search/llm - ValueError: option names {'--record'} already added

(leaving another comment about the llm tests further down, they're a special case)

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,119 @@
...
+      - name: Run integration tests (DB + Redis + FastAPI app + workflows)
+        # Excludes:
+        #   - Celery tests: covered by run-pytest-celery.yml
+        #   - LLM/embedding search tests: covered by run-llm-integration-tests.yml
+        # The DB-backed (non-LLM) search tests under test/integration_tests/search/
+        # ARE included here.
+        run: |
+          uv run pytest test/integration_tests \
+            --ignore=test/integration_tests/test_with_pytest_celery.py \
+            --ignore=test/integration_tests/search/llm \
```
Mark90: Ah, I forgot about these "specials"..

I think it makes sense to move them into a 3rd category, acceptance tests (which we have but I'm not sure how relevant the current tests in there are), so that we have all different types organized:


> 1. Unit tests — Test individual functions/classes in isolation, dependencies mocked. Fast, numerous.
> 2. Integration tests — Test multiple components working together (e.g., database queries, API routes with a real DB).
> 3. Acceptance tests — Test full user-facing flows through the entire stack.


With the difference between 2 and 3 being that for 2 the test structure "must" reflect the code structure (so quite similar to unit test structure, just with fixtures for the DBs instead of fixtures for mocking), whereas for 3 the test structure is completely free-form so that it can test the application in a black box way.

In that same context I think that `test/unit_tests/cli/test_generate_code.py` also belongs in the acceptance tests, as it's quite slow (due to filesystem interaction) and is not directly tied to the code structure.

What do you think?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,119 @@
...
+      - name: Install codecov dependencies
+        run: apt-get update && apt-get install -y git curl gpg
+
+      - name: "Upload coverage to Codecov"
+        uses: codecov/codecov-action@v6
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: false
+          files: ./coverage.xml
+          flags: integration
```
Mark90, line 108: Regarding this MR description

> So the real reason for the drop is two-fold, not one:
> 
> * Coverage architecture changed — the old "unit tests" workflow accidentally covered integration paths because it had services attached. This PR properly separates them, so the unit slice is honestly small.
> * Fewer matrix rows — main ran the unit-with-services workflow across Python {3.11..3.14} × Postgres {15, 16, 17} = 12 rows. Each row covered the same wide slice. This PR runs Python {3.11..3.14} for true-unit (4 rows, narrow slice) + 2 integration rows (different slice) + 1 each for Celery and LLM.

Isn't it just the unittests coverage report that's being used by codecov? When I run it locally the unittests cover 60% (which is what the online codecov report showed) whereas the integrationtests coverage is 72%, and it covers several branches that codecovs shows as "missed"

We should be able to combine those 2 in the end coverage somehow, perhaps by setting the `flags:` parameter on the unit tests as well.
https://docs.codecov.com/docs/merging-reports
https://docs.codecov.com/docs/flags

docs/contributing/development.md
```diff
@@ -26,33 +26,116 @@ More details in [About UV](#about-uv).
...
 export DATABASE_URI=postgresql+psycopg://nwa:nwa@localhost:5432/orchestrator-core-test
+export CACHE_URI=redis://:nwa@localhost:6379/0   # note the ':nwa@' password
+uv run pytest test/integration_tests \
+  --ignore=test/integration_tests/test_with_pytest_celery.py \
+  --ignore=test/integration_tests/search/llm
 ```
 
-``` shell
-uv run pytest test/unit_tests
+**Option C — Let testcontainers start ephemeral services:**
```
Mark90: Maybe advertise this as Option A, easiest solution for lazy developers (like me) :)

Mark90 (commented):
Not fully done reviewing yet, mostly been checking the structure and trying tests locally, looks awesome so far! Left a few comments above

.github/workflows/run-tests.yml
Mark90: A middle ground for now could be to only move the code generation to the integration tests, and leave the rest as is

```
$ pytest test/unit_tests --durations=0
...
======================================================= slowest durations =======================================================
24.77s setup    test/unit_tests/cli/test_generate_code.py::test_missing_or_extra_files
1.01s call     test/unit_tests/services/test_process_broadcast_thread.py::test_process_data_broadcast_thread_stop_joins_thread
1.01s call     test/unit_tests/services/test_process_broadcast_thread.py::test_process_data_broadcast_thread_processes_queue_item
1.00s call     test/unit_tests/services/test_process_broadcast_thread.py::test_process_data_broadcast_thread_stop_sets_shutdown
0.50s call     test/unit_tests/services/test_worker_status_monitor.py::test_monitor_stops_cleanly
0.21s call     test/unit_tests/distlock/test_memory_manager.py::test_non_expired_lock_survives
...
```

docs/contributing/development.md
```diff
@@ -26,33 +26,116 @@ More details in [About UV](#about-uv).
...
+```shell
+uv run pytest
 ```
 
-Set the password to something simple, like `nwa`.
+This runs every test under `test/unit_tests/` plus the doctests in
+`orchestrator/core`. No Postgres, no Redis, no Docker — fast feedback for
+day-to-day development.
+
+### Run integration tests
```
Mark90, line 54: There's also a `## Contributing` section in the readme which uses `just` to run tests, the `Justfile` also needs to be updated. Probably the "spin up docker compose" stuff in it can be removed in favor of using testcontainers?

test/unit_tests/search/retrieval/test_validation_exceptions.py
```diff
@@ -188,21 +173,6 @@ async def test_wildcard_path_bypasses_entity_prefix_check(mock_validate_path):
     mock_validate_path.assert_called_once_with("*.name")
```
Mark90, line 174: What was the reasoning of keeping, for example, `test_wildcard_path_bypasses_entity_prefix_check` as a unit test but moving `test_ltree_filter_takes_special_path` to integration tests?

test/unit_tests/workflows/test_utils.py
```diff
@@ -167,11 +166,6 @@ def test_workflow() -> StepList:
     assert test_workflow.description == ""
 
 
-def test_obsolete_step():
```
Mark90, line 170: Isn't this a unittest?

pyproject.toml
```diff
@@ -221,6 +222,21 @@ override-dependencies = [
...
+#
+# Integration tests are split into three suites that mirror CI:
+#   * standard:  uv run pytest test/integration_tests \
+#                  --ignore=test/integration_tests/test_with_pytest_celery.py \
+#                  --ignore=test/integration_tests/search/llm
+#   * celery:    uv run pytest test/integration_tests/test_with_pytest_celery.py
+#   * llm:       uv run pytest test/integration_tests/search/llm
+#
+# ``uv run pytest test/integration_tests`` (no ignores) runs the union of all
+# three; see docs/guides/testing.md for details.
```
Mark90: You've already described this extensively in the documentation. The fewer places something is documented, the better, IMHO :)

test/unit_tests/workflows/test_utils.py
torkashvand: Looks like a unit test, but obsolete_step is @step-decorated. Calling obsolete_step({}) runs the workflow engine, which reads     
  db.session. Without a real DB the call returns Failed, so assert result.issuccess() blows up so has to be in integration.

pboers1988 (approved):
Nice work @torkashvand! I

pyproject.toml
pboers1988: Agree!

pboers1988:
Once merged we need to change some of the Branch protection rules.

test/acceptance_tests/cli/data/generate/products/product_blocks/example1.py
```diff
@@ -16,10 +16,10 @@
 from typing import Annotated
 
 from annotated_types import Ge, Le, Len
-from orchestrator.core.domain.base import ProductBlockModel
-from orchestrator.core.types import SI, SubscriptionLifecycle
 from pydantic import computed_field
 
+from orchestrator.core.domain.base import ProductBlockModel
+from orchestrator.core.types import SI, SubscriptionLifecycle
```
Mark90, line 20: You're still working on this probably - just tested the cli acceptance tests locally and it shows some differences on these generated files, most likely the automatic code formatting reorganized these imports

```
Generated products/product_blocks/example1.py differs!
*** expected products/product_blocks/example1.py
--- actual products/product_blocks/example1.py
***************
*** 18,23 ****
  from annotated_types import Ge, Le, Len

  from pydantic import computed_field



- from orchestrator.core.domain.base import ProductBlockModel

- from orchestrator.core.types import SI, SubscriptionLifecycle

  from products.product_blocks.example2 import Example2Block, Example2BlockInactive, Example2BlockProvisioning

--- 18,23 ----
  from annotated_types import Ge, Le, Len

+ from orchestrator.core.domain.base import ProductBlockModel

+ from orchestrator.core.types import SI, SubscriptionLifecycle

  from pydantic import computed_field



  from products.product_blocks.example2 import Example2Block, Example2BlockInactive, Example2BlockProvisioning

```

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,124 @@
...
+    name: Celery integration tests
+    needs: unit-tests
+    uses: ./.github/workflows/run-pytest-celery.yml
+    secrets: inherit
+
+  llm-search-integration-tests:
+    name: LLM/embedding search integration tests
+    needs: unit-tests
+    uses: ./.github/workflows/run-llm-integration-tests.yml
+    secrets: inherit
```
Mark90, line 126: You probably weren't finished yet on this, maybe name these workflows/files to something like `run-<testsuite>-acceptation-tests.yml` where `testsuite` is celery/cli/llm?

Mark90 (approved):
The CI has gotten a lot quicker, the feedback loop for basic development errors is very quick now, and we can extend the test suite much more easily in the future.

Thanks so much for putting in this effort, it is tremendous!

.github/workflows/run-cli-acceptance-tests.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Install codecov dependencies
+        run: apt-get update && apt-get install -y git curl gpg
+
+      - name: "Upload coverage to Codecov"
+        uses: codecov/codecov-action@v6
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: false
+          files: ./coverage.xml
+          flags: acceptance
```
Mark90:
```suggestion
          flags: cli
```

PR #1644, opened 2026-05-18
Fix tasks with form input returning error on the first page

### LINKED ISSUE #1639: [Bug]: Task "Validate all subscriptions of Product Type" shows a validation error on the first page
### Contact Details

_No response_

### What happened?

# Problem

Trying to start the task `Validate all subscriptions of Product Type` (`task_validate_product_type`) is not possible as on the first page it directly shows the following validation error:

```
1 input field(s) have validation errors
```

# Cause

To be determined. It seems to have been introduced in the changes between 5.0.0rc3 and 5.0.0.

Edit: I think that the cause is this https://github.com/workfloworchestrator/orchestrator-core/pull/1632

### Version

5.0.0

_No response_

### Relevant log output

```shell

```

tjeerddie:
Fixes regression from #1632 

Closes #1639

### DIFF
--- a/orchestrator/core/forms/scheduler_form.py
+++ b/orchestrator/core/forms/scheduler_form.py
@@ -209,7 +209,7 @@ async def configure_schedule_form(state: State) -> FormGeneratorAsync:
         "workflow_name": task_name,
         "trigger": schedule_type_data["schedule_type"].name.lower(),
         "trigger_kwargs": trigger_kwargs,
-        "user_inputs": user_inputs,
+        "user_inputs": user_inputs if user_inputs else [{}],
         "scheduled_type": "create",
         "name": description,
     }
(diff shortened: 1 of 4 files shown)


orchestrator/core/forms/scheduler_form.py
```diff
@@ -209,7 +209,7 @@ class ScheduleDateForm(SubmitFormPage):
         "workflow_name": task_name,
         "trigger": schedule_type_data["schedule_type"].name.lower(),
         "trigger_kwargs": trigger_kwargs,
-        "user_inputs": user_inputs,
+        "user_inputs": user_inputs if user_inputs else [{}],
```
Mark90, line 212: Wondering why this isn't stored with the schedule definition at the moment

Mark90: Ok so this is needed because of 

```py
# submodules/orchestrator-core/orchestrator/core/workflow.py:255

    if initial_input_form is None:
        # We always need a form to prevent starting a workflow when no input is needed.
        # This would happen on first post that is used to retrieve the first form page
        initial_input_form = cast(InputStepFunc, const(FormPage))
```

Mark90 (approved):

PR #1645, opened 2026-05-18
Fix scheduled tasks lost after 5.0 namespace change (#1638)

### LINKED ISSUE #1638: [Bug]: Deploying release 5.0.0 clears scheduled tasks
### Contact Details

_No response_

### What happened?

# Problem

Deploying release 5.0.0 clears scheduled tasks from the database.

# Caus

The function `run_start_workflow_scheduler_task()` in `.../schedules/service.py` is a wrapper function that executes scheduled tasks by name. This wrapper decouples the actual task code from the schedule definition, to be able to move or alter the task code freely without worrying about invalidating existing schedules (so long as the name stays the same, of course).

Unfortunately in 5.0 the entire `orchestrator` package was namepaced to `orchestrator.core`, which probably does exactly that: it invalidates scheduled tasks that use `run_start_workflow_scheduler_task()` - so all of them.

# Solution

We could try to add a DB migration that updates all existing apscheduler definitions so as to keep them valid, but they contain pickled objects so I'm not sure how difficult this will be. It would also require users to not make the (easy) mistake of deploying the 5.0.0 migrations while the scheduler is still running.

The least thing we can and should do is warn users in the 5.0 upgrade guide to create and restore a backup of their scheduled tasks.

Also, there should be a warning in the `run_start_workflow_scheduler_task()` function for future developers touching this code that they should not move or rename that function. (maybe a DB migration is handy, after all)

### Version

5.0.0

_No response_

### Relevant log output

```shell

```

tvdven:
## Summary

  Deploying v5.0 silently invalidates all existing scheduled tasks because APScheduler persists the fully qualified Python path of the task function in pickled job state (`apscheduler_jobs.job_state`). The v5.0 namespace change from
   `orchestrator.*` to `orchestrator.core.*` breaks these stored references, causing the scheduler to skip all previously created tasks.

  This PR:

  - **Adds a database migration** (`cab8b6a0ac92`) that deserializes each APScheduler job state, rewrites `func` references from the old `orchestrator.` prefix to `orchestrator.core.`, and re-serializes them. The migration is fully
  reversible.
  - **Adds a warning docstring** to `run_start_workflow_scheduler_task()` alerting future developers not to move or rename the function without a corresponding data migration.
  - **Adds upgrade documentation** to the 5.0 guide under the namespace packaging section, warning users to stop the scheduler and back up the `apscheduler_jobs` table before migrating.

  ## Test plan 
  
  - [ ] Verify the migration runs without errors on a database with existing scheduled tasks
  - [ ] Verify scheduled tasks are visible after migration (via GraphQL `scheduledTasks` query or `show-schedule` CLI)
  - [ ] Verify the downgrade path restores the old namespace references
  - [ ] Verify the upgrade documentation renders correctly in the docs site

  Closes #1638

### DIFF
--- a/orchestrator/core/schedules/service.py
+++ b/orchestrator/core/schedules/service.py
@@ -154,6 +154,11 @@ def _delete_linker_entry(workflow_id: UUID, schedule_id: str) -> None:
 def run_start_workflow_scheduler_task(workflow_name: str, user_inputs: list[State] | None = None) -> None:
     """Function to start a workflow from the scheduler.
 
+    Warning: Do not move or rename this function. Its fully qualified path is persisted in
+        pickled APScheduler job state (``apscheduler_jobs.job_state``). Changing the
+        path will invalidate all existing scheduled tasks. If a rename is unavoidable,
+        add a data migration to rewrite the stored ``func`` references.
+
     Args:
         workflow_name: str The name of the workflow to start.
         user_inputs: user inputs for the workflow to start.
--- a/orchestrator/core/schedules/service.py
+++ b/orchestrator/core/schedules/service.py
@@ -157,7 +157,8 @@ def run_start_workflow_scheduler_task(workflow_name: str, user_inputs: list[Stat
     Warning: Do not move or rename this function. Its fully qualified path is persisted in
         pickled APScheduler job state (``apscheduler_jobs.job_state``). Changing the
         path will invalidate all existing scheduled tasks. If a rename is unavoidable,
-        add a data migration to rewrite the stored ``func`` references.
+        add a data migration to rewrite the stored ``func`` references — see
+        ``2026-05-18_cab8b6a0ac92_fix_scheduled_tasks_issue.py`` for an example.
 
     Args:
         workflow_name: str The name of the workflow to start.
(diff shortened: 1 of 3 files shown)


orchestrator/core/schedules/service.py
```diff
@@ -154,6 +154,11 @@ def _delete_linker_entry(workflow_id: UUID, schedule_id: str) -> None:
 def run_start_workflow_scheduler_task(workflow_name: str, user_inputs: list[State] | None = None) -> None:
     """Function to start a workflow from the scheduler.
 
+    Warning: Do not move or rename this function. Its fully qualified path is persisted in
+        pickled APScheduler job state (``apscheduler_jobs.job_state``). Changing the
+        path will invalidate all existing scheduled tasks. If a rename is unavoidable,
+        add a data migration to rewrite the stored ``func`` references.
```
Mark90: Maybe add something like "If absolutely necessary, a DB migration can be prepared to rewrite the function references: see 2026-05-18_cab8b6a0ac92_fix_scheduled_tasks_issue.py"

Mark90 (approved):

PR #1646, opened 2026-05-18
Define function `default_user_inputs()` and use it as default factory for APSchedulerJobCreate.user_inputs

Mark90:
## Summary

Alternative solution to #1632:

The payload for creating apscheduler jobs has a `user_inputs` field that defaults to an empty list.
This PR changes that to a list with a `{}` similar to other parts of the codebase.

To make this more visible, I've added it to a reusable `default_user_inputs()` and applied this where needed.


pboers1988 (approved):

Mark90:
Can't reproduce the problem anymore. Closing

PR #1676, opened 2026-06-03
Capture exception that fails workflow for error tracking

Mark90:
## Summary

* Adapt workflow error logging to capture the exception that failed a workflow. This allows for more fine-grained error tracking in Sentry, and client-side filtering with `before_send=`
* Pass the exception to `consolelogger.error("An exception occurred while executing the workflow step.")` for unexpected exceptions

### DIFF
--- a/orchestrator/core/workflow.py
+++ b/orchestrator/core/workflow.py
@@ -1551,7 +1551,7 @@ def _exec_steps(steps: StepList, starting_process: Process, dblogstep: StepLogFu
             process = process.map(lambda s: s | {"__last_step_started_at": nowtz().timestamp()})
             step_result_process = process.execute_step(step)
         except Exception as e:
-            consolelogger.error("An exception occurred while executing the workflow step.")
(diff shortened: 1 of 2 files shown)


orchestrator/core/workflow.py
```diff
@@ -1523,6 +1524,23 @@ def invalidate_status_counts() -> None:
     broadcast_invalidate_status_counts()
 
 
+def capture_workflow_failure(err: Any) -> None:
+    """Capture the exception from a failed workflow.
+
+    This may provide useful insights into application or infra errors.
+    However, this may also cause sentry issues for exceptions that are "business as usual".
+    In that case, you can pass a function `app.add_sentry(before_send=before_send)` in which you can evaluate
```
tjeerddie, line 1532: might be handy to also add a part about this in the docs if there isn't already

Mark90: done!

tjeerddie (approved):

Issue #1251, opened 2025-11-28
[Improvement]: Make the API endpoints async


Mark90:
The API currently has a 50/50% split on synchronous and asynchronous endpoints. This mix is proving problematic when attempting to make changes in the API, such as in #1067 where we want to add an async callback for authorizing Workflow access. Additionally, it's difficult to determine the optimal deployment configuration since sync endpoints benefit from having more uvicorn workers per process (so as to have more worker threads), whereas for async endpoints benefit from 1 uvicorn worker per process (which then has complete access to the (v)CPU) and having more processes.

The current proposal is to:
- make all endpoints async
- change all blocking IO function calls to be awaitable, either by:
  - changing the underlying function to be async (**preferred option**)
  - running the call in a thread
- add linting rules that prevent adding sync endpoints in the future

We should reuse knowledge gained from #628 where Alex implemented using sqlalchemy in async fashion. (which had 2 blockers: 1) [using psycopg3 in orchestrator-core](https://github.com/workfloworchestrator/orchestrator-core/issues/1097) causes table locking problems, likely we're doing something wrong; 2) ORM model attribute lazy loading will no longer work without adding `await` in front of it)

Issue #628, opened 2024-04-24
Investigate using sqlalchemy.ext.asyncio for graphql resolvers


Mark90:
We use fastapi and strawberry to utilise the power of asyncio. However, our asynchronous graphql resolvers perform synchronous database calls which block the event loop. This means that while the API process is waiting for a database query to return it will not handle new graphql requests, greatly reducing throughput.

Simply changing the graphql resolvers to be synchronous [should](https://github.com/strawberry-graphql/strawberry/discussions/3088) make them run in worker threads when using uvicorn & fastapi, but this doesn't seem to be the case in my tests (added below). I don't know if this is something you have to explicitly configure, if so I could try that to make a fair comparison.

At any rate, using asyncio is in line with the libraries we have chosen, so it makes sense to investigate async database queries. We could have a look at SQL Alchemy's [asyncio support](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#synopsis-orm) and attempt to use it within a graphql resolver and a REST endpoint. The 3rd testcase shows the increase in throughput this would warrant.

### Tests for reference

In each of these tests the graphql resolver performs a database call `pg_sleep(0.5)` to simulate a query taking 0.5 seconds.
Using the [locust](https://locust.io/) I have then tested maximum throughput by performing 30 concurrent graphql queries for a short period of time.

1. `postgres_async_sync.py` This is how we currently use async resolvers with sync DB calls. The maximum observed throughput is 2 rps.

```python
# `strawberry server postgres_async_sync`
from datetime import datetime as dt
import psycopg
import strawberry

DSN='postgresql://localhost:5432'
# db = psycopg.connect(DSN)

@strawberry.type
class SleepResult:
    msg: str

async def pg_sleep():
    db = psycopg.connect(DSN)  # not needed, but to make the comparison with the other script fair
    cur = db.execute('SELECT pg_sleep(0.5), version()')
    result = cur.fetchone()
    print(f"{dt.now().strftime('%H:%M:%S')} Performed query {__name__}")
    return SleepResult(msg=str(result))

@strawberry.type
class Query:
    sleepy: SleepResult = strawberry.field(resolver=pg_sleep)

schema = strawberry.Schema(query=Query)
print(f"Initialized {__name__}")
```

2. `fastapi_postgres_async_sync.py` Runs the previous example through uvicorn/fastapi. The maximum throughput is still 2 rps.

```python
# `uvicorn fastapi_postgres_async_sync:app`
[...]

PR #1692, opened 2026-06-11
#1251 - Async endpoints foundation

tvdven:
## Summary 

First step towards making all API endpoints async (#1251). This PR adds a guardrail against *new* sync endpoints, the async database foundation, and converts the first endpoint (`GET /api/health/`) as a reference implementation.

### Changes

**1. Guardrail against new sync endpoints**
- `test_no_new_sync_endpoints` pins the 33 currently-sync endpoints in a shrink-only allowlist. Adding a sync endpoint fails CI with a pointer to this issue; converting one forces removal from the list.
- Ruff `ASYNC` (flake8-async) rules enabled: flags blocking calls (`requests`, `open()`, `time.sleep`, subprocess) inside `async def`.

**2. Async database foundation**
- `Database` gains an async engine + session factory on the **same** `postgresql+psycopg` URL as the sync engine (psycopg 3 serves both modes — no second driver or URI to configure).
- `db.async_session()` returns an `AsyncSession` whose lifecycle the caller owns; `get_async_session` is the FastAPI dependency that yields one and guarantees close (and therefore rollback of anything uncommitted) after the request.
- Deliberately **no scoped/global async session**: explicit sessions avoid the implicit-IO pitfalls (lazy loading, expire-on-commit refresh) that blocked #628. `expire_on_commit=False` for the same reason.
- Async sessions wrap `WrappedSession` (`sync_session_class`), so the commit-disable safeguard behind `transactional()` remains available for converted legacy helpers. Atomic async write blocks should use the native `async with session.begin()` idiom instead.

**3. First conversion: health endpoint**
- `GET /api/health/` is now `async def` using the dependency; removed from the allowlist.

## How follow-up conversions should work

Two tools, per endpoint:

**Native async** — when the endpoint is a self-contained query, or a native async client exists:

```python
@router.get("/", response_model=str)
async def get_health(session: AsyncSession = Depends(get_async_session)) -> str:
    await session.execute(select(ProductTable.name).limit(1))
    return "OK"
```

**Threadpool offload** — the default for existing endpoints whose body relies on the scoped session, domain models, or lazy loading. The body stays sync and unchanged; only the signature changes:

```python
@router.get("/{process_id}", response_model=ProcessDeprecationsSchema)
async def show(process_id: UUID) -> ProcessTable:
    return await run_in_threadpool(_show_process, process_id)  # existing sync helper, untouched
```

ContextVars propagate into `run_in_threadpool`, so `db.session` inside the helper still resolves to the request's scoped session — `transactional()`, rollback-on-teardown, and the shared-session semantics all keep working.

Non-goals (stay sync by design): the workflow engine (`start_process`/`resume_process` run on their own threads with their own `database_scope()`) and domain-model hydration.

## Testing

- New integration tests: guardrail, async session behavior (query execution, same-URL pinning, commit-disable carry-over, dependency closes session).
- Health connection-failure test re-pointed to patch `AsyncSession.execute` — with the `Depends` pattern that's where a dead database actually fails.
- Known limitation: the integration rollback fixture only wraps the sync engine; async sessions read committed state. Fine for reads (health only reads); needs an async counterpart before the first async write path lands.

Issue: #1251


test/unit_tests/api/test_endpoint_guardrails.py
```diff
@@ -0,0 +1,78 @@
...
+    }
+)
+
+
+def _qualified_name(route: APIRoute) -> str:
+    endpoint = inspect.unwrap(route.endpoint)
+    return f"{endpoint.__module__}.{endpoint.__qualname__}"
+
+
+def test_no_new_sync_endpoints(fastapi_app):
```
Mark90, line 88: This should be a unittest. Will require some effort to create a fastapi_app fixture that mocks out the database and such

test/integration_tests/api/test_health.py
```diff
@@ -23,8 +24,7 @@ def test_get_health(test_client):
     assert response.json() == "OK"
 
 
-@mock.patch("orchestrator.core.db.db.session")
-def test_get_health_no_connection(mock_session, test_client):
-    mock_session.execute.side_effect = OperationalError("THIS", "IS", "KABOOM")
-    response = test_client.get("/api/health/")
+def test_get_health_no_connection(test_client):
```
Mark90: This should be a unittest as well

Mark90 (commented):
Looks like a sensible approach to me 👍 

I would prefer if we keep this open for next sprint, so that someone can continue and rewrite all endpoints to be async with the 2 approaches you described.

Given this foundation it should be doable within a sprint, and I think it's better for someone to do everything at once so that the implementation is as consistent as possible.


---

After reconsideration (and because the comments I had on this PR are resolved) I agree we can merge this now and create a follow-up ticket.

Mark90 (approved):

PR #1694, opened 2026-06-12
Bug process status stays on running instead of failed

### LINKED ISSUE #1624: [BUG]: Process status stays on RUNNING instead of FAILED on General tab of a subscription after failed workflow
## Problem

On the General tab of a detail page, the "Most recent validate" field stays on `RUNNING` after a process ends in a `FAILED` state. The UI never reflects the actual final status.

The root cause is that subscription caches are never invalidated when a process transitions to a failed end-state. Because the cache is stale, the frontend keeps displaying the last known status (`RUNNING`) rather than the true status (`FAILED`).

## Fix

A fix is to invalidate the subscription cache in `process_broadcast_fn` when the process ends in a status that is part of `UPDATE_SUB_STATUSES`. Currently the function only receives a `process_id`, which is not enough information to do this. The parameter should be changed to accept the full process object (or we need to use the id to fetch process from the DB again):

```python
def process_broadcast_fn(process: Process) -> None:
    try:
        broadcast_process_update_to_websocket(process.process_id)
        if process.last_status in UPDATE_SUB_STATUSES:
            for subscription in process.process_subscriptions:
                sync_invalidate_subscription_cache(subscription.subscription_id)
    except Exception as e:
        logger.exception(e)
```

Note: this is a breaking change for consumers who implemented this function themselves — call sites will need to be updated to pass the full process object.

## Additional consideration

Currently `process_broadcast_fn` lives in surf and is documented as a copy-paste snippet. It would be worth moving a default implementation into orchestrator-core directly, so consumers get correct behaviour out of the box rather than having to copy boilerplate that can go stale.

tvdven:
## Summary

  Fixes stale subscription status in the UI when a workflow ends. Every in-step cache invalidation (`unsync`/`resync`) runs while the process is still `RUNNING`, so the websocket cache never reflected the terminal status: `done` sets `COMPLETED` without invalidating, and failed/aborted
  processes skip `resync` entirely. As a result the UI kept showing the subscription's process as `Running` after the workflow had failed (or completed).

  ### Changes:
  - `_db_log_step` now broadcasts a cache invalidation for each related subscription when the process reaches a terminal status (`COMPLETED`, `FAILED`, `INCONSISTENT_DATA`, `API_UNAVAILABLE`, `ABORTED`).
  - Added `invalidate_subscription_cache_by_id` (and a sync wrapper) to `orchestrator.core.websocket`, which invalidates only the single subscription's cache entry rather than the whole `subscriptions` type.
  - Tests: parameterized integration tests verifying invalidation fires for terminal end states (`Failed`/`Abort`/`Complete`) and not for non-terminal ones (`Success`/`Suspend`), plus a unit test for the new websocket broadcast.

### As opposed to proposed change in issue

  1. **No breaking change.** The other approach changes the `BroadcastFunc` signature from `UUID` to `ProcessTable`, breaking every consumer that copy-pasted `process_broadcast_fn` from the docs. The current branch leaves that contract untouched and can ship in a patch release.

  2. **Guaranteed correctness in one place.** `_db_log_step` is the single point every executor passes through when the terminal status is written, so one hook covers all paths — versus patching three broadcast paths (queue thread, direct websocket, Celery callback) where a consumer
  overriding `process_broadcast_fn` would silently lose the invalidation.

  ## Related Issue
  Fixes #1624

### DIFF
--- a/orchestrator/core/services/processes.py
+++ b/orchestrator/core/services/processes.py
@@ -43,7 +43,11 @@ from orchestrator.core.types import BroadcastFunc
 from orchestrator.core.utils.datetime import nowtz
 from orchestrator.core.utils.errors import StartPredicateError, error_state_to_dict
 from orchestrator.core.utils.json import json_dumps, json_loads
-from orchestrator.core.websocket import broadcast_invalidate_status_counts, broadcast_process_update_to_websocket
+from orchestrator.core.websocket import (
+    broadcast_invalidate_status_counts,
+    broadcast_process_update_to_websocket,
+    sync_invalidate_subscription_cache_by_id,
+)
 from orchestrator.core.workflow import (
     CALLBACK_TOKEN_KEY,
     DEFAULT_CALLBACK_PROGRESS_KEY,
@@ -72,6 +76,17 @@ logger = structlog.get_logger(__name__)
 StateMerger = Merger([(dict, ["merge"])], ["override"], ["override"])
 
 SYSTEM_USER = "SYSTEM"
+
+# Terminal statuses that skip the workflow's `resync` step, so its subscription-cache
+# invalidation never runs. `_db_log_step` re-issues the invalidation for these to avoid a stale UI.
+UPDATE_SUB_STATUSES = frozenset(
+    {
+        ProcessStatus.FAILED,
+        ProcessStatus.INCONSISTENT_DATA,
+        ProcessStatus.API_UNAVAILABLE,
+        ProcessStatus.ABORTED,
+    }
+)
 
 _workflow_executor = None
 _active_threadpool_jobs = 0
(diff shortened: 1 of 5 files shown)


orchestrator/core/services/processes.py
```diff
@@ -343,6 +361,10 @@ def _db_log_step(
     if broadcast_func:
         broadcast_func(p.process_id)
 
+    if p.last_status in TERMINAL_SUB_STATUSES:
+        for subscription_id in {ps.subscription_id for ps in p.process_subscriptions}:
+            sync_invalidate_subscription_cache_by_id(subscription_id)
+
```
Mark90, line 367: The broadcast function was added to prevent asyncio runtime errors for Orchestrator implementors that use the `THREADPOOL` executor.

This could reintroduce that error. We ourselves won't notice as we're using the `CELERY` executor, but most people start out with the threadpool solution so it's crucial to keep that working.

We should move the above lines into the `ProcessDataBroadcastThread`'s broadcast action; likely merge it into the function `orchestrator.core.websocket.broadcast_process_update_to_websocket()`. Probably best to have that function do a DB query to get the process subscriptions.

(this function is also used by celery users through the `process_broadcast_fn` as documented [here](https://workfloworchestrator.org/orchestrator-core/guides/scaling/?h=process_broadcast_fn#implementing-the-worker), which is how the behavior is kept consistent for both architectures)

PR #1728, opened 2026-06-23
Bug process status stays on running instead of failed

### LINKED ISSUE #1624: [BUG]: Process status stays on RUNNING instead of FAILED on General tab of a subscription after failed workflow

tvdven:
## Summary                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
After a workflow finishes with a terminal status (FAILED, COMPLETED, etc.) the UI keeps showing the process as RUNNING. The websocket broadcast opened a fresh DB scope to read the process, but by the time it ran, the session had not yet committed — so it always read RUNNING.                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                                              
Additionally, subscription caches were never invalidated on terminal statuses when a workflow ended outside of a normal `_db_log_step` call, causing stale subscription state in the UI.                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                              
### Root cause                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
  In `_run_process_async`, `broadcast_func` was called inside the `db.session.commit()` try-block, before the commit completed. The broadcast function opened a new database scope, but the status update was not yet visible.                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                              
 ### Fix                                                                                                                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                                              
  - Move the `broadcast_func` call to after `db.session.commit()` in `_run_process_async`, so the broadcast always observes the committed terminal status.                                                                                                                                                                                                                                                          
  - Move subscription cache invalidation for terminal statuses from `_db_log_step` (PR #1694) into `broadcast_process_update_to_websocket_async`, where it runs with the correct committed state.                                                                                                                                                                                                                 
  - Have the sync `broadcast_process_update_to_websocket` delegate entirely to the async variant via `anyio.run(...)` to avoid duplicating logic.

### Related Issues
Fixes #1624

### DIFF
--- a/orchestrator/core/services/executors/threadpool.py
+++ b/orchestrator/core/services/executors/threadpool.py
@@ -96,7 +96,7 @@ def thread_start_process(
     # Trigger the task in the current thread or threadpool (depends on executor mode).
     pstat.update(state=pstat.state.map(lambda state: StateMerger.merge(state, input_data.input_state)))
     _safe_logstep_with_func = partial(safe_logstep, broadcast_func=broadcast_func)
-    return _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_with_func))
+    return _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_with_func), broadcast_func)
 
 
 def thread_resume_process(
@@ -137,7 +137,7 @@ def thread_resume_process(
 
     # Trigger the task in the current thread or threadpool (depends on executor mode).
     _safe_logstep_prep = partial(safe_logstep, broadcast_func=broadcast_func)
-    _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_prep))
+    _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_prep), broadcast_func)
     return pstat.process_id
 
 
--- a/orchestrator/core/services/executors/threadpool.py
+++ b/orchestrator/core/services/executors/threadpool.py
@@ -96,7 +96,7 @@ def thread_start_process(
     # Trigger the task in the current thread or threadpool (depends on executor mode).
     pstat.update(state=pstat.state.map(lambda state: StateMerger.merge(state, input_data.input_state)))
     _safe_logstep_with_func = partial(safe_logstep, broadcast_func=broadcast_func)
-    return _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_with_func), broadcast_func)
+    return _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_with_func), broadcast_func=broadcast_func)
 
 
 def thread_resume_process(
@@ -137,7 +137,7 @@ def thread_resume_process(
 
     # Trigger the task in the current thread or threadpool (depends on executor mode).
     _safe_logstep_prep = partial(safe_logstep, broadcast_func=broadcast_func)
-    _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_prep), broadcast_func)
+    _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_prep), broadcast_func=broadcast_func)
     return pstat.process_id
 
 
--- a/orchestrator/core/websocket/__init__.py
+++ b/orchestrator/core/websocket/__init__.py
@@ -10,6 +10,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import asyncio
 from typing import Any, cast
 from urllib.parse import urlparse
 from uuid import UUID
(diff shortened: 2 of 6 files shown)


orchestrator/core/services/executors/threadpool.py
```diff
@@ -96,7 +96,7 @@ def thread_start_process(
     # Trigger the task in the current thread or threadpool (depends on executor mode).
     pstat.update(state=pstat.state.map(lambda state: StateMerger.merge(state, input_data.input_state)))
     _safe_logstep_with_func = partial(safe_logstep, broadcast_func=broadcast_func)
-    return _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_with_func))
+    return _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_with_func), broadcast_func)
```
Mark90: nit: keyword argument name is missing


```suggestion
    return _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_with_func), broadcast_func=broadcast_func)
```

orchestrator/core/services/executors/threadpool.py
```diff
@@ -137,7 +137,7 @@ def thread_resume_process(
 
     # Trigger the task in the current thread or threadpool (depends on executor mode).
     _safe_logstep_prep = partial(safe_logstep, broadcast_func=broadcast_func)
-    _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_prep))
+    _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_prep), broadcast_func)
```
Mark90: nit: keyword argument name is missing


```suggestion
    _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_prep), broadcast_func=broadcast_func)
```

orchestrator/core/websocket/__init__.py
```diff
@@ -163,6 +171,17 @@ async def broadcast_process_update_to_websocket_async(
     await broadcast_invalidate_cache({"type": "processes", "id": "LIST"})
     await broadcast_invalidate_cache({"type": "processes", "id": str(process_id)})
 
+    with db.database_scope():
+        process = db.session.get(ProcessTable, process_id, options=[selectinload(ProcessTable.process_subscriptions)])
+
+        subscription_ids = set()
+
+        if process is not None and process.last_status in _TERMINAL_PROCESS_STATUSES:
+            subscription_ids = {ps.subscription_id for ps in process.process_subscriptions}
```
Mark90: Given that this is an async function this will block the eventloop. 
Perhaps you can use the async session for this? Or if that's too much off a hassle, there is also a commonly used pattern in the codebase to run sync code asynchronously through a shortlived thread

orchestrator/core/websocket/__init__.py
```diff
@@ -163,6 +171,17 @@ async def broadcast_process_update_to_websocket_async(
...
+
+    for subscription_id in subscription_ids:
+        await invalidate_subscription_cache_by_id(subscription_id)
```
Mark90: This will wait for each subscription id invalidation event to be sent sequentially. For a few this might not matter, but if there are processes with lots of subscriptions then it'd be nicer to fan out

A solution is to use `asyncio.gather` but this has the potential of doing _too much_ concurrently.

A while ago we added `gather_nice()` (in `nwastdlib.asyncio`) which I think fits well here:

```suggestion
    tasks = [invalidate_subscription_cache_by_id(subscription_id) for subscription_id in subscription_ids]
    await gather_nice(tasks, limit=10)
```

The limit=10 is the upper limit of concurrently running coroutines

Mark90:
Nice! Few comments regarding the async code.

Mark90 (approved):

PR #1742, opened 2026-06-29
Build container variants for mcp, celery and mcp+celery extras

pboers1988:
## What

Extends the container CI so each release publishes images for the optional extras, not just the base image.

- **Dockerfile**: add `ARG EXTRAS=""`; install becomes `uv pip install "$(ls /tmp/*.whl)${EXTRAS}"`. Empty default keeps base behavior identical.
- **build-push-container.yml**: fold the build job into a `variant` matrix — `base`, `mcp`, `celery`, `mcp-celery`. Each passes `build-args: EXTRAS=...` and a tag `suffix` (e.g. `5.1.0-mcp`, `latest-celery`, `edge-mcp-celery`).

## Verified

Local build of the `mcp` variant succeeds and `fastmcp` is present in the image. This PR also exercises the matrix build-test path (it touches `Dockerfile`).

## Notes

- Matrix is now 4 variants × 2 platforms = 8 builds per release; arm64 via QEMU is slow. Add `cache-from/to: type=gha` if build time bites.
- Variant images still default to the uvicorn `CMD` — they ship the extra deps but don't change the entrypoint. Add a per-variant command if the celery/mcp images should run the worker/mcp-server directly.


Mark90 (approved):
Looks good.

Seeing as each of the 4 images takes 2 minutes to build, maybe you can briefly try if build-push-container can be adapted to use the previously built images as a starting point / cache

https://docs.docker.com/build/ci/github-actions/cache/#github-cache

pboers1988:
Good call — added `type=gha` cache-from/cache-to, scoped per variant so the 4 variants don't thrash each other's cache. Pushed in f2443856.

PR #1744, opened 2026-06-29
Add optional timeout to callback_step

### LINKED ISSUE #1736: [Enhancement]: Optional timeout for callback_step (AWAITING_CALLBACK hangs indefinitely)
`callback_step` has no way to bound how long it waits for the external callback. If the external system never POSTs back — a lost callback after a network problem — the process sits in `AWAITING_CALLBACK` forever: no timeout, no escalation, and no scheduled task reaps stale awaiting-callback processes.

**Version:** orchestrator-core 5.0.4

**Where:** in `orchestrator/core/workflow.py`, `_awaitstep` returns `AwaitingCallback(state)` with no timeout argument; the callback route/token (`/api/processes/{pid}/callback/{token}`) has no TTL; nothing under `schedules/` expires awaiting-callback processes.

**Operational fallout.** The only escape from an indefinite hang is to abort the process. But a workflow that ran `set_status(PROVISIONING)` before the callback step leaves the subscription stuck in `PROVISIONING` after the abort (the `insync` flag is unrelated and doesn't help). Modify / validate / reconcile workflows are only offered on `ACTIVE` subscriptions, so the subscription can no longer be recovered *through the orchestrator* — its status has to be set back to `active` directly (DB / orchestrator-shell). Any product whose create/provision/etc. uses a callback to a flaky external system can hit this.

**Suggested fix.** Add an optional `timeout` (seconds) to `callback_step`, defaulting to `None` — i.e. current behaviour, wait indefinitely, no change for existing workflows. When set, the await is bounded: a scheduled task fails timed-out `AWAITING_CALLBACK` steps (storing the await start time + timeout on the process), turning a silent indefinite hang into a normal failed, retryable step that's visible and alertable.

```python
callback_step(
    name="...",
    action_step=...,
    validate_step=...,
    timeout=300,  # default None = wait forever
)
```

This keeps the default unchanged while letting workflow authors bound external interactions and recover them through the normal retry/abort flow instead of manual DB surgery.

hanstrompert:
Closes #1736.

A `callback_step` waits **indefinitely** in `AWAITING_CALLBACK`; if the external system never calls back, the process hangs forever and can only be recovered via the database.

This adds an optional `timeout` (seconds) to `callback_step`, default `None` = wait indefinitely (backward compatible). When set, a process past its deadline is moved to `FAILED` so it can be retried or aborted normally.

- Fail timed-out processes in place via `fail_awaiting_wf` / `fail_awaiting_process`, guarded against racing an arriving callback.
- New `task_validate_awaiting_callbacks` sweep (30s interval), gated by `awaiting_callbacks_exist` so it only runs when an awaiting callback has a timeout set.
- Register the task (LazyWorkflowInstance, migration, initial-schedule entry); document in `callbacks.md`.

Notes: `timeout` is a minimum (resolution ~30s); retry re-runs the whole callback group (re-issues the action step), so design it to be safe to repeat. Unit + integration tests added; `mypy`/`ruff` clean.

### DIFF
--- a/docs/reference-docs/workflows/callbacks.md
+++ b/docs/reference-docs/workflows/callbacks.md
@@ -303,3 +303,58 @@ For long-running jobs, such as executing Terraform or Ansible playbooks, the Orc
 Progress updates should be delivered to `{callback_route}/progress`, where `"/progress"` is a fixed endpoint appended to the callback URL.
 
 The remote service should send a JSON or plain string payload with each progress callback, this replaces the previous progress update and refreshes the UI. Once the final callback is triggered and the job completes, progress updates are removed, leaving only the callback result. Therefore, all troubleshooting or diagnostic information must be included in the final callback payload, as progress updates are not retained for debugging.
+
+## Timeouts
+
+By default a callback step waits **indefinitely** for the remote service to call back. If that
+service never responds (for example after a network failure), the process stays in
+`AWAITING_CALLBACK` forever and the only recovery is manual database intervention.
+
+To guard against this, pass an optional `timeout` (in **seconds**) to `callback_step`:
+
+```python
+from orchestrator.core.workflow import callback_step
+
+
+callback_step(
+    name="Call ext system",
+    action_step=call_ansible_playbook,
+    validate_step=_evaluate_callback_results,
+    timeout=300,  # fail if no callback arrives within 5 minutes
+)
+```
+
+`timeout=None` (the default) keeps the original behaviour of waiting forever, so existing
+callbacks are unaffected.
+
+### What happens when the timeout is exceeded
+
+When the deadline passes, the process is moved to `FAILED` with `failed_reason = "Callback timed
+out"`. This unblocks the standard recovery actions, so you no longer have to edit the database:
+
+- **Retry** re-runs the *entire callback step group*: it executes the action step again (re-issuing
+  the external request and generating a fresh callback URL) and waits anew. Design the action step
+  so it is safe to repeat (idempotent, or otherwise tolerant of being run more than once).
+- **Abort** marks the process `ABORTED` and stops it. Note that abort does **not** run the
+  workflow's remaining steps (the callback cleanup step is skipped) and does **not** itself change
+  the subscription's lifecycle state — any follow-up on the subscription is left to the operator or
+  a separate workflow.
+
+The deadline is measured from the moment the await started. Progress updates sent to
+`{callback_route}/progress` do **not** extend it.
+
+### Enforcement resolution
+
+Timeouts are enforced by the `task_validate_awaiting_callbacks` scheduled task, which runs every
+**30 seconds** by default. As a result `timeout` is a *minimum*: a process is failed on the first
+sweep at or after its deadline (so up to ~30 seconds late), and a `timeout` smaller than the sweep
+interval is effectively rounded up to the next sweep. Do not rely on sub-30-second precision.
+
+The task is registered like the other core tasks and is loaded with:
+
+```bash
+orchestrator-core scheduler load-initial-schedule
+```
+
+The interval is tunable through the scheduler API. The sweep does not start a run (and therefore
+creates no process record) when no process is awaiting a callback.
--- a/docs/reference-docs/workflows/callbacks.md
+++ b/docs/reference-docs/workflows/callbacks.md
@@ -345,16 +345,8 @@ The deadline is measured from the moment the await started. Progress updates sen
 
 ### Enforcement resolution
 
-Timeouts are enforced by the `task_validate_awaiting_callbacks` scheduled task, which runs every
-**30 seconds** by default. As a result `timeout` is a *minimum*: a process is failed on the first
-sweep at or after its deadline (so up to ~30 seconds late), and a `timeout` smaller than the sweep
-interval is effectively rounded up to the next sweep. Do not rely on sub-30-second precision.
-
-The task is registered like the other core tasks and is loaded with:
-
-```bash
-orchestrator-core scheduler load-initial-schedule
-```
-
-The interval is tunable through the scheduler API. The sweep does not start a run (and therefore
-creates no process record) when no process is awaiting a callback.
+Timeouts are enforced by the `task_validate_awaiting_callbacks` scheduled task, which sweeps every
+**30 seconds** by default. So `timeout` is a *minimum*: a process is failed on the first sweep at or
+after its deadline (up to ~30 seconds late) — don't rely on sub-30-second precision. Load the task
+with `orchestrator-core scheduler load-initial-schedule`; the interval can be changed to any value
+via the scheduler API (`PUT /api/schedules/`, which reschedules the job in place).
--- a/orchestrator/core/migrations/versions/schema/2026-06-29_f4a7c9e21b08_add_validate_awaiting_callbacks_task.py
+++ b/orchestrator/core/migrations/versions/schema/2026-06-29_f4a7c9e21b08_add_validate_awaiting_callbacks_task.py
@@ -19,10 +19,10 @@ Create Date: 2026-06-29 09:00:00.000000
 
 """
 
-from uuid import uuid4
-
 import sqlalchemy as sa
 from alembic import op
+
+from orchestrator.core.migrations.helpers import create_task
 
 # revision identifiers, used by Alembic.
 revision = "f4a7c9e21b08"
@@ -31,26 +31,17 @@ branch_labels = None
 depends_on = None
 
 
-workflow = {
+task = {
     "name": "task_validate_awaiting_callbacks",
-    "target": "SYSTEM",
     "description": "Fail callback steps that exceeded their timeout",
-    "workflow_id": uuid4(),
 }
 
 
 def upgrade() -> None:
     conn = op.get_bind()
-    # Named columns (not positional) so is_task is set explicitly; this is a task, not a process.
-    conn.execute(
-        sa.text(
-            "INSERT INTO workflows (workflow_id, name, target, description, is_task, created_at) "
-            "VALUES (:workflow_id, :name, :target, :description, true, now()) ON CONFLICT DO NOTHING"
-        ),
-        workflow,
-    )
+    create_task(conn, task)
 
 
 def downgrade() -> None:
     conn = op.get_bind()
-    conn.execute(sa.text("DELETE FROM workflows WHERE name = :name"), {"name": workflow["name"]})
+    conn.execute(sa.text("DELETE FROM workflows WHERE name = :name"), {"name": task["name"]})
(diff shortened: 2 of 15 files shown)


docs/reference-docs/workflows/callbacks.md
```diff
@@ -303,3 +303,58 @@ For long-running jobs, such as executing Terraform or Ansible playbooks, the Orc
...
+
+The deadline is measured from the moment the await started. Progress updates sent to
+`{callback_route}/progress` do **not** extend it.
+
+### Enforcement resolution
+
+Timeouts are enforced by the `task_validate_awaiting_callbacks` scheduled task, which runs every
+**30 seconds** by default. As a result `timeout` is a *minimum*: a process is failed on the first
+sweep at or after its deadline (so up to ~30 seconds late), and a `timeout` smaller than the sweep
+interval is effectively rounded up to the next sweep. Do not rely on sub-30-second precision.
```
kvklink: I find this section a bit confusing to read. Is it just "it happens every 30 seconds"? If so, then this whole block of text could be a lot shorter and less complex IMO.

orchestrator/core/migrations/versions/schema/2026-06-29_f4a7c9e21b08_add_validate_awaiting_callbacks_task.py
```diff
@@ -0,0 +1,56 @@
...
+def upgrade() -> None:
+    conn = op.get_bind()
+    # Named columns (not positional) so is_task is set explicitly; this is a task, not a process.
+    conn.execute(
+        sa.text(
+            "INSERT INTO workflows (workflow_id, name, target, description, is_task, created_at) "
+            "VALUES (:workflow_id, :name, :target, :description, true, now()) ON CONFLICT DO NOTHING"
+        ),
+        workflow,
+    )
```
kvklink: Was there a reason not to use the `add_workflow()` helper?

orchestrator/core/workflow.py
```diff
@@ -75,6 +75,7 @@
 DEFAULT_CALLBACK_ROUTE_KEY = "callback_route"
 CALLBACK_TOKEN_KEY = "__callback_token"  # noqa: S105
 DEFAULT_CALLBACK_PROGRESS_KEY = "callback_progress"  # noqa: S105
+CALLBACK_TIMEOUT_KEY = "__callback_timeout"  # seconds; stored in the await step state when a timeout is set
```
kvklink: I believe when formatted like one of these options, it will show up in the generated documentation. You will have to test this however to find out which one it is.

```suggestion
CALLBACK_TIMEOUT_KEY = "__callback_timeout"
"""Callback timeout in seconds, stored in the await step state when a timeout is set."""
```

```suggestion
#: Callback timeout in seconds, stored in the await step state when a timeout is set.
CALLBACK_TIMEOUT_KEY = "__callback_timeout"
```

orchestrator/core/workflow.py
```diff
@@ -463,11 +464,14 @@ def stepfunc(process_id: UUID) -> State:
...
+    extra_state = {
+        **({"__callback_result_key": result_key} if result_key else {}),
+        **({CALLBACK_TIMEOUT_KEY: timeout} if timeout is not None else {}),
+    }
+
     def await_(state: State) -> Process:
-        if result_key:
-            state = {**state, "__callback_result_key": result_key}
-        return AwaitingCallback(state)
+        return AwaitingCallback({**state, **extra_state})
```
Mark90: An alternative (and, subjectively, easier to debug) pattern we often use instead of this is to have a function yield the items of the dict. Also the `**` dict unwrapping is unnecessary

```suggestion
    def extra_state() -> Iterator[tuple[str, Any]]:
        if result_key:
            yield "__callback_result_key", result_key
        if timeout is not None:
            yield CALLBACK_TIMEOUT_KEY, timeout

    def await_(state: State) -> Process:
        return AwaitingCallback(state | dict(extra_state())})
```

orchestrator/core/workflows/predicates.py
```diff
@@ -46,3 +48,37 @@ def no_uncompleted_instance(context: PredicateContext) -> RunPredicateResult:
...
+
+
+def awaiting_callbacks_exist(context: PredicateContext) -> RunPredicateResult:
+    """Predicate that only runs the workflow when a process is awaiting a callback that has a timeout set.
+
+    Composes with :func:`no_uncompleted_instance` so the timeout sweep neither overlaps a previous run nor starts a run
+    (and thus creates a process row) when there is nothing to check. The timeout lives in the awaiting step's JSONB
+    state, so we join to the awaiting step and require the ``__callback_timeout`` key; callbacks without a timeout (the
+    default) never need the sweep. Joining on ``last_status`` excludes the lingering awaiting-callback step rows of
+    aborted/failed processes.
```
Mark90, line 60: Clever. With the 30s schedule interval this saves quite a few unnecessary runs.

test/integration_tests/workflows/test_async_workflow.py
```diff
@@ -342,3 +359,106 @@ def test_wf_callback_progress_with_multiple_callback_steps(test_client):
...
+    process_id = response.json()["id"]
+    assert test_client.get(f"api/processes/{process_id}").json()["last_status"] == "awaiting_callback"
+    return instance, process_id
+
+
+def test_callback_step_timeout_stored_in_state(test_client):
+    instance, process_id = _start_timeout_wf(test_client, timeout=300)
+    try:
+        process = db.session.get(ProcessTable, process_id)
+        assert process.steps[-1].state[CALLBACK_TIMEOUT_KEY] == 300
```
Mark90: To my taste this tests too much of the implementation details

test/integration_tests/workflows/test_async_workflow.py
```diff
@@ -342,3 +359,106 @@ def test_wf_callback_progress_with_multiple_callback_steps(test_client):
...
+    finally:
+        instance.__exit__(None, None, None)
+
+
+def test_callback_step_without_timeout_has_no_timeout_key(test_client):
+    instance, process_id = _start_timeout_wf(test_client, timeout=None)
+    try:
+        process = db.session.get(ProcessTable, process_id)
+        assert CALLBACK_TIMEOUT_KEY not in process.steps[-1].state
```
Mark90: Also too much implementation detail

test/integration_tests/workflows/test_predicates.py
```diff
@@ -57,3 +59,64 @@ def test_no_uncompleted_by_process_last_status(test_process, test_workflow, proc
...
+    db.session.commit()
+    return process
+
+
+def test_awaiting_callbacks_exist_pass(test_workflow, generic_subscription_1) -> None:
+    _awaiting_process(test_workflow, with_timeout=True)
+
+    # workflow_key with no uncompleted instances of itself, so only the awaiting-existence check matters
+    context = MagicMock(spec=PredicateContext)
+    context.workflow_key = "task_validate_awaiting_callbacks"
```
Mark90: Would prefer not to have mocks in integration tests. Let's either use PredicateContext and test the whole integration, or also mock the DB access and make it a unit test

orchestrator/core/workflows/tasks/validate_awaiting_callbacks.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+@step("Find timed-out awaiting-callback processes")
+def find_timed_out_callbacks(process_id: UUID) -> State:
+    now = nowtz()
+    awaiting_processes = db.session.scalars(
+        select(ProcessTable).filter(
+            ProcessTable.last_status == ProcessStatus.AWAITING_CALLBACK,
+            ProcessTable.process_id != process_id,
+        )
```
Mark90: ProcessTable.steps could/should be eagerloaded here to prevent unnecessary N+1 queries

Mark90 (changes requested):
Looking good already, some comments

orchestrator/core/workflow.py
Mark90: It's an implementation detail though, do we even want to have it appear in the docs?

kvklink: Good point, it doesn't add anything useful to the docs. Might as well remove it

docs/reference-docs/workflows/callbacks.md
```diff
@@ -303,3 +303,50 @@ For long-running jobs, such as executing Terraform or Ansible playbooks, the Orc
...
+The deadline is measured from the moment the await started. Progress updates sent to
+`{callback_route}/progress` do **not** extend it.
+
+### Enforcement resolution
+
+Timeouts are enforced by the `task_validate_awaiting_callbacks` scheduled task, which sweeps every
+**30 seconds** by default. So `timeout` is a *minimum*: a process is failed on the first sweep at or
+after its deadline (up to ~30 seconds late) — don't rely on sub-30-second precision. Load the task
+with `orchestrator-core scheduler load-initial-schedule`; the interval can be changed to any value
+via the scheduler API (`PUT /api/schedules/`, which reschedules the job in place).
```
Mark90, line 352: > via the scheduler API (`PUT /api/schedules/`, which reschedules the job in place).

Or via the scheduled tasks page in the metadata tab

Mark90 (approved):

docs/reference-docs/workflows/callbacks.md
hanstrompert: That is true, but the dropdown with available intervals is very coarse, the smallest is one hour, which is a bit long for this feature, making the schedules tab not really on option. Besides that, in the tab you cannot modify a schedule, only delete and re-create. I can mention all this in the docs. Or I can create an issue to make the schedules tab more flexible.

Mark90: All good points. We could make an issue for the things you describe as they will make editing the schedules more user friendly.

orchestrator/core/migrations/versions/schema/2026-06-29_f4a7c9e21b08_add_validate_awaiting_callbacks_task.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+    create_task(conn, task)
+
+
+def downgrade() -> None:
+    conn = op.get_bind()
+    conn.execute(sa.text("DELETE FROM workflows WHERE name = :name"), {"name": task["name"]})
```
kvklink:
```suggestion
    delete_task(conn, task["name"])
```

🙏🏼

docs/reference-docs/workflows/callbacks.md
hanstrompert: I created issue https://github.com/workfloworchestrator/orchestrator-ui-library

kvklink (approved):

PR #1746, opened 2026-06-29
Add docs, small test cases

### LINKED ISSUE #857: [Feature]: Document conditional steps
### Contact Details

_No response_

### What should we build?

Create documentation on how to use `conditional`. 

Question thread on discord: https://discord.com/channels/1295834294270558280/1343606814868111371

Small example for a usecase where some steps should only be run when the user input contained `skip_config=True`:

```python
from orchestrator.workflow import conditional

@terminate_workflow("Terminate L2vpnPP", initial_input_form=terminate_initial_input_form_generator)
def terminate_l2vpn_pp() -> StepList:

    def skip_config() -> Callable:
        
        def skip_config_is_true(state: dict) -> bool:
            return state['skip_config'] is True
        
        return conditional(skip_config_is_true)
    
    return (
        begin
        >> skip_config()(
            begin
            >> lso.indifferent_pp_interaction(call_ansible_playbook_dryrun)
            >> lso.indifferent_pp_interaction(call_ansible_playbook_live)
        )
        >> ims_remove_l2vpn
        >> ims_remove_subints
    )
```

### Relevant pseudo code

```shell

```

Sparrow1029:
## Summary

This PR adds some more documentation around `conditional` steps + some usage examples.

I learned doing this work that `conditional`, when passed a `StepList` will use `.map` to wrap the conditional around all contained steps. I saw there wasn't test cases related to this specific functionality, so thought it prudent to add a couple.

## Related Issues
Resolves #857 

## Checklist

### DIFF
--- a/docs/reference-docs/workflows/conditional-steps.md
+++ b/docs/reference-docs/workflows/conditional-steps.md
@@ -31,17 +31,34 @@ the predicate result, later steps in the same group may behave differently.
 
     ```python
     from orchestrator.core.workflow import StepList, begin, conditional
+    from pydantic_forms.types import FormGenerator, UUIDstr
 
 
-    if_has_nodes = conditional(lambda state: len(state["nodes_to_deploy"]) > 0)
+    def initial_input_form_generator(subscription_id: UUIDstr) -> FormGenerator:
+        class ModifyDnsForm(FormPage):
+            description: str
+            ttl: int = 3600
+            new_cname: str | None = None  # leave empty to keep current CNAME
+
+        user_input = yield ModifyDnsForm
+        return user_input.model_dump()
+
+    # ... step definitions ...
+
+    if_cname_changed = conditional(lambda state: state.get("new_cname") is not None)
 
 
-    @modify_workflow("Add Nodes", initial_input_form=initial_input_form_generator)
-    def add_nodes() -> StepList:
+    @modify_workflow(
+        "Modify DNS Record",
+        initial_input_form=initial_input_form_generator,
+    )
+    def modify_dns_record() -> StepList:
         return (
             begin
-            >> if_has_nodes(deploy_nodes)
-            >> finalize
+            >> update_subscription
+            >> if_cname_changed(notify_customer)
+            >> update_dns_in_ipam
+            >> set_status_provisioning
         )
     ```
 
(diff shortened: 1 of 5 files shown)


docs/reference-docs/workflows/conditional-steps.md
```diff
@@ -0,0 +1,216 @@
...
+        return (
+            begin
+            >> check_deployment_status
+            >> assemble_payload
+            >> _if_deployed(remove_legacy_config)
+            >> deploy_changes
+        )
+    ```
+
+### Reusable conditional sub-pipelines
```
kvklink: I agree with the fact that conditionals can be re-used in different workflows, but I don't quite see how the example shows this.

docs/reference-docs/workflows/conditional-steps.md
```diff
@@ -0,0 +1,216 @@
...
+
+
+    @modify_workflow("Add Nodes", initial_input_form=initial_input_form_generator)
+    def add_nodes() -> StepList:
+        return (
+            begin
+            >> if_has_nodes(deploy_nodes)
+            >> finalize
+        )
+    ```
```
Mark90, line 63: Would be nice to have a step prior to the conditional step that inserts the state value on which the conditional hinges. Or was the intent to provide this value from the form inputs? If so, would be nice to include (part of) the form generator :)

docs/reference-docs/workflows/conditional-steps.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Wrapping multiple steps
+
+Pass a `StepList` (built with `begin >> ...`) to skip an entire group of steps
+under a single condition:
```
Mark90: This description is not complete; as I understand it (and to be honest as I just learned today) the conditional is evaluated for each wrapped step. So it can indeed skip all steps as described here, but also only a subset of the steplist

test/integration_tests/workflows/test_workflow.py
```diff
@@ -443,6 +443,72 @@ def inc_n(n=0):
...
+    assert_complete(result)
+    assert_state(result, {"enable": True, "counter": 2})
+    assert len([x for x in log if x[1].isskipped()]) == 0, "no steps should be skipped"
+
+
+def test_conditional_group_predicate_reevaluated_between_steps():
+    """Predicate is evaluated independently for each step in the group.
+
+    A step *inside* the group that mutates state can change the predicate
+    result for subsequent steps in the same group.
```
Mark90, line 482: Surprising, did not know this! Good to have it tested.

Mark90 (approved):

docs/reference-docs/workflows/conditional-steps.md
Sparrow1029: Changed to this:
```diff
### Wrapping multiple steps

- Pass a `StepList` (built with `begin >> ...`) to skip an entire group of steps
- under a single condition:
+ Pass a `StepList` (built with `begin >> ...`) to guard multiple steps with the
+ same predicate. The predicate is re-evaluated for **each** step individually,
+ so if a step **within** the `StepList` modifies the state, later steps in the
+ group may be skipped or executed differently:
```

This is making me wonder what happens if you wrap a `@step_group` with a `conditional`... 🤔

docs/reference-docs/workflows/conditional-steps.md
Sparrow1029: I took your suggestion and added a form generator with a contrived example. Good call out

docs/reference-docs/workflows/conditional-steps.md
Sparrow1029: You're right, this was kind of showing how to re-use a `StepList` across workflows... Instead I changed to show how to use it as `@conditional` decorator and re-use _that_ function across a couple workflows (the new example is basically from ESnet code)

Mark90 (approved):

docs/reference-docs/workflows/conditional-steps.md
Mark90: We'll leave that as an exercise to the readers :)

kvklink (approved):

PR #1753, opened 2026-07-02
1284 persist name of user triggering task create/resume

### LINKED ISSUE #1284: [Feature]: store user performing the resume action
### Contact Details

_No response_

### What should we build?

See ticket https://github.com/workfloworchestrator/orchestrator-ui-library/issues/2182 where the frontend implementation is described. This will require backend work.

- [ ] use existing created_by field in `ProcessStepsTable` to store user
- [ ] update `create_process()` to set pstat.current_user to the user identified by user_model, if present, else set to the default `SYSTEM_USER`
- [ ] update `load_process()` to return the `current_user` from the database
- [ ] In `_get_current_step_to_update()`  check that the user is the same, if not create a new process step
- [ ] expand the unit tests for the modified functions
- [ ] add or expand unit tests to ensure the correct user is returned through the API on retries by different users

gertvdw:
## Summary
When using the worker as task executor (versus threadpool), there is no actual user triggering the task since Celery doesn't know about it. This information is provided with the task however, just not persisted.

In `start_process` and `resume_process` we update the `current_user` in ProcessState with the `user` parameter. This `user` is the actual user that created the task/ran the workflow/retried the workflow.
The `created_by` column in `process_steps` is now filled with this user instead of the default `SYSTEM`.

The method `_get_current_step_to_update` now also takes this `created_by` value into account in the decision to update the last step (same user) or record a new step (another user retried the step).

## Related Issues
Fixes #1284

## Checklist

### DIFF
--- a/orchestrator/core/services/executors/threadpool.py
+++ b/orchestrator/core/services/executors/threadpool.py
@@ -18,6 +18,7 @@ import structlog
 from sqlalchemy import select
 
 from oauth2_lib.fastapi import OIDCUserModel
+from orchestrator.core import app_settings
 from orchestrator.core.db import ProcessTable, db
 from orchestrator.core.db.database import transactional
 from orchestrator.core.services.executors.types import ExecutorFunction
@@ -32,6 +33,7 @@ from orchestrator.core.services.processes import (
     load_process,
     safe_logstep,
 )
+from orchestrator.core.settings import ExecutorType
 from orchestrator.core.types import BroadcastFunc
 from orchestrator.core.workflow import (
     ProcessStat,
@@ -95,6 +97,14 @@ def thread_start_process(
 
     # Trigger the task in the current thread or threadpool (depends on executor mode).
     pstat.update(state=pstat.state.map(lambda state: StateMerger.merge(state, input_data.input_state)))
+
+    # When using celery, the current_user prop is set to SYSTEM, because there is no actual user involved.
+    # The `user` property is supplied via the task and is the name of the user that triggered this task.
+    # When using the WORKER type, override `current_user` with this value so we have it available in the
+    # `created_by` property of a step.
+    if app_settings.EXECUTOR == ExecutorType.WORKER and user != SYSTEM_USER:
+        pstat.current_user = user
+
     _safe_logstep_with_func = partial(safe_logstep, broadcast_func=broadcast_func)
     return _run_process_async(pstat.process_id, lambda: runwf(pstat, _safe_logstep_with_func), broadcast_func=broadcast_func)
 
--- a/orchestrator/core/services/executors/threadpool.py
+++ b/orchestrator/core/services/executors/threadpool.py
@@ -18,7 +18,6 @@ import structlog
 from sqlalchemy import select
 
 from oauth2_lib.fastapi import OIDCUserModel
-from orchestrator.core import app_settings
 from orchestrator.core.db import ProcessTable, db
 from orchestrator.core.db.database import transactional
 from orchestrator.core.services.executors.types import ExecutorFunction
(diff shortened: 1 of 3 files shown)


test/integration_tests/services/executors/test_threadpool.py
```diff
@@ -164,7 +164,7 @@ def test_thread_resume_process_resumed(
     pstat.user = expected_user
     mock_set_process_status_running.assert_called_once()
     mock_retrieve_input_state.assert_called_once_with(pstat.process_id, "user_input", False)
-    assert pstat.update.call_args_list == [call(current_user="other user"), call(state={"state": "test"})]
+    assert pstat.update.call_args_list == [call(state={"state": "test"})]
     mock_run_process_async.assert_called_once()
     assert result == process_id
```
Mark90, line 235: Since this broke once we should have a regression test to cover the functionality.

It will have to be an integration test that sets up and runs a workflow until a step that always fails once, and then the test can call the resume endpoint. There are integration tests which set up workflows. The `Depends(user_name)` return value can be controlled by [dependency overrides](https://fastapi.tiangolo.com/advanced/testing-dependencies/).

orchestrator/core/services/executors/threadpool.py
```diff
@@ -95,6 +97,14 @@ def thread_start_process(
 
     # Trigger the task in the current thread or threadpool (depends on executor mode).
     pstat.update(state=pstat.state.map(lambda state: StateMerger.merge(state, input_data.input_state)))
+
+    # When using celery, the current_user prop is set to SYSTEM, because there is no actual user involved.
+    # The `user` property is supplied via the task and is the name of the user that triggered this task.
+    # When using the WORKER type, override `current_user` with this value so we have it available in the
+    # `created_by` property of a step.
+    if app_settings.EXECUTOR == ExecutorType.WORKER and user != SYSTEM_USER:
```
Mark90: Is the `app_settings.EXECUTOR == ExecutorType.WORKER` condition necessary?

gertvdw: I may have been overly cautious :) Removed the condition entirely

Mark90 (approved):
LGTM!

orchestrator/core/services/executors/threadpool.py
```diff
@@ -123,7 +130,9 @@ def thread_resume_process(
     input_data = retrieve_input_state(process.process_id, retrieve_input_str, False)
 
     if user:
-        pstat.update(current_user=user)
```
tjeerddie, line 126: if this doesn't work, is the line below (L136) this also useless?

Mark90: Good point. Upon taking a closer look, `ProcessStat.update` creates a new ProcessStat instance with the updated variables and leaves the old instance unaffected:

```py
>>> 
>>> stat.update(state={'foo': 'bar'})
ProcessStat(process_id=UUID('54d9ec43-c6bc-40fe-bb4b-a1f3e107716d'), workflow='wf', state={'foo': 'bar'}, log=[], current_user='SYSTEM', user_model=None)
>>> stat.update(current_user='john')
ProcessStat(process_id=UUID('54d9ec43-c6bc-40fe-bb4b-a1f3e107716d'), workflow='wf', state={}, log=[], current_user='john', user_model=None)
>>> stat
ProcessStat(process_id=UUID('54d9ec43-c6bc-40fe-bb4b-a1f3e107716d'), workflow='wf', state={}, log=[], current_user='SYSTEM', user_model=None)
>>> 
>>> stat = stat.update(current_user='john', state={'foo': 'bar'})
>>> stat
ProcessStat(process_id=UUID('54d9ec43-c6bc-40fe-bb4b-a1f3e107716d'), workflow='wf', state={'foo': 'bar'}, log=[], current_user='john', user_model=None)
>>> 
```


I think it'd be nicer to revert to the `pstat.update()` syntax but then change it to actually do `pstat = pstat.update()`. 

Dear me: we have quite a few of these cases...

Image

Mark90: Will create a separate ticket for that

PR #1755, opened 2026-07-02
Fix cron schedule parsing (wildcards dropped → every second, valid fields rejected, 6 fields → 500)

### LINKED ISSUE #1754: Scheduler cron form mishandles standard cron expressions (`*` dropped → every second, valid fields rejected, 6 fields → 500)
Three defects in the CRON path of `orchestrator/core/forms/scheduler_form.py`:

**1. `*` (and ranges/lists/steps) are turned into `None` → schedule fires every second.**
`parse_cron_field` does `int(value)` and returns `None` on failure, so `*`, `*/5`, `1-5`, `1,3` all become `None`. For `* * * * *` every field is `None`:
```python
>>> str(CronTrigger(minute=None, hour=None, day=None, month=None, day_of_week=None))
'cron[]'   # no constraints → fires every second, not every minute
```

**2. A 6-field expression returns HTTP 500.**
`cron_regex` accepts 5–8 fields, but `get_cron_kwargs` hard-unpacks 5: `minute, hour, day, month, day_of_week = form_data["cron"].split(" ")`. Input `* * * * * *` → `ValueError: too many values to unpack (expected 5)` → unhandled → 500.

**3. `cron_regex` rejects valid standard cron fields.**
The first four fields only allow `*` or a single number (`[0-5]?\d`) with optional `/step`; ranges and lists are rejected there, so `0-59 * * * *`, `1,15,30 * * * *`, `0 9-17 * * 1-5` all fail validation. Since this regex is the field's `StringConstraints`, it also blocks these in the UI. It also permits Quartz-only tokens (`? L W #`) APScheduler doesn't support.

**Docs:** `docs/reference-docs/workflows/callbacks.md` tells users to change the callback-timeout sweep interval via `PUT /api/schedules/`; with the fix this can be done from the UI as a cron schedule (including seconds), so the doc should be updated to the delete-and-recreate-cron flow.

hanstrompert:
Closes #1754.

The CRON path in `scheduler_form.py` had three defects:
- `parse_cron_field` int-coerced fields, turning `*`, `*/n`, ranges and lists into `None` → an unconstrained `CronTrigger` firing **every second**.
- a 6-field expression passed the regex but crashed the 5-way unpack → **500**.
- the regex rejected valid `-`/`,` fields (e.g. `0-59 * * * *`).

Now the field strings are passed straight to `CronTrigger` (which accepts `*`, `*/n`, ranges, lists), and the form accepts exactly **5 fields** (`minute…day_of_week`) or **6** (leading `second`), rejecting other counts. Also updates the callback-timeout doc to the delete-and-recreate-cron flow.

Verified via `test_get_cron_kwargs`: `* * * * *` → every minute; `0-59 * * * *`, `1,15,30 * * * *`, `0 9-17 * * 1-5` accepted; `*/15 * * * * *` → every 15s; 4-/7-field input rejected.

### DIFF
--- a/orchestrator/core/forms/scheduler_form.py
+++ b/orchestrator/core/forms/scheduler_form.py
@@ -72,31 +72,26 @@ INTERVAL_MAPPING = {
     Intervals.ONE_MONTH: {"weeks": 4},
 }
 
-cron_regex = r"^(?:(\*|([0-5]?\d))(?:/(\d+))?\s+){4}(?:(\*|([0-5]?\d))(?:/(\d+))?\s+)?(?:([0-9,/*\-?LW#]+)(?:\s+([0-9,/*\-?LW#]+))?(?:\s+([0-9,/*\-?LW#]+))?)$"
+# 5 fields (minute hour day month day_of_week) or 6 with a leading second; each field allows the
+# cron characters (digits, *, /, -, ,, and names). CronTrigger validates the field values.
+cron_regex = r"^([\w*,\-/]+)(\s+[\w*,\-/]+){4,5}$"
 
 
 def get_interval_kwargs(form_data: dict) -> dict:
     return {"start_date": form_data["start_date"]} | INTERVAL_MAPPING[form_data["interval"]]
 
 
-def parse_cron_field(value: str) -> int | None:
-    try:
-        return int(value)
-    except ValueError:
-        return None
-
-
 def get_cron_kwargs(form_data: dict) -> dict:
-    minute, hour, day, month, day_of_week = form_data["cron"].split(" ")
+    # Pass the raw field strings to CronTrigger, which accepts *, */n, ranges and lists. A 5-field
+    # expression is minute..day_of_week; a 6-field one adds a leading second.
+    names = ["minute", "hour", "day", "month", "day_of_week"]
+    fields = form_data["cron"].split()
+    if len(fields) == 6:
+        names = ["second", *names]
+    elif len(fields) != 5:
+        raise ValueError("A cron schedule needs 5 fields (minute hour day month day_of_week) or 6 with a leading second")
 
-    return {
-        "start_date": form_data["start_date"],
-        "minute": parse_cron_field(minute),
-        "hour": parse_cron_field(hour),
-        "day": parse_cron_field(day),
-        "month": parse_cron_field(month),
-        "day_of_week": parse_cron_field(day_of_week),
-    }
+    return {"start_date": form_data["start_date"]} | dict(zip(names, fields, strict=True))
 
 
 async def _check_authorize(workflow: Workflow, user_model: OIDCUserModel | None) -> bool:
--- a/orchestrator/core/forms/scheduler_form.py
+++ b/orchestrator/core/forms/scheduler_form.py
@@ -17,7 +17,8 @@ from typing import Annotated
 from uuid import UUID
 
 import structlog
-from pydantic import ConfigDict, StringConstraints
+from apscheduler.triggers.cron import CronTrigger
+from pydantic import AfterValidator, ConfigDict
 from sqlalchemy import select
 from typing_extensions import TypedDict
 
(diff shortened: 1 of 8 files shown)


tjeerddie (approved):

orchestrator/core/forms/scheduler_form.py
```diff
@@ -72,31 +72,26 @@ class Buttons(TypedDict):
...
-        "day_of_week": parse_cron_field(day_of_week),
-    }
+    # Pass the raw field strings to CronTrigger, which accepts *, */n, ranges and lists. A 5-field
+    # expression is minute..day_of_week; a 6-field one adds a leading second.
+    names = ["minute", "hour", "day", "month", "day_of_week"]
+    fields = form_data["cron"].split()
+    if len(fields) == 6:
+        names = ["second", *names]
+    elif len(fields) != 5:
+        raise ValueError("A cron schedule needs 5 fields (minute hour day month day_of_week) or 6 with a leading second")
```
Mark90: Seems like a job for match/case. And a little bit of duplication here is ok to prevent having to use list slicing.

```suggestion
    match (fields := form_data["cron"].split()):
        case _ if len(fields) == 5:
            names = ["minute", "hour", "day", "month", "day_of_week"]
        case _ if len(fields) == 6:
            names = ["second", "minute", "hour", "day", "month", "day_of_week"]
        case _:
            raise ValueError("A cron schedule needs 5 fields (minute hour day month day_of_week) or 6 with a leading second")
```

orchestrator/core/forms/scheduler_form.py
```diff
@@ -72,31 +72,26 @@ class Buttons(TypedDict):
     Intervals.ONE_MONTH: {"weeks": 4},
 }
 
-cron_regex = r"^(?:(\*|([0-5]?\d))(?:/(\d+))?\s+){4}(?:(\*|([0-5]?\d))(?:/(\d+))?\s+)?(?:([0-9,/*\-?LW#]+)(?:\s+([0-9,/*\-?LW#]+))?(?:\s+([0-9,/*\-?LW#]+))?)$"
+# 5 fields (minute hour day month day_of_week) or 6 with a leading second; each field allows the
+# cron characters (digits, *, /, -, ,, and names). CronTrigger validates the field values.
+cron_regex = r"^([\w*,\-/]+)(\s+[\w*,\-/]+){4,5}$"
```
Mark90: The old regex used to prevent more bad inputs than this one. 
Like `61 * * * *`
Where are we catching these now?

orchestrator/core/forms/scheduler_form.py
Sparrow1029: ^Would this look a little clearer as

```python
fields = form_data["cron"].split()
match len(fields):
    case 5:
        names = [...]
    case 6:
        names = [...]
    case _:
        raise ValueError(...)
```
?

Also, TIL that you can add a guard to a default case (and have multiple default cases with different guards)

hanstrompert: I agree, using @Sparrow1029's optimised version of @Mark90's match.

orchestrator/core/forms/scheduler_form.py
hanstrompert: Good point. The original regex did prevent more bad inputs, but also rejected valid inputs, one of the reasons to open this issue. Now that both 5 and 6 field cron expressions are supported, it is almost impossible to create a regex to fully express the syntax, and it would be uncomfortably long, if at all possible. But we do want solid validation on the input form, and on the API as well, the latter was still missing in this PR. Now using the apscheduler triggers to do the validation for us. This way we loose frontend validation, but gain watertight backend validation for both the input form as the create and update API endpoints.

orchestrator/core/forms/scheduler_form.py
Mark90: Yeah, Alex has outdone me here. :)

orchestrator/core/forms/scheduler_form.py
Mark90: Fair enough!

Mark90 (approved):

PR #1766, opened 2026-07-07
Add support for type conversion of SubscriptionModel unions in step params

### LINKED ISSUE #1758: [Bug]: Step parameter type conversion of Subscription Models does not work with unions
When a step argument is typed as subscription: `ProvisioningSub | ActiveSub`, the type-conversion layer fails to resolve the union to either concrete model and leaves subscription as a plain `dict`.

This breaks the "strict" lifecycle pattern, where a step is intentionally written to be reusable across both the provisioning and active lifecycle stages via a union annotation — e.g.:

```
def example_step(subscription: ProvisioningSub | ActiveSub) -> State:
    assert type(subscription) == dict  # expected ProvisioningSub or ActiveSub instance
```

Impact: any step relying on attribute access on subscription (rather than dict indexing) raises AttributeError at runtime instead of failing type-checking, since mypy sees the annotated union type but the runtime value is an unconverted dict.

Expected: the conversion logic should detect Union/| annotations, attempt each member type in order (or match on a discriminator), and instantiate the matching concrete model — matching the behavior already in place for single concrete-type annotations.

---

Affected version: 5.1.0

tjeerddie:
Closes: https://github.com/workfloworchestrator/orchestrator-core/issues/1758

### DIFF
--- a/orchestrator/core/utils/state.py
+++ b/orchestrator/core/utils/state.py
@@ -267,24 +267,43 @@ def _handle_optional_subscription_model_param(
     return None
 
 
-def _handle_union_subscription_model_param(param_name: str, state: State) -> SubscriptionModel:
+def _handle_union_subscription_model_param(param_name: str, annotation: Any, state: State) -> SubscriptionModel:
     """Load a SubscriptionModel from database for a union type annotation (Node | NodeProvisioning).
 
     Uses SubscriptionModel.from_subscription to load the type based on the
-    subscription's lifecycle, so the correct subclass is returned regardless of which
-    type of the union the subscription currently is.
+    subscription's lifecycle, which then gets verified against the types declared in the union annotation.
 
     Args:
         param_name: Parameter name
+        annotation: The union type annotation (e.g. ProductActive | ProductProvisioning)
         state: Workflow state
 
     Returns:
         Loaded and validated SubscriptionModel instance
 
     Raises:
+        TypeError: If the union's types are not all lifecycle variants of the same base SubscriptionModel
         KeyError: If subscription_id not found in state
+        ValueError: If the loaded model's type does not match any type in the union
     """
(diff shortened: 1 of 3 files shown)


pboers1988 (approved):

orchestrator/core/utils/state.py
```diff
@@ -255,6 +267,26 @@ def _handle_optional_subscription_model_param(
     return None
 
 
+def _handle_union_subscription_model_param(param_name: str, state: State) -> SubscriptionModel:
+    """Load a SubscriptionModel from database for a union type annotation (Node | NodeProvisioning).
+
+    Uses SubscriptionModel.from_subscription to load the type based on the
+    subscription's lifecycle, so the correct subclass is returned regardless of which
+    type of the union the subscription currently is.
```
Mark90: Seems like a bit of a hack; if someone is not using a type checker then they can accidentally fill in `subscription: ProductX | ProductXProvisioning` and get back a `ProductY`.

We should 
1) validate that all models of the union are of the same base type
2) ensure that the returned SubscriptionModel is one of those types

Mark90 (changes requested):

orchestrator/core/utils/state.py
pboers1988: Pydantic has a whole doc about unions: https://pydantic.dev/docs/validation/latest/concepts/unions/ when we implemented Unions as a descriminator in the Orchestrator you had to define your most explicit model first in the union to make sure things worked and got serialised correctly.

tjeerddie: its not necessary anymore to have your most explicit model first, pydantic defaults to `smart` mode https://pydantic.dev/docs/validation/latest/concepts/unions/#smart-mode

PR #1774, opened 2026-07-08
Use full path in structured queries and allow for multiple matching fields

tvdven:
## Summary

Structured (filter-only) search returned the filter's input as `matching_field.path`, so filtering on a global (dotless) field like `status` reported `"path": "status"` instead of the full indexed path. The matching field was fabricated from the filter input in `format_search_response`, which has no DB access — while the actual matched path is per-entity data (the same `status=active` filter can match `subscription.status` for one subscription and `subscription.product.status` for another).

**Changes:**

- **`matching_field` → `matching_fields`** (breaking): the response schema now returns a list of matching fields instead of a single optional one, in both the REST/GraphQL APIs and the `SearchResult` model. This lets multi-filter structured queries report every matched field, not just the first.
- `StructuredRetriever` now receives the query's filters and emits a single `highlight_matches` JSON column — a JSON array of arrays, one inner array per positive filter leaf, each containing all index rows that matched that leaf (`{"value", "path", "idx"}`). This is a new column distinct from the `highlight_text`/`highlight_path` columns used by the fuzzy/semantic retrievers.
- Extracted `PathFilter.matched_row_predicate()` so the highlight subqueries reuse the exact row predicate used for filtering (dotless path → `*.` lquery, full path → exact match), preventing the two from drifting apart.
- `format_search_response` reads the `highlight_matches` JSON column from the DB row and expands it into a list of `MatchingField` entries; absence filters (`not_has_component`) are excluded since they match by the *absence* of an index row.
- When multiple paths match for one entity+filter, all unique `(text, path)` combinations are returned — every matched path is included in `matching_fields`.
- The `StructuredRetriever` now handles `order_by` with proper keyset pagination (`order_value` column, entity_id tiebreaker), fixing cursor-based paging for sorted structured queries.

### Example:

**Frontend:**
Shows Search-POC page with 'Terminated' subscriptions and filter `status == "active"` - This shows a single subscription which `subscription.status` is terminated but the `product.status` is still active:
image

**Backend API response:**

```json
{
    "matching_fields": [
        {
            "text": "terminated",
            "path": "subscription.status",
            "highlight_indices": [[0, 10]]
        },
        {
            "text": "active",
            "path": "subscription.product.status",
            "highlight_indices": [[0, 6]]
        }
    ]
}
```

❗️This PR is a breaking change requires updates in the `orchestrator-ui` / `orchestrator-ui-library`:
- API no longer determines `matching_field` but `matching_fields` instead containing multiple results
- For the different tabs `Active`, `Terminated`, `Transient`, `All` on the 'Search POC` page this means that there needs some logic to determine which highlight should be shown. E.g. in the example above where the 'status == "active"' filter is applied, the second list item (with path `subscripiton.product.status` should be shown as highlight to indicate where it matched on status active.

### Tests
Covered by new integration tests (global path, per-entity resolution, explicit full path, `ends_with`, multi-leaf) and unit tests for the response formatting.

**Related issues**:
- https://git.ia.surf.nl/netdev/automation/projects/orchestrator/-/work_items/2838
- https://git.ia.surf.nl/netdev/automation/projects/orchestrator/-/work_items/2839


orchestrator/core/search/query/results.py
```diff
@@ -312,28 +307,34 @@ def format_search_response(
...
+        leaf
+        for leaf in filters.get_all_leaves()
+        if not (isinstance(leaf.condition, LtreeFilter) and leaf.condition.op == FilterOp.NOT_HAS_COMPONENT)
+    ]
+
+    flat_matches = [m for inner in leaf_arrays if inner for m in inner]
+    unique_matches = {(str(m["value"]), str(m["path"])): m for m in flat_matches}
+
+    results: list[MatchingField] = []
+    for (text, path), m in unique_matches.items():
```
tjeerddie: use list comprehension with a function call inside?

orchestrator/core/graphql/resolvers/search.py
```diff
@@ -86,7 +86,7 @@ def _result_to_gql(result: SearchResult) -> SearchResultType:
         entity_title=result.entity_title,
         score=result.score,
         perfect_match=result.perfect_match,
-        matching_field=_matching_field_to_gql(result.matching_field) if result.matching_field else None,
```
tjeerddie, line 89: you could keep this property and give it the first `matching_fields` entry and a deprecated comment to be removed later, so its not a breaking change

test/integration_tests/search/retrieval/test_structured_matching_field.py
```diff
@@ -0,0 +1,165 @@
...
+    assert set(matching_by_id) == {str(sub_a), str(sub_b)}
+    assert len(matching_by_id[str(sub_a)]) == 1
+    assert matching_by_id[str(sub_a)][0].path == "subscription.status"
+    assert matching_by_id[str(sub_a)][0].text == "active"
+    assert len(matching_by_id[str(sub_b)]) == 1
+    assert matching_by_id[str(sub_b)][0].path == "subscription.block.status"
+    assert matching_by_id[str(sub_b)][0].text == "provisioning"
+
+
+async def test_multi_leaf_filter_returns_all_matching_fields(indexed_subscriptions):
```
tjeerddie, line 140: what about more complicated filters, using `or` and `not equal`?

orchestrator/core/graphql/resolvers/search.py
tvdven: Discussed this with Wouter. It's not an issue if the highlight bar will not be displayed currently

orchestrator/core/search/query/results.py
tvdven: Good one

test/integration_tests/search/retrieval/test_structured_matching_field.py
tvdven: Included test to cover those as well

PR #1795, opened 2026-07-21
Fix schedule creation redirect

### LINKED ISSUE #1794: [Bug]: Schedule task form in v5.1.0 giving error on create
### Contact Details

massimo.simone@garr.it

### What happened?

Hello i'm trying the new schedule form task, but in the end, after i select all the required info, when i click "Create Schedule", it gives me multiple errors. By using Burp Suite I've noticed that on create schedule, it posts to /api/forms/configure_schedule and I get a 200, but after it does another post to /api/schedules that gives a HTTP/2 307 Temporary Redirect leading to an unsucceful creation of the task. I've noticed by using the repeater function of Burp that instead of posting to  /api/schedules to  /api/schedules/ it returns me a 200 and it adds correctly to queue.

https://discordapp.com/channels/1295834294270558280/1526210673950855228 here you can see the picture of the troubleshooting

### Version

5.1.0

Python 3.12

### Relevant log output

```shell

```

Sanjays2402:
## Summary
Register the documented `POST /api/schedules` URL directly so schedule creation returns 201 instead of a 307 redirect. The trailing-slash route remains available for compatibility.

## Related Issues
Fixes #1794

## Checklist

### DIFF
--- a/orchestrator/core/api/api_v1/endpoints/schedules.py
+++ b/orchestrator/core/api/api_v1/endpoints/schedules.py
@@ -40,7 +40,8 @@ async def validate_schedule_authorization(task: Workflow | None, user_model: OID
         raise_status(HTTPStatus.FORBIDDEN, f"User is not authorized to manage schedule with '{task.name}' task")
 
 
-@router.post("/", status_code=HTTPStatus.CREATED)
+@router.post("/", status_code=HTTPStatus.CREATED, include_in_schema=False)
+@router.post("", status_code=HTTPStatus.CREATED)
 async def create_scheduled_task(
     payload: APSchedulerJobCreate, user_model: OIDCUserModel | None = Depends(authenticate)
 ) -> dict[str, str]:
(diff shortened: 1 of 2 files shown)


tjeerddie (approved):

tjeerddie (changes requested):
accepted it too casually, Its weird to specifically add this only for this endpoint, whilst all endpoints expect / suffixed.
I don't think this is a problem of orchestrator-core but the calling side. all our endpoints use the slash and it should correctly redirect to `/api/schedules/` and it depends on the client to correctly handle the redirect.

Mark90:
Hi Sanjay, 
Thanks for the contribution. However, ending URLs with a trailing slash is -at least for the "base" routes like /api/products/ etc- the common practice in the API. So the correct thing to do is to ensure callers (like the UI) follow this same pattern. Left a comment on the ticket #1794 about this as well and created a ticket to address this in the UI. 
Closing this.

PR #1798, opened 2026-07-22
search: Add schema-aware type inference for ambiguous ES DSL terms (i.e. digits)

florisie:
## Summary

Preserve the intended type of digit-only Elasticsearch DSL term values for subscription fields.

Previously, a value such as `"26"` was inferred as numeric solely because it contains digits. This is incorrect for string-backed subscription fields, such as a VLAN range, and can cause filters to return no matching subscriptions.

## Changes

- Accept `FilterTree` and validated Elasticsearch DSL filters in `SearchRequest`.
- Convert Elasticsearch DSL to `FilterTree` in `SearchRequest.to_query(entity_type)`, where the entity type is available.
- Derive and cache searchable field types from registered subscription models and their lifecycle variants.
- Resolve digit-only term values from the schema when the field type is unambiguous:
  - string fields remain strings;
  - integer and float fields become numbers;
  - unknown or ambiguous paths retain the existing generic inference.
- Support exact paths, dotless suffix paths, nested fields, list fields, aliases, and computed fields.
- Preserve existing behavior for native `FilterTree` filters, non-digit strings, JSON numeric values, and callers that do not provide a schema resolver.

## Verification

- Added resolver coverage for unique, exact, ambiguous, unknown, nested, computed, aliased, and list-indexed fields.
- Added request-level coverage for string-backed and numeric subscription fields.
- Ran focused Elasticsearch DSL, request-schema, field-type, and type-mapping test modules.
- Ran mypy for the changed modules.

## Related Issues
Fixes # (issue number)

## Checklist

### DIFF
--- a/orchestrator/core/schemas/search_requests.py
+++ b/orchestrator/core/schemas/search_requests.py
@@ -14,10 +14,11 @@
 
 from typing import Any
 
-from pydantic import BaseModel, ConfigDict, Field, TypeAdapter, field_validator, model_validator
+from pydantic import BaseModel, ConfigDict, Field, PrivateAttr, TypeAdapter, model_validator
 
-from orchestrator.core.search.core.types import EntityType, RetrieverType
+from orchestrator.core.search.core.types import EntityType, FieldType, RetrieverType, UIType
 from orchestrator.core.search.filters import ElasticQuery, FilterTree, elastic_to_filter_tree
+from orchestrator.core.search.indexing.field_types import resolve_field_types
 from orchestrator.core.search.query.mixins import StructuredOrderBy
 from orchestrator.core.search.query.queries import SelectQuery
 
@@ -26,27 +27,46 @@ _ES_DSL_KEYS = frozenset({"term", "range", "wildcard", "regexp", "exists", "bool
 _ES_QUERY_ADAPTER: TypeAdapter[ElasticQuery] = TypeAdapter(ElasticQuery)
 
 
+def _resolve_digit_only_string_kind(entity_type: EntityType, path: str) -> UIType | None:
+    field_types = resolve_field_types(entity_type, path)
+    if field_types == {FieldType.STRING}:
+        return UIType.STRING
+    if field_types and field_types <= {FieldType.INTEGER, FieldType.FLOAT}:
+        return UIType.NUMBER
+    return None
+
+
 class SearchRequest(BaseModel):
     """API request model for search operations.
 
     Only supports SELECT action, used by search endpoints.
     Accepts filters in either FilterTree format or Elasticsearch DSL format.
-    ES DSL filters are auto-converted to FilterTree before processing.
+    ES DSL filters are validated on receipt and converted after the entity type is known.
     """
 
     filters: FilterTree | None = Field(
         default=None,
         description="Structured filters to apply to the search. Accepts FilterTree or Elasticsearch DSL format.",
     )
+    _elastic_filters: ElasticQuery | None = PrivateAttr(default=None)
 
-    @field_validator("filters", mode="wrap")
+    @model_validator(mode="wrap")
     @classmethod
-    def _convert_elastic_dsl_filters(cls, value: Any, handler: Any) -> FilterTree | None:
-        """Detect and convert ES DSL filters to FilterTree, bypassing re-parse."""
-        if isinstance(value, dict) and _ES_DSL_KEYS & value.keys():
-            es_query = _ES_QUERY_ADAPTER.validate_python(value)
-            return elastic_to_filter_tree(es_query)
-        return handler(value)
+    def _parse_elastic_dsl_filters(cls, value: Any, handler: Any) -> "SearchRequest":
+        """Validate ES DSL while retaining it for entity-aware conversion in to_query."""
+        if not isinstance(value, dict):
+            return handler(value)
+
+        raw_filters = value.get("filters")
+        if not (isinstance(raw_filters, dict) and _ES_DSL_KEYS & raw_filters.keys()):
+            return handler(value)
+
+        parsed_filters = _ES_QUERY_ADAPTER.validate_python(raw_filters)
+        input_data = dict(value)
+        input_data["filters"] = None
+        request = handler(input_data)
+        request._elastic_filters = parsed_filters
+        return request
 
     query: str | None = Field(
         default=None,
@@ -82,9 +102,16 @@ class SearchRequest(BaseModel):
         Returns:
             SelectQuery for search operation
         """
+        filters = self.filters
+        if self._elastic_filters is not None:
+            filters = elastic_to_filter_tree(
+                self._elastic_filters,
+                value_kind_resolver=lambda path, _value: _resolve_digit_only_string_kind(entity_type, path),
+            )
+
         return SelectQuery(
             entity_type=entity_type,
-            filters=self.filters,
+            filters=filters,
             query_text=self.query,
             limit=self.limit,
             retriever=self.retriever,
--- a/orchestrator/core/schemas/search_requests.py
+++ b/orchestrator/core/schemas/search_requests.py
@@ -14,20 +14,11 @@
 
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 
-from orchestrator.core.search.core.types import EntityType, FieldType, RetrieverType, UIType
+from orchestrator.core.search.core.types import EntityType, RetrieverType
 from orchestrator.core.search.filters import ElasticQuery, FilterTree, elastic_to_filter_tree
-from orchestrator.core.search.indexing.field_types import resolve_field_types
+from orchestrator.core.search.indexing.field_types import resolve_field_value_kind
 from orchestrator.core.search.query.mixins import StructuredOrderBy
 from orchestrator.core.search.query.queries import SelectQuery
-
-
-def _resolve_digit_only_string_kind(entity_type: EntityType, path: str) -> UIType | None:
-    field_types = resolve_field_types(entity_type, path)
-    if field_types == {FieldType.STRING}:
-        return UIType.STRING
-    if field_types and field_types <= {FieldType.INTEGER, FieldType.FLOAT}:
-        return UIType.NUMBER
-    return None
 
 
 class SearchRequest(BaseModel):
@@ -81,7 +72,7 @@ class SearchRequest(BaseModel):
         if filters is not None and not isinstance(filters, FilterTree):
             filters = elastic_to_filter_tree(
                 filters,
-                value_kind_resolver=lambda path, _value: _resolve_digit_only_string_kind(entity_type, path),
+                value_kind_resolver=lambda path, _value: resolve_field_value_kind(entity_type, path),
             )
 
         return SelectQuery(
--- a/orchestrator/core/schemas/search_requests.py
+++ b/orchestrator/core/schemas/search_requests.py
@@ -12,19 +12,13 @@
 # limitations under the License.
 
 
-from typing import Any
-
-from pydantic import BaseModel, ConfigDict, Field, PrivateAttr, TypeAdapter, model_validator
+from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 from orchestrator.core.search.core.types import EntityType, FieldType, RetrieverType, UIType
 from orchestrator.core.search.filters import ElasticQuery, FilterTree, elastic_to_filter_tree
 from orchestrator.core.search.indexing.field_types import resolve_field_types
 from orchestrator.core.search.query.mixins import StructuredOrderBy
 from orchestrator.core.search.query.queries import SelectQuery
-
-# Keys that identify an ES DSL query at the top level
-_ES_DSL_KEYS = frozenset({"term", "range", "wildcard", "regexp", "exists", "bool"})
-_ES_QUERY_ADAPTER: TypeAdapter[ElasticQuery] = TypeAdapter(ElasticQuery)
 
 
 def _resolve_digit_only_string_kind(entity_type: EntityType, path: str) -> UIType | None:
@@ -41,32 +35,13 @@ class SearchRequest(BaseModel):
 
     Only supports SELECT action, used by search endpoints.
     Accepts filters in either FilterTree format or Elasticsearch DSL format.
-    ES DSL filters are validated on receipt and converted after the entity type is known.
+    ES DSL filters are converted to FilterTree in to_query, once the entity type is known.
     """
 
-    filters: FilterTree | None = Field(
+    filters: FilterTree | ElasticQuery | None = Field(
         default=None,
         description="Structured filters to apply to the search. Accepts FilterTree or Elasticsearch DSL format.",
     )
-    _elastic_filters: ElasticQuery | None = PrivateAttr(default=None)
-
-    @model_validator(mode="wrap")
-    @classmethod
-    def _parse_elastic_dsl_filters(cls, value: Any, handler: Any) -> "SearchRequest":
-        """Validate ES DSL while retaining it for entity-aware conversion in to_query."""
-        if not isinstance(value, dict):
-            return handler(value)
-
-        raw_filters = value.get("filters")
-        if not (isinstance(raw_filters, dict) and _ES_DSL_KEYS & raw_filters.keys()):
-            return handler(value)
-
-        parsed_filters = _ES_QUERY_ADAPTER.validate_python(raw_filters)
-        input_data = dict(value)
-        input_data["filters"] = None
-        request = handler(input_data)
-        request._elastic_filters = parsed_filters
-        return request
 
     query: str | None = Field(
         default=None,
@@ -103,9 +78,9 @@ class SearchRequest(BaseModel):
             SelectQuery for search operation
         """
         filters = self.filters
-        if self._elastic_filters is not None:
+        if filters is not None and not isinstance(filters, FilterTree):
             filters = elastic_to_filter_tree(
-                self._elastic_filters,
+                filters,
                 value_kind_resolver=lambda path, _value: _resolve_digit_only_string_kind(entity_type, path),
             )
 
--- a/orchestrator/core/schemas/search_requests.py
+++ b/orchestrator/core/schemas/search_requests.py
@@ -69,7 +69,7 @@ class SearchRequest(BaseModel):
             SelectQuery for search operation
         """
         filters = self.filters
-        if filters is not None and not isinstance(filters, FilterTree):
+        if isinstance(filters, ElasticQuery):
             filters = elastic_to_filter_tree(
                 filters,
                 value_kind_resolver=lambda path, _value: resolve_field_value_kind(entity_type, path),
--- a/orchestrator/core/search/filters/elastic_dsl.py
+++ b/orchestrator/core/search/filters/elastic_dsl.py
@@ -55,8 +55,8 @@ _INVERT_OP: dict[FilterOp, FilterOp] = {
 }
 
 
+# Given a field path and its raw term value, return the UIType to use, or None to fall back to inference.
 ValueKindResolver = Callable[[str, Any], UIType | None]
-"""Given a field path and its raw term value, return the UIType to use, or None to fall back to inference."""
 
 
 def _infer_value_kind(value: Any) -> UIType:
--- a/orchestrator/core/search/filters/elastic_dsl.py
+++ b/orchestrator/core/search/filters/elastic_dsl.py
@@ -364,8 +364,9 @@ def elastic_to_filter_tree(es_query: ElasticQuery, value_kind_resolver: ValueKin
     Args:
         es_query: A parsed ElasticQuery (TermQuery, RangeQuery, WildcardQuery,
             ExistsQuery, or BoolQuery).
-        value_kind_resolver: Optional resolver for term values, used to override
-            the default value-kind inference for digit-only strings.
+        value_kind_resolver: Optional schema-aware override for ambiguous term
+            values. It is consulted for digit-only strings; all other values use
+            normal inference.
 
     Returns:
         A FilterTree suitable for compilation to SQL.
--- a/orchestrator/core/search/filters/elastic_dsl.py
+++ b/orchestrator/core/search/filters/elastic_dsl.py
@@ -21,7 +21,7 @@ FilterTree representation used by the search engine.
 from __future__ import annotations
 
 from collections.abc import Callable
-from typing import Annotated, Any, Literal
+from typing import Any, Literal, TypeAlias
 
 from pydantic import BaseModel, Field, field_validator, model_validator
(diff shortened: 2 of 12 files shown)


orchestrator/core/search/indexing/schema.py
```diff
@@ -0,0 +1,28 @@
...
+from pydantic import BaseModel
+from pydantic.fields import ComputedFieldInfo, FieldInfo
+
+
+def iter_model_field_annotations(model_type: type[BaseModel]) -> Iterable[tuple[str, Any]]:
+    """Yield declared and computed field names with their annotations."""
+    fields: dict[str, FieldInfo | ComputedFieldInfo] = dict(model_type.model_fields)
+    fields.update(getattr(model_type, "__pydantic_computed_fields__", {}))
+    for name, field in fields.items():
+        yield name, field.return_type if isinstance(field, ComputedFieldInfo) else field.annotation
```
Mark90: Instead of first concatenating dicts and then iterating and yielding it, iterate over each separately and then yield. Then you also don't need to have the `dict[str, FieldInfo | ComputedFieldInfo]` and `isinstance` check

orchestrator/core/search/indexing/field_types.py
```diff
@@ -0,0 +1,85 @@
...
+            for nested_model_type in nested_model_types:
+                _collect_field_types(nested_model_type, nested_path, field_types, ancestors)
+        else:
+            indexed_path = f"{field_path}.*" if is_list else field_path
+            field_types[indexed_path].add(FieldType.from_type_hint(annotation))
+
+
+def _all_subclasses(model_type: type[BaseModel]) -> set[type[BaseModel]]:
+    subclasses = set(model_type.__subclasses__())
+    return subclasses | {subclass for child in subclasses for subclass in _all_subclasses(child)}
```
Mark90: I would restrict to `SubscriptionModel` types and use the lifecycle registry maintained with `register_specialized_type()`

orchestrator/core/search/indexing/field_types.py
```diff
@@ -0,0 +1,85 @@
...
+    model_type: type[BaseModel],
+    path: str,
+    field_types: dict[str, set[FieldType]],
+    ancestors: set[type[BaseModel]],
+) -> None:
+    if model_type in ancestors:
+        return
+
+    ancestors = ancestors | {model_type}
+    for name, annotation in iter_model_field_annotations(model_type):
```
Mark90: We should probably restrict this to `SubscriptionModel` types. That's what we are indexing. 
Then, there are already dicts `_product_block_fields_` and `_non_product_block_fields_` that contain field names and types, that can be used instead of iterating over the raw pydantic annotations

orchestrator/core/search/filters/elastic_dsl.py
```diff
@@ -324,20 +340,21 @@ def _translate_bool(query: BoolQuery) -> FilterTree | PathFilter:
...
 
-def elastic_to_filter_tree(es_query: ElasticQuery) -> FilterTree:
+def elastic_to_filter_tree(es_query: ElasticQuery, value_kind_resolver: ValueKindResolver | None = None) -> FilterTree:
     """Convert an Elasticsearch DSL query to a FilterTree.
 
     Args:
         es_query: A parsed ElasticQuery (TermQuery, RangeQuery, WildcardQuery,
-                  ExistsQuery, or BoolQuery).
+            ExistsQuery, or BoolQuery).
+        value_kind_resolver: Optional resolver for digit-only string term values.
```
Mark90: "        value_kind_resolver: Optional resolver for digit-only string term values."

Is very specifically documented for something that should have a generic purpose

orchestrator/core/search/filters/elastic_dsl.py
```diff
@@ -285,19 +301,19 @@ def _translate_node(query: ElasticQuery) -> FilterTree | PathFilter:
...
     """Translate a bool query into a FilterTree."""
     clause = query.bool
     children: list[FilterTree | PathFilter] = []
 
-    must_children = [_translate_node(q) for q in clause.must]
-    should_children = [_translate_node(q) for q in clause.should]
-    must_not_children = _translate_must_not(clause.must_not)
+    must_children = [_translate_node(q, value_kind_resolver) for q in clause.must]
+    should_children = [_translate_node(q, value_kind_resolver) for q in clause.should]
+    must_not_children = _translate_must_not(clause.must_not, value_kind_resolver)
```
Mark90, line 331: Why does value_kind_resolver need to be passed along everywhere? This seems easy to forget at some point.

orchestrator/core/search/filters/elastic_dsl.py
```diff
@@ -52,11 +53,22 @@
...
 def _infer_value_kind(value: Any) -> UIType:
     """Infer a UIType from a raw filter value using FieldType heuristics."""
     return UIType.from_field_type(FieldType.infer(value))
 
 
+def _resolve_value_kind(field: str, value: Any, value_kind_resolver: ValueKindResolver | None) -> UIType:
+    """Resolve digit-only strings from the indexed model schema when available."""
+    if value_kind_resolver and isinstance(value, str) and value.isdigit():
+        if value_kind := value_kind_resolver(field, value):
+            return value_kind
```
Mark90: If we have to "fix" digit handling, is there anything else that's not being handled properly?

orchestrator/core/schemas/search_requests.py
```diff
@@ -26,27 +27,46 @@
...
+        raw_filters = value.get("filters")
+        if not (isinstance(raw_filters, dict) and _ES_DSL_KEYS & raw_filters.keys()):
+            return handler(value)
+
+        parsed_filters = _ES_QUERY_ADAPTER.validate_python(raw_filters)
+        input_data = dict(value)
+        input_data["filters"] = None
+        request = handler(input_data)
+        request._elastic_filters = parsed_filters
+        return request
```
Mark90: This is modifying the model instance in .. interesting ways. Difficult to follow. Why do we need a new `_elastic_filters` attribute, and are we clearing the original filters altogether?

orchestrator/core/schemas/search_requests.py
```diff
@@ -82,9 +102,16 @@ def to_query(self, entity_type: EntityType) -> SelectQuery:
         Returns:
             SelectQuery for search operation
         """
+        filters = self.filters
+        if self._elastic_filters is not None:
+            filters = elastic_to_filter_tree(
+                self._elastic_filters,
+                value_kind_resolver=lambda path, _value: _resolve_digit_only_string_kind(entity_type, path),
```
Mark90: Dealing with specific field types in the top-level SearchRequest class is not something I'd expect to see.

Mark90 (changes requested):
Not familiar enough with the philosophy of the new search, but I'm not completely convinced that this is in line with how it is meant to work. I might be wrong though.

Either way, since this was mostly generated code I have not gone lightly with review comments.

orchestrator/core/search/filters/elastic_dsl.py
florisie: value_kind_resolver is translation-scoped context: it affects only term-node value-kind determination, but a term may occur at any depth under bool.must, bool.should, or bool.must_not. Passing it explicitly guarantees the same resolver is available at every recursive node while keeping the individual translation helpers module-level, independently testable, and free of hidden state.
Alternatives would be:
• nesting all translation helpers inside elastic_to_filter_tree() to close over the resolver, which makes those helpers inaccessible for direct testing/reuse; or
• adding a context object, which would add indirection for a single dependency.
The explicit parameter is therefore the smallest design that makes the dependency visible and preserves correct behavior for nested queries

orchestrator/core/search/filters/elastic_dsl.py
florisie: Well, I don't know. I cannot directly think of any. It basically concerns these types:
"26"                  → INTEGER
"3.14"                → FLOAT
"true" / "false"      → BOOLEAN
"2024-01-15"          → DATETIME
UUID-shaped string     → UUID

As far as I can think of there are no float, boolean, datetime or uuid values that should be resolved to type string. The situation for the vlan-range is quite specific I think.

orchestrator/core/schemas/search_requests.py
florisie: I replaced that approach. filters is now explicitly typed as FilterTree | ElasticQuery | None, so the validated input remains directly on the public field. to_query(entity_type) distinguishes the two forms and converts ES DSL only once the endpoint-provided entity type is available for schema-aware term typing.
Native FilterTree inputs continue unchanged. We no longer clear or relocate the original filters, and _elastic_filters has been removed.

orchestrator/core/schemas/search_requests.py
florisie: I moved that policy to the indexing layer as resolve_field_value_kind(entity_type, path). It resolves the indexed schema types for an entity field and returns a UI type only when it is unambiguous.

orchestrator/core/search/filters/elastic_dsl.py
florisie: Agreed. I generalized the resolver’s public contract and clarified the current invocation policy:

Optional schema-aware override for ambiguous term values.
It is consulted for digit-only strings; all other values use normal inference.

orchestrator/core/schemas/search_requests.py
```diff
@@ -82,9 +68,16 @@ def to_query(self, entity_type: EntityType) -> SelectQuery:
         Returns:
             SelectQuery for search operation
         """
+        filters = self.filters
+        if filters is not None and not isinstance(filters, FilterTree):
```
Mark90: Is more explicit:

```suggestion
        filters = self.filters
        if filters and isinstance(filters, ElasticQuery):
```

orchestrator/core/search/filters/elastic_dsl.py
```diff
@@ -341,20 +358,23 @@ def _translate_bool(query: BoolQuery) -> FilterTree | PathFilter:
...
+def elastic_to_filter_tree(es_query: ElasticQuery, value_kind_resolver: ValueKindResolver | None = None) -> FilterTree:
     """Convert an Elasticsearch DSL query to a FilterTree.
 
     Args:
         es_query: A parsed ElasticQuery (TermQuery, RangeQuery, WildcardQuery,
-                  ExistsQuery, or BoolQuery).
+            ExistsQuery, or BoolQuery).
+        value_kind_resolver: Optional schema-aware override for ambiguous term
+            values. It is consulted for digit-only strings; all other values use
+            normal inference.
```
Mark90: > It is consulted for digit-only strings; all other values use inference.

That's an implementation detail about the resolver which is not relevant here. First sentence is good enough as description.

orchestrator/core/search/indexing/field_types.py
florisie: I’ve restricted the collection to models registered in SUBSCRIPTION_MODEL_REGISTRY, which are SubscriptionModel types, and resolve their lifecycle variants through lookup_specialized_type(). This uses the registry maintained by register_specialized_type() rather than walking __subclasses__(), so only registered lifecycle specializations contribute their fields.

orchestrator/core/search/indexing/field_types.py
```diff
@@ -0,0 +1,122 @@
...
+    if entity_type != EntityType.SUBSCRIPTION:
+        return frozenset()
+
+    field_types = _subscription_field_types()
+    if "." in path:
+        schema_path = ".".join("*" if segment.isdigit() else segment for segment in path.split("."))
+        return field_types.get(schema_path, frozenset())
+    return frozenset().union(
+        *(types for field_path, types in field_types.items() if field_path.rsplit(".", maxsplit=1)[-1] == path)
+    )
```
Mark90: This can be optimized quite a bit by deriving (and caching) a secondary dict from _subscription_field_types

Something like

```py
import itertools
from more_itertools import flatten

# Util function, place somewhere in orchestrator/core/utils/functional.py
def group_list_values_by_key(items: Iterable[tuple[str, list[str]]]) -> dict[str, list[str]]:
    # items must be sorted
    groups = itertools.groupby(items, key=lambda item: item[0])
    def unpack(group_lists):
        return list(flatten(lists for _group, lists in group_lists))
    return {k: unpack(v) for k, v in groups}


@lru_cache(maxsize=1)
def _subscription_field_suffix_types():
    field_types = _subscription_field_types()

    def suffix(path):
        return path.rsplit('.', maxsplit=1)[-1]

    suffix_types = sorted(((suffix(k), v) for k, v in d.items()), key=lambda i: i[0])
    return group_list_values_by_key(suffix_types)
```

Example usage

```py

In [59]: d = {
    ...:   'a.b.c': ['foo'],
    ...:   'a.b.d.c': ['bar'],
    ...:   'a.b': ['foo']
    ...:  }

In [60]:

In [60]: sorted(((suffix(k), v) for k, v in d.items()), key=lambda i: i[0])
Out[60]: [('b', ['foo']), ('c', ['foo']), ('c', ['bar'])]

In [61]: def suffix(path):
    ...:     return path.rsplit('.', maxsplit=1)[-1]
    ...:

In [62]: suffix_types = sorted(((suffix(k), v) for k, v in d.items()), key=lambda i: i[0])

In [63]: suffix_types
Out[63]: [('b', ['foo']), ('c', ['foo']), ('c', ['bar'])]

In [64]: group_list_values_by_key(suffix_types)
Out[64]: {'b': ['foo'], 'c': ['foo', 'bar']}
```

orchestrator/core/search/indexing/field_types.py
```diff
@@ -0,0 +1,122 @@
...
+        else:
+            indexed_path = f"{field_path}.*" if is_list else field_path
+            field_types[indexed_path].add(FieldType.from_type_hint(annotation))
+
+
+def _specialized_subscription_types(model_type: type[SubscriptionModel]) -> set[type[SubscriptionModel]]:
+    """Return a registered subscription model together with its lifecycle-specialized variants.
+
+    Uses the lifecycle registry maintained by ``register_specialized_type()`` instead of walking
+    ``__subclasses__()``, so only variants actually registered for this model are included.
```
Mark90:
```suggestion
    Uses the lifecycle registry maintained by ``register_specialized_type()`` so only variants
    actually registered for this model are included.
```

Remove AI's tendency to comment something it's currently doing which has no value in the future :)

orchestrator/core/search/indexing/field_types.py
florisie: I now start only from the subscription models registered by the application, together with their lifecycle-specific versions.

orchestrator/core/search/indexing/field_types.py
```diff
@@ -0,0 +1,122 @@
...
+    field_types: dict[str, set[FieldType]],
+    ancestors: set[type[BaseModel]],
+) -> None:
+    if model_type in ancestors:
+        return
+
+    ancestors = ancestors | {model_type}
+    for name, annotation in _field_annotations(model_type):
+        field_path = f"{path}.{name}"
+        is_list = get_origin(annotation) is list
```
Mark90:
```
In [66]: get_origin(list[str])
Out[66]: list

In [67]: get_origin(Annotated[list[str], "bla"])
Out[67]: typing.Annotated
```

Typing is a can of worms. Not innocent little rainworms either, more like sandworms from Arrakis :) 

Use `orchestrator.core.types.is_list_type` instead which solves most of the problems.

orchestrator/core/search/indexing/field_types.py
```diff
@@ -0,0 +1,122 @@
...
+        nested_model_types = _model_types(annotation)
+        if nested_model_types:
+            nested_path = f"{field_path}.*" if is_list else field_path
+            for nested_model_type in nested_model_types:
+                _collect_field_types(nested_model_type, nested_path, field_types, ancestors)
+        else:
+            indexed_path = f"{field_path}.*" if is_list else field_path
```
Mark90: Same in both if/else

test/unit_tests/search/indexing/test_field_types.py
```diff
@@ -0,0 +1,167 @@
...
+    """A model that references itself must not recurse infinitely."""
+
+    class RecursiveModel(BaseModel):
+        name: str
+        child: "RecursiveModel | None" = None
+
+    RecursiveModel.model_rebuild()
+
+    field_types: dict[str, set[FieldType]] = defaultdict(set)
+    _collect_field_types(RecursiveModel, "subscription", field_types, set())
```
Mark90: We have plenty of tests already: no need to test private functions

test/unit_tests/search/indexing/test_field_types.py
```diff
@@ -0,0 +1,167 @@
...
+
+    clear_field_type_cache()
+    with patch.dict(
+        SUBSCRIPTION_MODEL_REGISTRY,
+        {"STRING": StringSubscription, "NUMERIC": NumericSubscription},
+        clear=True,
+    ):
+        string_kind = resolve_field_value_kind(EntityType.SUBSCRIPTION, "string_value")
+        numeric_kind = resolve_field_value_kind(EntityType.SUBSCRIPTION, "numeric_value")
+        ambiguous_kind = resolve_field_value_kind(EntityType.SUBSCRIPTION, "ambiguous_value")
```
Mark90, line 158: If one of these paths is explicitly searched:

```
subscription.string_block.ambiguous_value
subscription.numeric_block.ambiguous_value
```

Then the kind would not be ambiguous, right? Only when the suffix `ambiguous_value` is searched?

Mark90 (commented):
Thanks for all the changes! 

I understand the purpose of the change better now.

Please update the PR description, and I have some followup comments below.

orchestrator/core/search/filters/elastic_dsl.py
Mark90: Ok. I'm still a little unsure about this part, but I don't have a better solution either.

orchestrator/core/schemas/search_requests.py
florisie: OK, updated it. The truthiness check is not even necessary when we check directly for the ElasticQuery type.

orchestrator/core/search/filters/elastic_dsl.py
florisie: Removed it.

orchestrator/core/search/indexing/field_types.py
florisie: OK, updated.

test/unit_tests/search/indexing/test_field_types.py
florisie: True, the cases where we searched for the specific paths were not covered. Now they are.

test/unit_tests/search/indexing/test_field_types.py
florisie: Addressed. I removed the direct _collect_field_types() test. Recursive-model handling remains covered through resolve_field_types()

orchestrator/core/search/indexing/field_types.py
florisie: OK, we calculate it only once now.

orchestrator/core/search/indexing/field_types.py
florisie: OK, changed it.

orchestrator/core/search/indexing/field_types.py
florisie: Nice, it has been implemented now.

orchestrator/core/schemas/search_requests.py
Mark90: Cool. Yeah I wasn't sure about the case of checking `isinstance` when the value is `None`, I remember having issues with that in the past (although it might have been `NoneType` and `issubclass` or something.. oh well, nevermind)

Mark90 (approved):

orchestrator/core/search/indexing/field_types.py
```diff
@@ -0,0 +1,125 @@
...
+        yield from (
+            (name, computed_field.return_type)
+            for name, computed_field in getattr(model_type, "__pydantic_computed_fields__", {}).items()
+        )
+        return
+
+    yield from iter_model_field_annotations(model_type)
+
+
+def _collect_field_types(
```
mrijk, line 50: This functions has 1) side effects (it mutates field_types) 2) has nested for loops and 3) is recursive (and hopefully will not reach max recursion depth).

orchestrator/core/search/indexing/schema.py
```diff
@@ -0,0 +1,27 @@
...
+from collections.abc import Iterable
+from typing import Any
+
+from pydantic import BaseModel
+
+
+def iter_model_field_annotations(model_type: type[BaseModel]) -> Iterable[tuple[str, Any]]:
+    """Yield declared and computed field names with their annotations."""
+    for name, field in model_type.model_fields.items():
+        yield name, field.annotation
```
mrijk, line 25: Minor, but you could probably rewrite this as:

`yield from ((name, filed.annotation_ for name, field in model_type.model_fields.items())`

mrijk (approved):
LCGTM (Looks Claude Generated To Me)

orchestrator/core/search/indexing/field_types.py
Mark90: The core has several functions that iterate over the domain model in a recursive way. We have not run into problems with this so far, so for me it is an acceptable pattern. 
While there is indeed a side-effect on field_types, it is in a function that is private to this module and therefore won't be tested directly. 
The nested for loop is not super pretty but not IMHO not an immediate concern until it grows larger.

orchestrator/core/search/indexing/schema.py
Mark90: Potato potahto, similarly readable to me

orchestrator/core/search/indexing/field_types.py
```diff
@@ -0,0 +1,125 @@
...
+def _model_types(annotation: Any) -> set[type[BaseModel]]:
+    """Return Pydantic model types contained in an annotation."""
+    if isinstance(annotation, type) and issubclass(annotation, BaseModel):
+        return {annotation}
+    return {model_type for arg in get_args(annotation) for model_type in _model_types(arg)}
+
+
+def _field_annotations(model_type: type[BaseModel]) -> Iterable[tuple[str, Any]]:
+    """Yield fields using domain-model classification when available."""
+    if issubclass(model_type, DomainModel):
```
torkashvand, line 38: `_field_annotations` misses all inherited fields, so none of the base `SubscriptionModel` fields ever resolve. like subscription.customer_id [string], subscription.description [string], subscription.note, subscription.product.name [string]

inspect.get_annotations(cls) returns own annotations only

Mark90: Thanks, looking into this. I don't quite see the impact yet, will discuss with @florisie

PR #1812, opened 2026-08-03
Fix search `HAS_` and `NOT_HAS_COMPONENT`

tvdven:
## Summary

Fixes the "has component / doesn't have component" filter on the (beta) subscriptions search, which returned no results, and stops these filters from producing meaningless matching fields.

**ES DSL translation (`elastic_dsl.py`)**
- `exists` now translates to `has_component` (lquery `*.port.*`) instead of `ends_with` (`*.port`). Component labels only appear mid-path in index rows, so the old pattern matched nothing — the root cause of the empty results.
- `must_not: {exists: ...}` now inverts to `not_has_component`; previously the negation was silently dropped.

**Matching fields**
- `has_component` / `not_has_component` no longer produce matching fields: every result satisfies them by definition, so the reported field was just noise (an arbitrary field from inside the component).
- The exclusion lives in a shared helper, `FilterTree.get_highlightable_leaves()`, used by both the retriever and the result resolver to keep leaf indexing aligned.
- Other path-predicate leaves (`ends_with`, …) still report all matching rows, but without highlight spans — they match on the path, not the value text.

Covered by unit tests and end-to-end integration tests against PostgreSQL.

## Related Issues
Fixes https://git.ia.surf.nl/netdev/automation/projects/orchestrator-ui/-/work_items/203

### DIFF
--- a/test/unit_tests/search/filters/test_elastic_dsl.py
+++ b/test/unit_tests/search/filters/test_elastic_dsl.py
@@ -14,7 +14,7 @@
 """Tests for orchestrator.core.search.filters.elastic_dsl: Elasticsearch DSL to FilterTree conversion, including term, range, wildcard, exists, and bool queries."""
 
 from datetime import datetime
-from typing import Any
+from typing import Any, Literal
 
 import pytest
 from pydantic import TypeAdapter, ValidationError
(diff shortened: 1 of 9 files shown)


test/unit_tests/search/filters/test_elastic_dsl.py
```diff
@@ -158,11 +174,65 @@ def test_exists_to_ltree_filter() -> None:
...
     assert leaf.value_kind == UIType.COMPONENT
 
 
+def test_must_not_exists_inverts_to_not_has_component() -> None:
+    leaf = _parse_and_get_leaf({"bool": {"must_not": [{"exists": {"field": "port"}}]}})
+    assert leaf.path == "*"
+    assert isinstance(leaf.condition, LtreeFilter)
+    assert leaf.condition.op == FilterOp.NOT_HAS_COMPONENT
+    assert leaf.condition.value == "port"
+    assert leaf.value_kind == UIType.COMPONENT
```
Mark90: Too many asserts here to my taste

test/unit_tests/search/filters/test_elastic_dsl.py
```diff
@@ -158,11 +174,65 @@ def test_exists_to_ltree_filter() -> None:
...
+    tree = elastic_to_filter_tree(es)
+    assert isinstance(tree, FilterTree)
+    assert tree.op == BooleanOperator.AND
+    status_leaf, component_leaf = tree.children
+    assert isinstance(status_leaf, PathFilter)
+    assert status_leaf.path == "subscription.status"
+    assert isinstance(component_leaf, PathFilter)
+    assert isinstance(component_leaf.condition, LtreeFilter)
+    assert component_leaf.condition.op == FilterOp.NOT_HAS_COMPONENT
+    assert component_leaf.condition.value == "port"
```
Mark90: There's way too many asserts here.

Is it possible to convert the `tree` result to something of which we can assert the structure in one go? Maybe `dic(tree)` or something

test/unit_tests/search/retrieval/test_structured_highlights.py
```diff
@@ -0,0 +1,98 @@
...
+
+def test_mixed_value_and_has_component_highlights_value_leaf_only() -> None:
+    filters = FilterTree(op=BooleanOperator.AND, children=[_value_leaf(), _component_leaf(FilterOp.HAS_COMPONENT)])
+    sql = _compile(filters)
+    assert sql.count("json_agg(") == 1
+
+
+def test_no_filters_produces_no_highlight_column() -> None:
+    sql = _compile(None)
+    assert Retriever.HIGHLIGHT_MATCHES_LABEL not in sql
```
Mark90: Not the most useful unittests ever added. I think FilterTree is a bit too low-level to test?

Mark90 (approved):

test/unit_tests/search/retrieval/test_structured_highlights.py
tvdven: Hmm, yes I agree. I removed those

test/unit_tests/search/filters/test_elastic_dsl.py
tvdven: I've converted the model to json to compare against

PR #1823, opened 2026-08-06
Enable to run specific workflows on a separate queue / worker

tvdven:
## Summary
Adds a `CELERY_TARGET_QUEUES` setting: an optional mapping of workflow `Target` to a dedicated Celery queue (e.g. `CELERY_TARGET_QUEUES='{"RECONCILE": "reconcile"}'`), so specific workflow targets can be processed by dedicated workers.

- Targets in the mapping are routed to their queue for both new starts (`_celery_start_process`) and resumes (`_celery_resume_process`); unlisted targets keep the default `task_routes` routing (`.delay()` behavior preserved via `apply_async` without a queue).
- Queue resolution happens before the session/transaction boundary in the executor, since no workflow attributes may be read after the session is closed.
- Setting is validated (non-empty queue names) and ignored under `EXECUTOR=threadpool`.
- Includes unit tests for routing and settings validation, a new "Scaling" docs section, and a CHANGELOG entry.

## Related Issues
Fixes # (issue number)

## Checklist

### DIFF
--- a/docs/guides/scaling.md
+++ b/docs/guides/scaling.md
@@ -116,6 +116,50 @@ celery.conf.task_routes = {
 
 If you decide to override the queue names in this configuration, you must also update the names accordingly after the `-Q` flag.
 
+### Dedicated queues per workflow target
+
+The four default queues separate tasks from workflows, but they cannot isolate a *class* of
+workflows whose volume or duration profile differs from interactive work. A typical example is a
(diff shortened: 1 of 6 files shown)


Mark90 (approved):
Nice!

tjeerddie (approved):

PR #1829, opened 2026-08-10
Restructure AI / Hybrid Search documentation

### LINKED ISSUE #1602: [Documentation]: Review documentation of new search in 5.0
The reference documentation describes the current 2 variants of search, DB filter queries and TSV search.
https://workfloworchestrator.org/orchestrator-core/reference-docs/search/

The new search functionality in 5.0 should be added to this, documenting how pg vector is used, the different types of retrievers, what embeddings add, etc. https://timfrohlich.com/blog/postgresql-hybrid-search can be a good start. 

There's also an old closed PR on the orchestrator-core that intended to add this documentation but it was closed. https://github.com/workfloworchestrator/orchestrator-core/pull/1075 This may give some hints.

---

**Update**: there is now a lot of documentation on https://workfloworchestrator.org/orchestrator-core/reference-docs/ai-search/#overview. The new goal is to review this and, where possible, make it less verbose.

torkashvand:
## Summary

Reworked the AI / Hybrid Search docs.

The page had decent content but it was in the wrong order. It opened with the database table and the index definitions before explaining what the thing actually does, and there was nothing at all about how to turn it on or run it. Reordered it to: what it is, how it works, how to run it, then the implementation details at the end. Added two diagrams and short explanations of ltree, trigrams and embeddings, since not everyone reading this will know those.

Moved a few things around:

* The `ai_search_paths` section was in `search.md` but belongs in `ai-search.md`. The AI search page was linking back into the classic search page three times to reach it.
* Folded the two files in `orchestrator/core/search/docs/` into `ai-search.md` and deleted them. They were never in the nav so nobody could actually read them, and `index.md` referred to tables that don't exist.
* Linked to the 5.0 upgrade guide for setup instead of repeating the steps.

While checking the text against the code I found some things that were wrong:

* Hybrid only ranks fields that trigram match, capped at 100 rows. The doc said it fused both modalities over all of the entity's embeddings.
* GraphQL field names were written snake_case, but Strawberry camelCases them (`searchPaths`, `productBlocks`, `resourceTypes`).
* Not every search is stored with a `query_id`. The agent tools always store one, REST and GraphQL only when there is a next page.
* `GET /definitions` is keyed by UI type, not field type.
* The refresh steps also run in `reconcile_workflow`, and do not run in `validate_workflow` or a plain `@workflow`.
* Process search matches the last step's `state` JSONB, not `process.state`.
* `search.md` didn't mention that TSV subscription search is deprecated since 5.0.

Also added one line to `mcp.md` pointing at the search tools, since its table didn't list them.

`mkdocs build --strict` passes.

## Related Issues
Fixes #1602

## Left alone

A few things I ran into but didn't touch here:

* `mcp.md` documents an `AgentTag` class that doesn't exist. It's `AGENT_EXPOSED_TAG` plus the `openapi_extra` constants now. It's wrong in a code sample and several bullets, so it needs its own PR.
* The 100 row limit in the hybrid retriever applies to the whole candidate set, not per entity, so a broad single word query could drop matching entities before ranking. Not sure if that's intentional.
* The index CLI docstrings still say "Index subscription_search_index", which doesn't exist. It shows up in `--help`.
* `.github/labeler.yml` has `area/search` globbed to `orchestrator/search/**`, but the package is at `orchestrator/core/search/**`, so that rule never matches. I set the label by hand here. Worth a one-line fix in a CI PR.


docs/reference-docs/ai-search.md
```diff
@@ -1,303 +1,503 @@
...
-| `ix_flat_embed_hnsw` | `HNSW (embedding vector_l2_ops) WITH (m=16, ef_construction=64)` | semantic NN via L2 distance (`<->`) |
-| `ix_flat_value_trgm` | `GIN (value gin_trgm_ops)` | fuzzy word-similarity (`<%`, `word_similarity`) |
-| `ix_flat_path_gist` | `GIST (path gist_ltree_ops)` | ltree matching (`~`, `@>`, `<@`) |
-| `ix_flat_path_btree` | `btree (path)` | exact path equality (the EAV pivot) |
-| `ix_ai_search_index_entity_id` | `btree (entity_id)` | candidate lookups by entity |
-| `idx_ai_search_index_content_hash` | `btree (content_hash)` | change detection during indexing |
+### What embeddings add
 
-Note the operator-class coupling: HNSW is built with **`vector_l2_ops`**, so semantic and hybrid retrieval use **L2 distance (`<->`)**, not cosine (`<=>`).
+An **embedding** is a list of numbers that represents the meaning of a piece of text. Texts with
```
Mark90:
```suggestion
An **embedding** is a vector (list) of numbers that represents the meaning of a piece of text. Texts with
```

docs/reference-docs/ai-search.md
```diff
@@ -1,303 +1,503 @@
...
+are embedded in batches sized against the embedding model's context window; everything else is
+stored with no embedding.
+
+Indexing is triggered from three places:
+
+- **Workflow steps**: `refresh_subscription_search_index` and `refresh_process_search_index` are
+  appended by the `create_workflow`, `modify_workflow`, `terminate_workflow` and
+  `reconcile_workflow` decorators. They swallow errors, so a failed re-index never fails the
+  workflow. Note that `validate_workflow` and workflows built with the bare `@workflow` decorator
+  do **not** refresh the index, so index those entities another way.
```
Mark90: > so index those entities another way.

Which entities?

docs/reference-docs/ai-search.md
```diff
@@ -1,303 +1,503 @@
...
-Because `EMBEDDING_DIMENSION` is baked into the `embedding` column type, changing it requires the **`embedding resize` CLI** (`python main.py embedding resize`), which deletes all indexed rows and `ALTER`s the vector columns to the new dimension (followed by a re-index).
+### Running without embeddings
 
-## Retrieval and ranking
+Leaving `EMBEDDING_API_ENABLED` at its default of `False` is a supported mode, not a broken one:
 
-Located in `orchestrator/core/search/retrieval/` and `orchestrator/core/search/query/`.
+- Indexing stores every row with `embedding = NULL`. Nothing errors.
+- Searches route to fuzzy and structured retrieval. Semantic and hybrid are simply unavailable,
+  and requests that would have used them degrade to fuzzy.
```
Mark90: This could be condensed into 1 or 2 sentences (and without the "not a broken" terminology :) )

Mark90 (approved):
Few small comments but it reads really nice now. The reference documentation at the bottom is still very packed with implementation details, but that's probably fine. 
Thanks!

PR #1840, opened 2026-08-13
Support both oauth2-lib 2.x and 3.x (httpx and httpx2)

### LINKED ISSUE #1834: Support both httpx and httpx2 (oauth2-lib 2.x and 3.x)
Goal: let consumers move to oauth2-lib 3.0.0 (`httpx` -> `httpx2`) without a major release here.

**We are already compatible with both.** orchestrator-core never touches `AsyncClient` — it does not
subclass `OIDCAuth`, only accepts one at `app.py:343`. All eleven oauth2-lib symbols we import are
byte-identical between 2.7.0 and 3.0.0. `HTTPX_SSL_CONTEXT` is a plain `ssl.SSLContext`, and `httpx`
cannot disappear because `litellm` requires it unconditionally.

Three items, then widen the constraint.

**1. `graphql/schema.py:184`** — `isinstance(error.original_error, HTTPStatusError)` uses httpx's
class, so a consumer resolver on httpx2 won't match and a 404 logs at `error` instead of `debug`.
Log level only; we have no `raise_for_status` anywhere.

```python
try:
    from httpx2 import HTTPStatusError as Httpx2StatusError
except ModuleNotFoundError:          # only the httpx stack is installed
    Httpx2StatusError = HTTPStatusError
```

**2. Declare `httpx`** — imported at `graphql/schema.py:22` and `mcp/server.py:49`, absent from
`[project.dependencies]`. Needed regardless.

**3. Docs** — `docs/reference-docs/auth-backend-and-frontend.md:282,328` hard-code
`from httpx import AsyncClient`; should tell integrators to match their oauth2-lib major.

**Then** lift the `<3` cap (#1836) to `<4` in a minor, with release notes telling consumers to check
their `OIDCAuth` subclass first.

`mcp/server.py` stays on httpx (fastmcp pins `httpx<1.0` until v4). `pytest-httpx` tests are
unaffected — they run with `OAUTH2_ACTIVE=False`.

hanstrompert:
Closes #1834. Lets consumers move to oauth2-lib 3.0.0 (`httpx` -> `httpx2`) without a major release
here.

**We were already compatible with both.** orchestrator-core never touches the client it passes to
`OIDCAuth` subclasses — it only accepts one at `app.py:343`. All eleven oauth2-lib symbols we import
are byte-identical between 2.7.0 and 3.0.0. The widening to `<4` is the substance; the rest is
cleanup it exposed.

### Why we still declare `httpx`

Two imports, neither about oauth2-lib:

- `mcp/server.py:106` — `_forward_auth_header(request: httpx.Request)` is a hook handed to fastmcp's
  `httpx_client_kwargs`. fastmcp constructs that request and pins `httpx<1.0` until v4, so the
  annotation must match what fastmcp passes.
- `graphql/schema.py` — see below.

httpx was previously resolved transitively through oauth2-lib, which drops it on 3.x. `litellm`
requires it unconditionally, so it is always installed; the declaration makes that honest.

### The 404 check was never about oauth2-lib

`process_errors` matched `httpx.HTTPStatusError` to log a downstream 404 at debug. That check
predates httpx2, and oauth2-lib raises `HTTPStatusError` nowhere (verified against 2.7.0 and 3.0.0) —
it exists for 404s from **consumer resolvers' own HTTP calls**. The real axis is whichever client the
resolver used, so matching `{httpx, httpx2}` would be an arbitrary slice of an open set.

It now matches on `.response.status_code`, covering httpx, httpx2, `requests` and whatever follows.
That also removed a conditional import, a `cast`, and an `httpx2` dev dependency an earlier draft
carried.

### Not fixed here

`nwastdlib`'s `ErrorHandlerExtension` has the same httpx-only match, and that one decides what the
**API client** sees — a non-httpx 404 currently arrives as `INTERNAL_ERROR` / "Internal Server Error".
Filed as workfloworchestrator/nwa-stdlib#73, fixed in workfloworchestrator/nwa-stdlib#74.

If that ships as 1.12.2, our `nwa-stdlib~=1.12.1` picks it up on the next lock refresh and nothing is
needed here; only a 1.13 would require a pin bump.

### Verification

1487 unit tests, ruff and format clean. `uv sync` fails locally because #1835 moved litellm to
1.95.0, whose Rust component needs a newer Cargo than mine, so I ran against the existing venv; CI
uses prebuilt wheels. mypy reports one pre-existing `context_getter` error, identical on `main`.

CI never resolves oauth2-lib 3.x (the lock pins 2.7.0), so the `<4` range is not exercised. A matrix
leg would add little, since the suite runs with `OAUTH2_ACTIVE=False`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,13 +44,12 @@ dependencies = [
     "deprecated>=1.2.18",
     "fastapi~=0.141.1",
     "fastapi-etag==0.4.0",
+    "httpx>=0.28.0,<1.0",
     "itsdangerous>=2.2.0",
     "jinja2==3.1.6",
     "more-itertools~=11.1.0",
     "nwa-stdlib~=1.12.1",
-    # Capped at <3: oauth2-lib 3.0.0 swaps httpx for httpx2, which changes the client type passed to
-    # OIDCAuth subclasses downstream. Lift to <4 once #1834 lands and the change can be announced.
-    "oauth2-lib>=2.6.1,<3",
+    "oauth2-lib>=2.6.1,<4",
     "orjson==3.11.9",
     "pgvector>=0.4.1",
     "prometheus-client==0.25.0",
--- a/docs/reference-docs/auth-backend-and-frontend.md
+++ b/docs/reference-docs/auth-backend-and-frontend.md
@@ -273,6 +273,12 @@ workflows.
 
 Below is an example illustrating how to override the default configurations:
 
+!!! note "Which `AsyncClient` to import"
+
+    The client passed to `userinfo()` comes from oauth2-lib, so import it from the library oauth2-lib
+    uses: `httpx` on oauth2-lib 2.x, `httpx2` on 3.x. Values you pass back into that client
+    (`BasicAuth`, `Timeout`, `Limits`) must come from the same one.
+
 === "`orchestrator-core` ≥ 5.0"
 
     ```python
--- a/orchestrator/core/graphql/schema.py
+++ b/orchestrator/core/graphql/schema.py
@@ -19,7 +19,6 @@ import strawberry
 import structlog
 from fastapi.routing import APIRouter
 from graphql import GraphQLError
-from httpx import HTTPStatusError
 from strawberry.extensions import SchemaExtension
 from strawberry.fastapi import GraphQLRouter
 from strawberry.schema.config import StrawberryConfig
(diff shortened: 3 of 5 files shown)


Mark90 (approved):

pyproject.toml
```diff
@@ -44,13 +44,12 @@ dependencies = [
...
     "fastapi~=0.141.1",
     "fastapi-etag==0.4.0",
+    "httpx>=0.28.0,<1.0",
     "itsdangerous>=2.2.0",
     "jinja2==3.1.6",
     "more-itertools~=11.1.0",
     "nwa-stdlib~=1.12.1",
-    # Capped at <3: oauth2-lib 3.0.0 swaps httpx for httpx2, which changes the client type passed to
-    # OIDCAuth subclasses downstream. Lift to <4 once #1834 lands and the change can be announced.
-    "oauth2-lib>=2.6.1,<3",
```
Mark90, line 53: Is there a follow-up ticket to lift the lower bound to 3 and replace httpx with httpx2? That can be attached to milestone 6.0 and labelled breaking

Mark90: Created #1844 for this

hanstrompert: Thank you for the reminder, I have added some more context to the issue.

PR #1843, opened 2026-08-14
Support wrapped primitive types in step arguments

mrijk:
## Summary
Support the option for wrapped primitive types. For now these types only live in the SURF orchestrator and wrap for example an IMS id (a primitive int that can have any value) in a class that holds that int.

An example:

```python
def my_step(ims_circuit_id: ImsId) -> None:
     return {
          "another_ims_circuit_id": ims_circuit_id
     }
```

At function entrance, an integer will be converted to an ImsId class

The return value will be converted to an integer again.

At a later stage the base classes (for example IntId) will be also move to the core.

## Related Issues
Fixes # (issue number)

## Checklist


orchestrator/core/utils/state.py
```diff
@@ -158,13 +158,22 @@ def _convert_to_uuid(v: Any) -> UUID:
     return v if isinstance(v, UUID) else UUID(v)
 
 
-def _apply_type_conversion(value: Any, annotation: Any, param_name: str) -> Any:
+def _apply_wrapped_type_conversion(value: Any, annotation: Any) -> Any:
```
tjeerddie, line 161: no tests?

mrijk: Indeed. I added a note that these test will be added as soon as the base types (IntId and UuidId) will be moved to the core. That's also the moment documentation will be added

orchestrator/core/utils/state.py
```diff
@@ -158,13 +158,22 @@ def _convert_to_uuid(v: Any) -> UUID:
...
+    if is_list_type(annotation, Any):
+        return [annotation(item) if _can_unwrap(annotation) else item for item in value]
+
+    # Note: if there is a use-case for optional wrapped types, this needs to be added here
+
+    return annotation(value) if _can_unwrap(annotation) else value
+
+
+def _apply_type_conversion(value: Any, annotation: Any) -> Any:
     """Apply UUID conversion based on annotation type.
```
Mark90: Docstring already was too specific, now it is wrong

```suggestion
    """Apply type conversion based on annotation.
```

orchestrator/core/utils/state.py
```diff
@@ -158,13 +158,22 @@ def _convert_to_uuid(v: Any) -> UUID:
...
 
-def _apply_type_conversion(value: Any, annotation: Any, param_name: str) -> Any:
+def _apply_wrapped_type_conversion(value: Any, annotation: Any) -> Any:
+    """Apply conversion from primitive types to a wrapped type."""
+    if is_list_type(annotation, Any):
+        return [annotation(item) if _can_unwrap(annotation) else item for item in value]
+
+    # Note: if there is a use-case for optional wrapped types, this needs to be added here
+
+    return annotation(value) if _can_unwrap(annotation) else value
```
Mark90: Nit: maybe define the conversion once

```suggestion
    def convert(value: Any) -> Any:
        return annotation(value) if _can_unwrap(annotation) else value

    if is_list_type(annotation, Any):
        return [convert(item) for item in value]

    # Note: if there is a use-case for optional wrapped types, this needs to be added here

    return convert(value)
```

orchestrator/core/utils/state.py
```diff
@@ -406,6 +416,25 @@ def _build_arguments(func: StepFunc | InputStepFunc, state: State) -> list:
     return arguments
 
 
+def _can_unwrap(obj: Any) -> bool:
+    """Check if an object wraps a primitive type."""
+    return hasattr(obj, "unwrap") and callable(obj.unwrap)
```
Mark90, line 424: This is fairly simple and works. But it is something we are mainly implementing that fits our own usecase.

How could we generalize this to wrapping/unwrapping any custom data type? Perhaps in the form of a hook that users can implement.

mrijk: Either that or (as you already suggested) define a protocol in the core, and test if a custom data type implements it.

Mark90: I would like to have a plan for the future, i.e. in terms of how using a protocol would work. It can be a feature ticket for now, because there should also be docs and tests for it etc. Can you spec this?

orchestrator/core/utils/state.py
```diff
@@ -406,6 +419,25 @@ def _build_arguments(func: StepFunc | InputStepFunc, state: State) -> list:
...
+    def _unwrap(value: Any) -> Any:
+        if _can_unwrap(value):
+            return value.unwrap()
+        if isinstance(value, list):
+            return [_unwrap(item) for item in value]
+        if isinstance(value, dict):
+            return _unwrap_state(value)
+        return value
+
+    return {key: _unwrap(value) for key, value in state.items()}
```
Mark90, line 438: I'd like to see at least 1 unittest added/changed to test that unwrapping works

mrijk: Done

test/integration_tests/services/test_state.py
```diff
@@ -950,3 +950,124 @@ def step_union(generic_sub: GenericProductOneInactive | GenericProductTwoInactiv
...
+            "nested_dict": {"inner": wrapped},
+        }
+
+    state = {"text": "hello"}
+    new_state = step_returns_wrapper(state)
+
+    # All wrappers should be unwrapped in the returned state
+    assert new_state["single_wrapper"] == "hello"
+    assert new_state["list_of_wrappers"] == ["x", "y"]
+    assert new_state["nested_dict"]["inner"] == "hello"
```
Mark90, line 1073: Any particular reason these need to be in integration tests? They don't seem to require postgres or redis

orchestrator/core/utils/state.py
mrijk: https://github.com/workfloworchestrator/orchestrator-core/issues/1850

test/integration_tests/services/test_state.py
mrijk: Not really. Seems that all inject_args tests (for example UUID) are also here

tjeerddie (approved):

Mark90 (approved):

PR #1858, opened 2026-08-19
List subscriptions tool and raise on empty criteria searches for agents

timfdev:
## Summary
 When an agent is asked for some example subscriptions, or a list of all of them, it uses the search tool, which silently returns an empty list when no criteria are given. That is a valid design decision for interactive clients, which treat it as transient state and show an empty result page. An agent however takes it as ground truth and reports that there are no subscriptions. 

This PR  creates a new tool for listing subscriptions and raises a clear error when the search tool is called without search parameters,

## Related Issues
Fixes # (issue number)

## Type of change

## Checklist

### DIFF
--- a/orchestrator/core/api/api_v1/endpoints/mcp_tools.py
+++ b/orchestrator/core/api/api_v1/endpoints/mcp_tools.py
@@ -39,6 +39,9 @@ Why these are curated rather than auto-generated from the existing REST API:
 * ``list_recent_processes`` —> REST ``GET /processes/?filter=k,v,k,v`` uses
   a flat positional filter syntax that LLMs handle poorly; this exposes
   named kwargs.
+* ``list_subscriptions`` —> no list route exists on ``/subscriptions``
+  (browsing lives in GraphQL) and ``search`` deliberately refuses to
+  enumerate without criteria; this gives agents a deterministic listing.
 * ``get_subscription_details`` —> REST ``GET /subscriptions/domain-model/{id}``
   returns the full product-block tree; this returns a flat header.
 """
@@ -54,7 +57,7 @@ from sqlalchemy import select
 from sqlalchemy.orm import joinedload
 
 from orchestrator.core.api.error_handling import raise_status
-from orchestrator.core.db import ProcessTable, WorkflowTable, db
+from orchestrator.core.db import ProcessTable, SubscriptionTable, WorkflowTable, db
 from orchestrator.core.mcp.server import AGENT_EXPOSED_TAG, READONLY_TOOL
 from orchestrator.core.schemas.mcp_search import (
     AggregateToolRequest,
@@ -72,6 +75,8 @@ from orchestrator.core.schemas.mcp_search import (
 from orchestrator.core.schemas.mcp_tools import (
     GetWorkflowFormRequest,
     ListRecentProcessesRequest,
+    ListSubscriptionsRequest,
+    ListSubscriptionsResponse,
     ListWorkflowsRequest,
     ProcessIdRequest,
     ProcessStatusResponse,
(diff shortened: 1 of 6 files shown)


orchestrator/core/api/api_v1/endpoints/mcp_tools.py
```diff
@@ -261,6 +267,49 @@ def list_recent_processes_endpoint(params: ListRecentProcessesRequest) -> list[P
...
+    """List subscriptions, newest first, at most 10 per page.
+
+    While ``has_more`` is true, call again with ``offset=next_offset`` for the
+    next page. Returns flat summary rows without product blocks; use
+    get_subscription_details or get_subscription_domain_model for one
+    subscription's full data, and ``search`` to find or filter subscriptions.
+    """
+    stmt = (
+        select(SubscriptionTable)
+        .options(joinedload(SubscriptionTable.product))
```
Mark90: Might make sense to use [raiseload](https://docs.sqlalchemy.org/en/21/orm/queryguide/relationships.html#sqlalchemy.orm.raiseload) for other relationships, in this and other MCP DB queries to ensure they'll never accidentally start doing lazy loading

orchestrator/core/api/api_v1/endpoints/mcp_tools.py
```diff
@@ -261,6 +267,49 @@ def list_recent_processes_endpoint(params: ListRecentProcessesRequest) -> list[P
...
+        .order_by(SubscriptionTable.start_date.desc().nulls_first(), SubscriptionTable.subscription_id)
+        .offset(params.offset)
+        .limit(params.limit + 1)
```
Mark90, line 291: Does the MCP call need to be able to skip to a certain page and actually receive the newest first?
If not then the more performant and reliable (query results can shift in between calls) approach would be to sort on subscription id and use that as a cursor. Or make the start date somehow part of the parameters.

orchestrator/core/api/api_v1/endpoints/mcp_tools.py
```diff
@@ -261,6 +267,49 @@ def list_recent_processes_endpoint(params: ListRecentProcessesRequest) -> list[P
...
 
 
+@router.post(
+    "/list_subscriptions",
+    response_model=ListSubscriptionsResponse,
+    tags=[AGENT_EXPOSED_TAG],
+    operation_id="list_subscriptions",
+    openapi_extra=READONLY_TOOL,
+)
+async def list_subscriptions_endpoint(params: ListSubscriptionsRequest) -> ListSubscriptionsResponse:
```
Mark90, line 277: Missing a test for this endpoint

Mark90 (changes requested):

orchestrator/core/api/api_v1/endpoints/mcp_tools.py
timfdev: After thinking about it a little more I think we should not allow pagination at all. Agents handle a single bounded result better than page iteration: by the time a model reaches later pages, earlier pages may already be compressed out of its context, so cross-page aggregation is unreliable (JetBrains observed this after shipping offset pagination in their MCP tools ). Iterating pages also burns tool calls and tokens for little gain.

So the tool now returns a single list with a limit default and max. When it's truncated, the agent should switch to search to drill down rather than list further. This also aligns with our other tools, such as search, which return one bounded list with a limit parameter.


[JetBrains Blog](https://blog.jetbrains.com/ruby/2026/02/rubymine-mcp-and-the-rails-toolset/) implemented pagination, then noted there are various risks with pagination.

[Zuplo](https://zuplo.com/learning-center/mapping-rest-apis-to-mcp-tools) also recommends skipping pagination, limit with a sensible default and a separate search tool because agents are "bad at iterating through pages"

[Anthropic Blog](https://www.anthropic.com/engineering/writing-tools-for-agents) Recommends steering agents towards searching instead of listing.

Mark90: Sounds good.

Mark90 (approved):

Issue #1757, opened 2026-07-02
[Bug]: `generate workflows` overwrites existing workflows/shared.py


Mark90:
### What happened?

When following the advanced workshop, while [generating workflows for the l2p2p product type](https://workfloworchestrator.org/workshops/example-orchestrator/generator/#workflows) the generator writes `workflows/shared.py` without checking if it exists.

### Version

5.0.3

### Relevant log output

```shell
(.venv) orchestrator@761f4cc1c14a:~$ python main.py generate workflows -cf templates/l2-p2p.yaml --no-dryrun --force
2026-07-02 13:39:00 [info     ] Starting the orchestrator      version=5.0.3
2026-07-02 13:39:01 [info     ] Could not get git commit hash, not setting version reason="Command '['/usr/bin/env', 'git', 'rev-parse', 'HEAD']' returned non-zero exit status 128."
2026-07-02 13:39:02 [info     ] Could not get git commit hash, not setting version reason="Command '['/usr/bin/env', 'git', 'rev-parse', 'HEAD']' returned non-zero exit status 128."
2026-07-02 13:39:02 [debug    ] Using orjson
2026-07-02 13:39:02 [info     ] Database object configured, all methods referencing `db` should work. [orchestrator.core.db]
2026-07-02 13:39:02 [info     ] wrote file                     [orchestrator.core.cli.generate] append=False force=True path=workflows/l2_p2p/shared/forms.py
2026-07-02 13:39:02 [info     ] wrote file                     [orchestrator.core.cli.generate] append=False force=True path=workflows/shared.py
2026-07-02 13:39:02 [info     ] wrote file                     [orchestrator.core.cli.generate] append=False force=True path=workflows/l2_p2p/create_l2_p2p.py
2026-07-02 13:39:02 [info     ] wrote file                     [orchestrator.core.cli.generate] append=False force=True path=workflows/l2_p2p/modify_l2_p2p.py
2026-07-02 13:39:02 [info     ] wrote file                     [orchestrator.core.cli.generate] append=False force=True path=workflows/l2_p2p/validate_l2_p2p.py
2026-07-02 13:39:02 [info     ] wrote file                     [orchestrator.core.cli.generate] append=False force=True path=workflows/l2_p2p/terminate_l2_p2p.py
2026-07-02 13:39:02 [info     ] wrote file                     [orchestrator.core.cli.generate] append=False force=True path=translations/en-GB.json
[...]

PR #1862, opened 2026-08-20
Improve generator logging when writing/overwriting/appending to files

tjeerddie:
related: https://github.com/workfloworchestrator/orchestrator-core/issues/1757

### DIFF
--- a/orchestrator/core/cli/generate.py
+++ b/orchestrator/core/cli/generate.py
@@ -52,7 +52,9 @@ def write_file(path: Path, content: str, append: bool, force: bool) -> None:
         if not path.parent.exists():
             logger.info("creating missing folder(s)", path=str(path.parent))
             path.parent.mkdir(parents=True, exist_ok=True)
-        if not force and path.exists():
+
+        file_exists = path.exists()
+        if not force and file_exists:
             action = "append" if append else "overwrite"
             logger.warning(f"file already exists, rerun with --force if you want to {action}", path=str(path))
             return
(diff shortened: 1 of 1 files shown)


kvklink (approved):

orchestrator/core/cli/generate.py
```diff
@@ -52,7 +52,9 @@ def write_file(path: Path, content: str, append: bool, force: bool) -> None:
         if not path.parent.exists():
             logger.info("creating missing folder(s)", path=str(path.parent))
             path.parent.mkdir(parents=True, exist_ok=True)
-        if not force and path.exists():
+
+        file_exists = path.exists()
```
Mark90: This variable and `if not force and file_exists` section could be moved to before the `try:`, as an early return. Makes the flow clearer, at least for me.

orchestrator/core/cli/generate.py
```diff
@@ -63,7 +65,8 @@ def write_file(path: Path, content: str, append: bool, force: bool) -> None:
     except Exception as exception:
         logger.error("failed to write file", path=str(path), message=str(exception))
     else:
-        logger.info("wrote file", path=str(path), append=append, force=force)
+        overwrote = file_exists and not append
+        logger.info("overwrote file" if overwrote else "wrote file", path=str(path), append=append, force=force)
```
Mark90: If we're going to tinker with log messages: they can be even more consistent :) 


```
# error does not mention whether this was an append and/or an overwrite
logger.error("failed to write file", path=str(path), message=str(exception))  

# success messages uses past tense and alternating message based on overwrite (not ideal to `grep`), and append could be written out
logger.info("overwrote file" if overwrote else "wrote file", path=str(path), append=append, force=force)
```


What about something like this? 

```
# derive once based on the combination of `force`, `file_exists` and `append`
action = "write" / "overwrite" / "append"

logger.error("write file failed", action=action, path=str(path))

logger.info("write file successful", action=action, path=str(path))
```

Mark90 (approved):

PR #1876, opened 2026-08-26
Index processes and subscriptions automatically on process exit

### LINKED ISSUE #1867: [Search]: Update index on process exit
Issue to be refined further:

Currently there are a few steps at the end of each workflow that update the search index of the subscription and process.
The flaw in this approach is that when a workflow suspends or fails at an intermediate step, the search index is not updated.

Therefore we want to update the indexes during the "finalization" of the workflow, **after** all database changes from the workflow have been processed.

pboers1988:
## What

Moves search indexing of processes and subscriptions out of workflow steps into an automatic hook that fires on every process exit (complete, fail, abort, suspend, callback timeout). This fixes three problems:
- Workflows that failed mid-execution, were aborted, or suspended were never indexed
- Indexed processes showed `last_status="running"` forever instead of their real terminal status
- `@task()`, `@workflow()` (raw), and `@validate_workflow` workflows were never indexed at all

Closes #1867.

## How

New module `orchestrator/core/search/indexing/hooks.py` called from three process-exit points in `processes.py`. Each entity is indexed independently in its own `db.database_scope()`. Subscription ids are extracted from the final state via `match`/`case` with UUID validation. Failure handling is configurable via `SEARCH_INDEXING_STRICT`.

## Backwards compatibility

Not a breaking change. `refresh_subscription_search_index` and `refresh_process_search_index` remain as backwards-compatible no-ops that emit `DeprecationWarning` — downstream step lists import and run unchanged. Both are removed from the four decorator step lists.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


test/unit_tests/search/indexing/test_hooks.py
```diff
@@ -0,0 +1,245 @@
...
+    index_process_and_subscriptions(process_id, process_variant({}))
+
+    mock_run_indexing.assert_called_once_with(EntityType.PROCESS, str(process_id))
+
+
+@patch("orchestrator.core.search.indexing.hooks.run_indexing_for_entity")
+def test_subscriptions_from_state_are_indexed(mock_run_indexing):
+    process_id = uuid4()
+
+    index_process_and_subscriptions(process_id, Success({"subscription_id": SUB_ID_A}))
```
pboers1988: Testing mocks useless

pboers1988: Fixed in 9fad4672 — you were right, and it applied more broadly than this one test.

The whole integration file was patching `run_indexing_for_entity` and asserting it was called with ids the test itself supplied, so it would have stayed green through a wrong `EntityType`, an indexer raising inside its own scope, or broken session wiring. Those tests now run the real indexer and assert on rows in `ai_search_index` (no embedding service needed — `EMBEDDING_API_ENABLED` is off by default), including `process.last_status` in the indexed row, which is the actual defect this PR fixes and is only observable in the persisted row.

Same pattern elsewhere, all addressed:
- This test (`test_subscriptions_from_state_are_indexed`) — deleted. `extract_subscription_ids` is already covered as a pure function, and the end-to-end path is now covered for real.
- `test_process_is_always_indexed` parametrized 8 `Process` variants over a path that never branches on variant (`unwrap` is base-class only, and the process is indexed before it's called) — collapsed to the 3 state shapes that actually differ.
- `test_safe_index_process_calls_hook` — deleted; mocked the hook to assert the hook was called.
- `test_synchronous_call_sites_index_on_exit_in_a_fresh_scope` mocked `db` wholesale then asserted `database_scope` was called, so its 'protects the caller's session' claim was untested — replaced with a real-session integration test that makes indexing raise and asserts the caller's session still works.
- `test_refresh_search_index_step_does_not_index` — deleted; it patched `indexing.tasks` but `steps.py` no longer imports from `search` at all, so it could never fail.
- Stopped patching the hook's own `extract_subscription_ids` to pin an ordering the hook never promises.

Where a call count is still genuinely useful (proving the deprecated steps don't double-index), the spy uses `wraps=` so the real indexer still runs.

Net −65 lines. Full unit suite 1546 passed, integration 262 passed, mypy and ruff clean.

tjeerddie (approved):
LGTM

orchestrator/core/services/processes.py
```diff
@@ -95,6 +95,16 @@ def _safe_broadcast_process_update(process_id: UUID, broadcast_func: BroadcastFu
         logger.exception("Failed to broadcast process update", process_id=process_id)
 
 
+def _safe_index_process(process_id: UUID, result: WFProcess) -> None:
+    """Index a process and its subscriptions on exit; no-op when the search extra is not installed."""
+    try:
+        from orchestrator.core.search.indexing.hooks import index_process_and_subscriptions
+    except ImportError:
+        return
```
Mark90: Which search extra? I think this is no longer applicable?

Mark90 (approved):

PR #1879, opened 2026-08-28
Feature/1269 store process subscription

### LINKED ISSUE #1269: [Feature]: Add more information in workflow/task list page table for processes in status `CREATED`
### What should we build?

In the following screenshot you can see that the information of the started workflows is minimal. It is missing the product type, customer, affected subscription. Would be a nice improvement if these fields can be filled in from `initial state` (?) of the process. Only from the workflow name one can make a best effect guess which product is involved.

Image

kvklink:
## Summary
Deprecate using the `store_process_subscription()` step in favor of having orchestrator-core do that automatically. It will:

* Store when creating a workflow (modify, terminate)
* **not** store when this workflow is a task (system, validate) or a creation workflow
* Store when calling `SubscriptionModel.from_product_id()` with a `process_id`

The function is idempotent, so we only store if not exists. The old step also still works, but now emits a deprecation warning.

## Related Issues
Fixes #1269

## Type of change

## Checklist

### DIFF
--- a/docs/guides/product-modeling/imports.md
+++ b/docs/guides/product-modeling/imports.md
@@ -29,7 +29,23 @@ user input from a form page, but could also be provided programmatically by eith
 called. Data sources can include external API resources, CSV- or YAML files, etc. If desired, interaction with all
 external provisioning systems can be skipped, resulting in a creation workflow that could be as simple as follows.
 
-=== "`orchestrator-core` ≥ 5.0"
+=== "`orchestrator-core` ≥ 5.1"
+
+    ```python
+    from orchestrator.core.workflow import StepList, begin
+    from orchestrator.core.workflows.utils import create_workflow
+
+    @create_workflow("Create imported Node")
+    def create_imported_node() -> StepList:
+        """Workflow to import a Node without provisioning it."""
+        return (
+            begin
+            >> create_subscription
+            >> initialize_subscription
+        )
+    ```
+
+=== "`orchestrator-core` == 5.0"
 
     ```python
     from orchestrator.core.workflow import StepList, begin
--- a/docs/guides/product-modeling/imports.md
+++ b/docs/guides/product-modeling/imports.md
@@ -29,7 +29,7 @@ user input from a form page, but could also be provided programmatically by eith
 called. Data sources can include external API resources, CSV- or YAML files, etc. If desired, interaction with all
 external provisioning systems can be skipped, resulting in a creation workflow that could be as simple as follows.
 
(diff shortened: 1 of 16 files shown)


orchestrator/core/services/processes.py
```diff
@@ -527,6 +534,14 @@ def create_process(
 
     with transactional(db, logger):
         _db_create_process(pstat)
+        if user_inputs and (subscription_id := user_inputs[0].get("subscription_id")):
+            # Only store the Process Subscription relation in workflows where if a subscription ID is present.
```
tjeerddie:
```suggestion
            # Only store the Process Subscription relation in workflows where a subscription ID is present.
```

tjeerddie (approved):

docs/guides/product-modeling/imports.md
```diff
@@ -29,7 +29,23 @@ user input from a form page, but could also be provided programmatically by eith
 called. Data sources can include external API resources, CSV- or YAML files, etc. If desired, interaction with all
 external provisioning systems can be skipped, resulting in a creation workflow that could be as simple as follows.
 
-=== "`orchestrator-core` ≥ 5.0"
+=== "`orchestrator-core` ≥ 5.1"
```
Mark90: Realistically this will be in 5.4. But good idea to add this :)

orchestrator/core/services/processes.py
```diff
@@ -527,6 +534,14 @@ def create_process(
 
     with transactional(db, logger):
         _db_create_process(pstat)
+        if user_inputs and (subscription_id := user_inputs[0].get("subscription_id")):
+            # Only store the Process Subscription relation in workflows where a subscription ID is present.
+            # For creation workflows where this is not the case, this is handled in
+            # `SubscriptionModel.from_product_id()`.
+            # For tasks, this is not stored.
+            workflow_table = get_workflow_by_name(workflow.name)
+            if workflow_table and not workflow_table.is_task:
```
Mark90: Would be nice to condense this into 1 `if`, something like 

```py
if _is_workflow(workflow) and (subscription_id := _get_from_first_input(user_inputs, "subscription_id"):
```

Mark90 (approved):
LGTM, 2 small comments

docs/guides/product-modeling/imports.md
kvklink: Ah, something to keep up to date then. Shall we aim for 5.4?

Mark90: Yep!

PR #1882, opened 2026-08-31
Feature/1729 rework last status handling

### LINKED ISSUE #1729: [Improvement]: Rework assignee/last_status handling for tasks
The `assignee` and `last_status` of a task currently have this SURF specific handling

```py
        if process_state.isfailed() and p.is_task:
            # Check if we need a special failed status:
            # If it is an AssertionError:
            if step_state.get("class") == "AssertionError" or step_state.get("class") == "InconsistentData":
                p.assignee = Assignee.NOC
                p.last_status = ProcessStatus.INCONSISTENT_DATA
            # If we encounter a connectivity issue with an underlying api:
            elif step_state.get("class") == "MaxRetryError" or (
                step_state.get("class") == "ApiException"
                and step_state.get("status_code")
                in (
                    HTTPStatus.BAD_GATEWAY,
                    HTTPStatus.SERVICE_UNAVAILABLE,
                    HTTPStatus.GATEWAY_TIMEOUT,
                )
            ):
                p.assignee = Assignee.SYSTEM
                p.last_status = ProcessStatus.API_UNAVAILABLE
            else:
                p.assignee = Assignee.SYSTEM
```

Source: https://github.com/workfloworchestrator/orchestrator-core/blob/11c09d0e77d9586d918757525c664a68f3b71554/orchestrator/core/services/processes.py#L213

This code has not been touched since 2021.

Things to improve about it:
* Remove SURF specific handling (Assignee.NOC)
* Reliably match the correct exception (e.g. `InconsistentData` was renamed to `InconsistentDataError` some years ago, but because it's a string match this did not propagate here)
* Perhaps add a hook for implementors to provide their own handling

kvklink:
## Summary
Allow for custom handling of storing the process state after each step in the subscription database.
Also removes some SURF-specific behaviour and prevents drift on how we match error names to process states.

## Related Issues
Fixes #1729 

## Type of change

## Checklist

### DIFF
--- a/docs/reference-docs/workflows/process-step-handler.md
+++ b/docs/reference-docs/workflows/process-step-handler.md
@@ -1,0 +1,27 @@
+# Custom Process Step Handler
+
+After each workflow step is completed, the Orchestrator will store the output of this step in the `Process` table in the
+subscription database. By default, this will update the `last_status`, and in case the step was a failure, it will set
+the `failed_reason` and the `traceback`.
+
+You might want this function to include extra behaviour, custom error mapping, or special cases like setting the
+assignee of the process depending on step outcome. In this case, it is possible to override the default process function
+in `orchestrator.core.services.processes`. If you override this like the example below, you can include any kind of
+custom behaviour fit for your use-case.
+
+```python title="your_orchestrator/__init__.py"
+import orchestrator.core.services.processes
+
+def my_custom_handler(p: ProcessTable, process_state: WFProcess) -> ProcessTable:
+    """I want to set my custom assignees based on step outcome!"""
+
+    step_state: State = process_state.unwrap()
+    if process_state.isfailed():
+        error_name = step_state.get("class")
+
+        p.failed_reason = step_state.get("error", default="Blame the intern.")
+        p.assignee = "not-me@worfloworchestrator.org"
+        p.traceback = f"An '{error_name}' occurred! Help!"
+
+orchestrator.core.services.processes.PROCESS_STEP_HANDLER = my_custom_handler
+```
--- a/docs/reference-docs/workflows/process-step-handler.md
+++ b/docs/reference-docs/workflows/process-step-handler.md
@@ -1,5 +1,7 @@
 # Custom Process Step Handler
 
+!!! info "Added in version 5.4"
+
 After each workflow step is completed, the Orchestrator will store the output of this step in the `Process` table in the
 subscription database. By default, this will update the `last_status`, and in case the step was a failure, it will set
 the `failed_reason` and the `traceback`.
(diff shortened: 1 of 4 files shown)


tjeerddie (approved):

orchestrator/core/services/processes.py
```diff
@@ -234,6 +218,24 @@ def _update_process(process_id: UUID, step: Step, process_state: WFProcess) -> P
     return p
 
 
+#: The Process Step handler function is called before each step is to be stored in the database.
+#: By default, it will set the failed reason, traceback, and last status of the process.
+#: This function can be replaced by a custom `ProcessHandlerFunc` if the user needs custom behavior or error handling.
+PROCESS_STEP_HANDLER: ProcessHandlerFunc = _default_process_func
```
Mark90: Hmm this hook makes it possible to override a bit more than just the assignee/last_status; users now need to remember to reuse `_default_process_func` in their hook to not lose the core's handling of failed_reason/traceback (which is essential because this has safeguards for database contraints etc)

I would scope the hook such that it can be used _inside_ `_default_process_func` , i.e. that it returns a tuple of the custom assignee/status if the user wants to override that (or `None` for an error they just want to use the default handling for)

Also, it may be nice to have a function to register the hook which then also validates it has the correct the arguments/return type?

docs/reference-docs/workflows/process-step-handler.md
```diff
@@ -0,0 +1,27 @@
...
+After each workflow step is completed, the Orchestrator will store the output of this step in the `Process` table in the
+subscription database. By default, this will update the `last_status`, and in case the step was a failure, it will set
+the `failed_reason` and the `traceback`.
+
+You might want this function to include extra behaviour, custom error mapping, or special cases like setting the
+assignee of the process depending on step outcome. In this case, it is possible to override the default process function
+in `orchestrator.core.services.processes`. If you override this like the example below, you can include any kind of
+custom behaviour fit for your use-case.
+
+```python title="your_orchestrator/__init__.py"
```
Mark90, line 14: Useful to mention somewhere that this is new in 5.4 :)

Mark90 (changes requested):

docs/reference-docs/workflows/process-step-handler.md
kvklink: Agreed! But I think this is something that belongs in a changelog but not really anywhere else.

Mark90: That too, but also this document should say something along the lines of "Added in 5.4" so that users on <5.4 know they can't use this yet. 

A bit like how python does this in its docs:

image

We don't do this (much) yet in our docs but imo we should do it more often.

orchestrator/core/services/processes.py
kvklink: The default function now no longer takes care of setting the traceback to safe boundaries. This is moved to the orchestrator internals instead.

docs/reference-docs/workflows/process-step-handler.md
kvklink: I've used the "Note" admonition for now, I couldn't find a better alternative
image

Mark90 (approved):
LGTM

PR #1891, opened 2026-09-03
Use a per-entity HNSW index for semantic retrieval

timfdev:
## Summary
The semantic retriever’s exhaustive plan scores every embedded field of every candidate and ranks them with window functions, forcing a sequential scan instead of using the HNSW index. By default, multi-word text queries with an available embedding are routed to this retriever, making this the normal cost of multi-word semantic search. Using the index requires two additional measures: a shared index treats entity_type as a post-filter and can exhaust its search frontier before reaching the requested type, while a plain ORDER BY <-> LIMIT n can stop at roughly ef_search rows unless iterative scanning is enabled.

Eligible interactive semantic searches now take the nearest SEARCH_SEMANTIC_CANDIDATE_LIMIT fields (default 2000) directly from the entity type’s partial HNSW index and rank only those. The current production candidate shape, including structured filters, uses this bounded plan, with filters applied inside the index scan. Candidate queries containing additional relational semantics, such as joins or grouping, fall back to the unchanged exhaustive plan. Searches requesting more entities than the configured candidate window and all exports also remain exhaustive.

The shared HNSW index is replaced by one partial index per entity type, and the entity type is rendered as a literal so PostgreSQL can select the appropriate partial index. The engine applies hnsw.iterative_scan = relaxed_order for the duration of the search transaction. It applies the setting inside a savepoint so pgvector versions older than 0.8 continue without iterative scanning, returning a smaller candidate window instead of failing the transaction.

Behavior change: bounded semantic pagination ends at the candidate-window edge. Ranking within the window is exact; only the selection of fields entering the window is approximate. On a local database dump, unfiltered semantic queries dropped from approximately 450 ms to 20–30 ms.

Deploy notes: creating the indexes holds a write lock on ai_search_index for the duration of each index build. Iterative scanning requires pgvector 0.8 or newer, which the current CI database image supports.

## Related Issues
Fixes # (issue number)

## Type of change

## Checklist


pboers1988 (approved):

Mark90:
Will review today/tomorrow.

Question in advance: if there are any integration/acceptance tests that hit the semantic search path, would it be possible to have one decorated with codspeed?

timfdev:
> Will review today/tomorrow.
> 
> Question in advance: if there are any integration/acceptance tests that hit the semantic search path, would it be possible to have one decorated with codspeed?

Added it, but note that it won't show this PR's speedup. You can use the CLI speedtest against a full database to see the difference in performance on semantic retrieval.

orchestrator/core/search/retrieval/session.py
```diff
@@ -0,0 +1,72 @@
...
+    Each setting runs in its own savepoint so a value the server rejects does not abort the search
+    transaction, and rolling that savepoint back also undoes the setting. Rejection happens on
+    pgvector < 0.8, which reserves the ``hnsw`` prefix but does not define ``hnsw.iterative_scan``:
+    the search then runs without it, returning fewer semantic candidates rather than failing.
+    Rejections are logged once per setting per process.
+    """
+    for setting in settings:
+        try:
+            async with db_session.begin_nested():
+                await db_session.execute(text(setting.statement))
```
Mark90, line 60: With a remote DB these extra roundtrips on each search may start to add up. Not sure how much, but something to keep in mind.

docs/reference-docs/ai-search.md
```diff
@@ -320,16 +320,29 @@ Each index serves one match type:
...
+single entity type, and on a shared index that restriction is a post-filter: the scan walks the
+nearest vectors of *all* types and can exhaust its candidate frontier before reaching a single row
+of the type being searched. A partial index makes the scan walk only that type's vectors. For the
+planner to match the partial index the retriever renders the entity type as a **literal**, not a
+bind parameter.
+
+Reading a bounded window from the index also needs `hnsw.iterative_scan = relaxed_order`, which the
+engine applies with `SET LOCAL` for the duration of each search transaction. Without it the scan
+stops at roughly `ef_search` rows regardless of the requested limit. The setting arrived in
+**pgvector 0.8**; on older versions Postgres rejects it, and because each setting is applied inside
```
Mark90, line 343: Which versions support this? Good to document that here

orchestrator/core/search/retrieval/session.py
```diff
@@ -0,0 +1,72 @@
...
+async def apply_session_settings(db_session: AsyncSession, settings: Sequence[SessionSetting]) -> None:
+    """Apply ``SET LOCAL`` settings, which last for the search transaction and no longer.
+
+    Each setting runs in its own savepoint so a value the server rejects does not abort the search
+    transaction, and rolling that savepoint back also undoes the setting. Rejection happens on
+    pgvector < 0.8, which reserves the ``hnsw`` prefix but does not define ``hnsw.iterative_scan``:
+    the search then runs without it, returning fewer semantic candidates rather than failing.
+    Rejections are logged once per setting per process.
+    """
+    for setting in settings:
```
Mark90, line 57: Might as well exclude settings in `_rejected_session_settings` from the loop over `settings` here?

orchestrator/core/search/retrieval/session.py
```diff
@@ -0,0 +1,72 @@
...
+
+# pgvector >= 0.8: keep walking the HNSW index until the LIMIT is met instead of stopping at ~ef_search rows.
+HNSW_ITERATIVE_SCAN = SessionSetting("hnsw.iterative_scan", "relaxed_order")
+
+
+# Settings the database refused, so the warning is logged once per process instead of once per search.
+_rejected_session_settings: set[str] = set()
+
+
+async def apply_session_settings(db_session: AsyncSession, settings: Sequence[SessionSetting]) -> None:
```
Mark90, line 48: 2 points that Claude came up with that may or may not be useful, feel free to ignore

> 2. It re-applies within a transaction where it's already set. execute_search_with_fallback runs the structured pass and then up to three broadening rungs on the same session (fallback.py:203, then _run_broadening_fallback). SET LOCAL persists to the end of the transaction, so passes 2–4 redo all three roundtrips to set something that's already set. Memoizing on db_session.info keyed by the setting name would collapse that.

> 4. Worth considering: probe once, then skip the savepoint. The savepoint exists solely to survive one failure mode that is a fixed property of the deployment. Reading current_setting('hnsw.iterative_scan', true) once per engine, or checking the pgvector version, would let the steady state be a bare SET LOCAL — one roundtrip. The tempting further step, putting it in the DSN options or a connect event so it's already on the connection, gets you to zero, but it converts the setting from transaction-scoped to connection-scoped and hands it to the indexer's HNSW queries too. That may well be fine, but it's a real change in blast radius, not a free win.

Mark90 (approved):

PR #1902, opened 2026-09-10
Search: Add process subscriptions to indexing for process search filtering

tjeerddie:
- Add `workflow-target` as a property on the `Process` table for indexing.
- Add `selectinload` for `process_subscriptions`, `subscriptions`, and `product` to the process index.
- Add `_STATIC_ENTITY_SCHEMAS` dict for static entities. Currently only includes `process`, but is structured so `Product`, `ProductBlock`, etc. can be added later.
- Support list-valued response columns in search endpoints:
   - Response columns that resolve to a schema-known list field (e.g. `process.subscriptions.*.subscription_id`) are now auto-detected and returned as dict/grouped lists.
   - Non-schema-known lists using the `.*.` wildcard return a list instead of requiring indexes whose length is unknown.


tjeerddie:
draft as it still waits on merge of the targetted branch

orchestrator/core/schemas/process.py
```diff
@@ -85,3 +85,51 @@ class ProcessPatchSchema(OrchestratorBaseModel):
 
 
 Reporter = Annotated[str, Field(max_length=100)]
+
+
+class ProcessSubscriptionIndexSchema(OrchestratorBaseModel):
+    """Minimal subscription summary embedded in a process's search index entry."""
```
Mark90, line 91: Are there more of these search specific schemas?

I'd kind of prefer these to not spread throughout the schemas meant for the API because they serve a different purpose. Maybe move them somewhere below orchestrator/core/search ?

tjeerddie: yea will move them, there are no other schema's yet, but will be added later for other search entities

orchestrator/core/search/query/builder.py
```diff
@@ -427,6 +431,119 @@ def convert(row: Row, index: int) -> ResponseColumnValue:
     }
 
 
+def _list_prefix(entity_type: EntityType, path: str) -> str | None:
+    """Return the leading segment of `path` that names a list field, if any."""
+    segments = path.split(".")
+    prefixes = (".".join(segments[:i]) for i in range(1, len(segments)))
+    return next((prefix for prefix in prefixes if is_list_path(entity_type, prefix)), None)
```
Mark90, line 438: `is_list_path` iterates over all paths in the entities until it founds one for which `prefix` is a list prefix, then returns True, otherwise returns False


For a path `a.b.c.d`, the `prefixes` variable above resolves to 

```
In [26]: list(prefixes)
Out[26]: ['a', 'a.b', 'a.b.c']
```

So for each of these prefixes we'll iterate over all entity paths. Performance is not the main concern in python, but we can probably do better here

orchestrator/core/search/indexing/field_types.py
```diff
@@ -122,21 +130,54 @@ def _subscription_field_suffix_types() -> dict[str, frozenset[FieldType]]:
...
+    return None
+
+
+def is_list_path(entity_type: EntityType, path: str) -> bool:
+    """Return True if the schema-normalized path (numeric segments as '*') is a list field."""
+    field_types = _field_types_for_entity(entity_type)
+    if field_types is None:
+        return False
+    wildcard_path = f"{path}.*"
+    return any(candidate.startswith(wildcard_path + ".") for candidate in field_types)
```
Mark90, line 167: We could have a cached function that computes all list paths for an entity; then this lookup eventually becomes O(1)

orchestrator/core/search/query/builder.py
```diff
@@ -427,6 +431,119 @@ def convert(row: Row, index: int) -> ResponseColumnValue:
...
+
+def split_response_columns(response_columns: list[str], entity_type: EntityType) -> tuple[list[str], list[str]]:
+    """Split response columns into flat (scalar) paths and wildcard list paths.
+
+    A plain path through a schema-known list field is auto-detected and normalized to wildcard form.
+    A path pinning a concrete numeric index stays flat.
+    """
+    resolved = [(column, _as_list_path(entity_type, column)) for column in response_columns]
+    flat_paths = [column for column, list_path in resolved if list_path is None]
+    list_paths = [list_path for _, list_path in resolved if list_path is not None]
```
Mark90, line 465: Can use a generator expression and `more_itertools.partition` to replace these 3 lines into 1, and then call `list()` on either result for the return value

orchestrator/core/search/query/builder.py
```diff
@@ -427,6 +431,119 @@ def convert(row: Row, index: int) -> ResponseColumnValue:
...
+
+def build_response_list_rows_query(
+    entity_ids: list[str],
+    entity_type: EntityType,
+    list_paths: list[str],
+) -> Select:
+    """Build a query returning the raw EAV rows for one or more wildcard list paths in a single round-trip.
+
+    Unlike build_response_columns_query, this does not pivot in SQL: it fetches one row per
+    (entity_id, path) match and leaves grouping/merging by list index to process_response_list_columns.
```
Mark90, line 477: Remove the "Unlike ..." phrasing (documenting what other functions do is not needed)

orchestrator/core/search/query/builder.py
```diff
@@ -427,6 +431,119 @@ def convert(row: Row, index: int) -> ResponseColumnValue:
...
+    return select(AiSearchIndex.entity_id, AiSearchIndex.path, AiSearchIndex.value, AiSearchIndex.value_type).where(
+        AiSearchIndex.entity_id.in_(entity_ids),
+        AiSearchIndex.entity_type == entity_type.value,
+        or_(*lquery_matches),
+    )
+
+
+def _group_suffixes_by_prefix(list_paths: list[str]) -> dict[str, set[str]]:
+    """Group wildcard list paths into a prefix -> requested suffixes mapping."""
+    requested_suffixes: dict[str, set[str]] = defaultdict(set)
```
Mark90, line 493: Since defaultdicts aren't exactly the same as normal dicts, what do you think of using a normal dict and `requested_suffixes.setdefault(prefix, set()).add(suffix)` here instead?

orchestrator/core/search/query/builder.py
```diff
@@ -427,6 +431,119 @@ def convert(row: Row, index: int) -> ResponseColumnValue:
...
+    for path in list_paths:
+        prefix, _, suffix = path.partition(LIST_COLUMN_WILDCARD_MARKER)
+        requested_suffixes[prefix].add(suffix)
+    return requested_suffixes
+
+
+def _match_list_row(path_str: str, requested_suffixes: dict[str, set[str]]) -> tuple[str, int, str] | None:
+    """Return the (prefix, index, suffix) a row's path matches, or None if it matches no requested path."""
+    matching_prefixes = [prefix for prefix in requested_suffixes if path_str.startswith(f"{prefix}.")]
```
Mark90, line 502: A variable named `requested_suffixes` being used to match prefixes here, is a bit confusing

Mark90 (changes requested):
(Not fully reviewed yet)

orchestrator/core/search/schemas/__init__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2019-2026 SURF, GÉANT.
```
Mark90, line 1: Nit: copyright for new files should not automatically include all orgs

orchestrator/core/search/query/engine.py
```diff
@@ -31,15 +33,45 @@
...
+
+    if flat_paths:
+        col_stmt = build_response_columns_query(entity_ids, entity_type, flat_paths)
+        col_rows = (await db_session.execute(col_stmt)).all()
+        for entity_id, columns in process_response_columns(col_rows, flat_paths).items():
+            merged[entity_id].update(columns)
+
+    if list_paths:
+        list_stmt = build_response_list_rows_query(entity_ids, entity_type, list_paths)
+        list_rows = (await db_session.execute(list_stmt)).all()
```
Mark90, line 66: I presume we can have both flat and list paths in the same search query?

If so, try to perform both sqlalchemy queries concurrently to save a DB roundtrip

orchestrator/core/search/schemas/process.py
```diff
@@ -0,0 +1,81 @@
...
+    description: str
+    customer_id: str
+    product_name: str | None = None
+    product_tag: str | None = None
+    # Only populated when the registered SubscriptionTable exposes these (e.g. via OrchestratorCore.register_table).
+    customer_name: str | None = None
+    customer_abbreviation: str | None = None
+
+    @classmethod
+    def from_subscription(cls, subscription: Any) -> "ProcessSubscriptionIndexSchema":
```
Mark90, line 38: Can't we enforce `subscription: SubscriptionTable` here?

orchestrator/core/search/schemas/process.py
```diff
@@ -0,0 +1,81 @@
...
+        """Build a summary from a (possibly app-specific) SubscriptionTable instance."""
+        product = subscription.product
+        return cls(
+            subscription_id=subscription.subscription_id,
+            description=subscription.description,
+            customer_id=subscription.customer_id,
+            product_name=product.name,
+            product_tag=product.tag,
+            customer_name=getattr(subscription, "customer_name", None),
+            customer_abbreviation=getattr(subscription, "customer_abbreviation", None),
```
Mark90, line 48: These attributes come from our `SurfSubscriptionTable`, not the core's `SubscriptionTable`

orchestrator/core/search/query/engine.py
```diff
@@ -31,15 +33,45 @@
...
+        list_column_data = process_response_list_columns(list_rows, list_paths)
+        for prefix, entities_for_prefix in list_column_data.items():
+            for entity_id in entity_ids:
+                merged[entity_id][prefix] = entities_for_prefix.get(entity_id, [])
+
+    return dict(merged) if merged else None
```
Mark90, line 72: Can you ensure we have a codspeed/benchmark integration test that exercises both the flat, list and flat+list code paths? We don't have to completely optimize things yet, but will be good to start tracking the performance over time. Have a feeling there is some room for improvement here. :)

timfdev:
Each linked subscription adds three embedded string fields per process (description, product name, product tag). Product name and tag are the same text for every subscription of a product, so this compounds the existing embedding duplication problem I opened #1916 for.

Should not be fixed here but just flagging it so the index growth after the process reindex is explainable.

orchestrator/core/search/query/builder.py
```diff
@@ -14,8 +14,9 @@
 from collections import defaultdict
```
timfdev, line 14: This is more than a new shape for indexed processes: it rewrites the response-column path for every entity type. 

IMO that is a lot of risk  for client convenience. Given an entity_id a client can already fetch any data it wants, typed and nested, from the existing GraphQL routes. The search index is the wrong source for display data anyway: it does not hold all of an entity's fields, values are stringified, and nested data is trimmed.

Personally I would keep the indexing change only.

orchestrator/core/search/query/builder.py
```diff
@@ -401,29 +367,147 @@ def _restore_value_type(value: str | None, value_type: str | None) -> ResponseCo
...
+    if not matching_prefixes:
+        return None
+
+    prefix = max(matching_prefixes, key=len)
+    tail = path_str[len(prefix) + 1 :]
+    index_str, _, suffix = tail.partition(".")
+    if suffix not in suffixes_by_prefix[prefix]:
+        return None
+
+    return prefix, int(index_str), suffix
```
timfdev, line 479: int() looks reachable by non-ints

orchestrator/core/search/query/builder.py
```diff
@@ -401,29 +367,147 @@ def _restore_value_type(value: str | None, value_type: str | None) -> ResponseCo
...
+        if path_str in flat_path_set:
+            value = row.value
+            values_by_entity[str(row.entity_id)][path_str] = _restore_value_type(
+                None if value is None else str(value), row.value_type
+            )
+
+    return dict(values_by_entity)
+
+
+def _as_list_path(entity_type: EntityType, column: str) -> str | None:
```
timfdev, line 402: What about a path that is a list in one product and a block in another? On my local db
  `subscription.port.port` is both: SN8 Aggregated Service Port indexes
  `subscription.port.port.0.admin_state`, the MSC products index `subscription.port.port.admin_state`.

Requesting `subscription.port.port.admin_state` gets rewritten to the list form here, so the MSC
  rows are fetched by the lquery but dropped by `_match_list_row` (no numeric index segment), and
  those subscriptions get `[]` instead of their value.

orchestrator/core/search/query/builder.py
tjeerddie: will need to be discussed more but for now removed the response list logic

PR #1904, opened 2026-09-10
Fix MemoryWebsocketManager not cleaning up clients connections that die mid-send

### LINKED ISSUE #1903: [Bug]: Dead websocket leaks into connections_by_pid and silences broadcasts to healthy clients
### What happened?

When a websocket client's TCP connection dies without a close handshake, `MemoryWebsocketManager`
raises out of `connect()`, leaves the dead socket registered in `connections_by_pid`, and every
subsequent `broadcast_data()` call then delivers nothing to any client on that channel until the
process restarts.

Three separate defects on one path:

**1. `remove_ws` checks the wrong state attribute.** `memory_websocket_manager.py:68` guards on
`websocket.client_state`, but `WebSocket.send()` gates on `application_state`
(`starlette/websockets.py`, the final `else:` in `send`). The two diverge when a send fails: on
`OSError`, starlette sets `application_state = DISCONNECTED` and raises `WebSocketDisconnect(1006)`,
while `client_state` stays `CONNECTED` because no `websocket.disconnect` message was ever received.
`remove_ws` sees `CONNECTED`, calls `disconnect()` -> `close()` -> `send()`, and starlette raises
`RuntimeError: Cannot call "send" once a close message has been sent.`

Note this does *not* happen on a clean client disconnect — there `receive()` sets
`client_state = DISCONNECTED` before raising, and the guard correctly skips the close. It needs a
failed write, e.g. the `__pong__` reply in `connect()` hitting a reset socket.

**2. The exception escapes, so cleanup never runs.** The `RuntimeError` propagates out of
`disconnect()` before `remove_ws` reaches its removal block on line 69, so the socket stays in
`connections_by_pid` forever. It also escapes `connect()` and surfaces to the user as a full
`Exception in ASGI application` traceback from uvicorn.

**3. One leaked socket silences the whole channel.** `broadcast_data` wraps its entire nested loop
in `try: ... except (RuntimeError, ValueError): pass`. The leaked socket raises on `send_text`, and
the bare `pass` abandons the loop — every client after it in the list gets nothing. Because the
dead socket is never removed, this repeats on every broadcast.

### Expected

The dead socket is dropped from `connections_by_pid`, no traceback is logged, and other clients on
the channel keep receiving broadcasts.

### Suggested fix

- `remove_ws`: guard on `application_state` (what `send` actually checks), and make the removal
  block run regardless of whether `disconnect()` raised.
- `broadcast_data`: move the `try`/`except` inside the per-socket loop so one bad socket is skipped
  and removed rather than ending the broadcast.

### Reproduction

`repro_ws.py` is attached in a comment below — no server or browser needed, asserts all three
failures.

```
1. connect() raised: Cannot call "send" once a close message has been sent.
2. client_state=CONNECTED application_state=DISCONNECTED
3. dead socket still registered: {'events': [<starlette.websockets.WebSocket object at 0x10e622660>]}
4. healthy client received: []

All three failures reproduced.
```

### Version

5.3.0 (also present on `main` at b1e568aa — the file is byte-identical)

Python 3.13

### Relevant log output

```shell
[info ] "WebSocket /api/ws/events" [accepted]                      [uvicorn.error]
[info ] connection open                                            [uvicorn.error]
[info ] Websocket Connections: 1                                   [...memory_websocket_manager]
... six minutes later, client goes away ...
[error] Exception in ASGI application                              [uvicorn.error]
Traceback (most recent call last):
  File ".../uvicorn/protocols/websockets/websockets_sansio_impl.py", line 409, in run_asgi
    result = await self.app(self.scope, self.receive, self.send)
  [...]
  File ".../orchestrator/core/api/api_v1/endpoints/ws.py", line 78, in websocket_events
    await websocket_manager.connect(websocket, WS_CHANNELS.EVENTS)
  File ".../orchestrator/core/websocket/websocket_manager.py", line 59, in connect
    await self._backend.connect(websocket, channel)
  File ".../orchestrator/core/websocket/managers/memory_websocket_manager.py", line 42, in connect
    await self.remove_ws(websocket, channel)
  File ".../orchestrator/core/websocket/managers/memory_websocket_manager.py", line 69, in remove_ws
    await self.disconnect(websocket)
  File ".../orchestrator/core/websocket/managers/memory_websocket_manager.py", line 49, in disconnect
    await websocket.close(code=code)
  File ".../starlette/websockets.py", line 181, in close
    await self.send({"type": "websocket.close", "code": code, "reason": reason or ""})
  File ".../starlette/websockets.py", line 98, in send
    raise RuntimeError('Cannot call "send" once a close message has been sent.')
RuntimeError: Cannot call "send" once a close message has been sent.
```

hanstrompert:
Fixes #1903.

A client whose TCP connection dies without a close handshake left a dead websocket registered in
`connections_by_pid`, and every later broadcast on that channel then reached nobody.

### Changes

**`remove_ws` guards on `application_state`.** `WebSocket.send()` gates on `application_state`, not
`client_state`. The two diverge when a send fails rather than the client disconnecting: starlette
sets `application_state = DISCONNECTED` and raises `WebSocketDisconnect(1006)`, while `client_state`
stays `CONNECTED` because no disconnect message ever arrived. The old guard therefore called
`close()` on an already-closed socket and starlette raised `RuntimeError`.

**The close is suppressed and cleanup always runs.** The close inherently races the peer going away,
so `RuntimeError`/`WebSocketDisconnect` no longer escape and the socket is removed from the registry
either way. This also stops the traceback surfacing as `Exception in ASGI application`.

**`broadcast_data` sends per client.** The blanket `try`/`except` around the whole nested loop meant
one failing socket abandoned the broadcast for every client after it. Each send is now guarded
individually; clients that fail are dropped afterwards.

A `_connections()` helper returns a snapshot of `(channel, websocket)` pairs so the registry can be
mutated while iterating.

### Tests

Seven new parametrized cases covering the state guard, cleanup when the close raises, `connect()`
after a failed send, and a broadcast reaching clients queued behind a dead one. All seven fail on
`main` and pass here.

The mock websocket helper now sets `application_state` (defaulting to `client_state`, with an
override to model the two diverging). No existing assertion needed changing.

`1568 passed` in `test/unit_tests`; ruff and mypy clean.

### DIFF
--- a/orchestrator/core/websocket/managers/memory_websocket_manager.py
+++ b/orchestrator/core/websocket/managers/memory_websocket_manager.py
@@ -12,6 +12,10 @@
 # limitations under the License.
 
 
+from collections.abc import Iterable
+from contextlib import suppress
+from itertools import chain
+
 from fastapi import WebSocket, WebSocketDisconnect, status
 from starlette.websockets import WebSocketState
 from structlog import get_logger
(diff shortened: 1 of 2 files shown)


Mark90 (approved):
LGTM

Mark90:
Slightly reworded the the phrasing of the title. MemoryWebsocketManager is not really recommended for production, but definitely good to fix this.

PR #1915, opened 2026-09-16
Add load_schedules helper for registering schedules from code

### LINKED ISSUE #1873: Registering schedules from code is undocumented
The guides describe one way to create a schedule: `POST /api/schedules` (guides/tasks.md,
guides/upgrading/4.7.md). No Python entry point is mentioned.

Core's own `scheduler load-initial-schedule` does not use that endpoint — it calls
`add_unique_scheduled_task_to_queue` (`orchestrator/core/cli/scheduler.py:220`), which has no
`__all__` and no documentation.

The 4.7 guide makes removing the decorator mandatory for 5.0.0, so projects are pushed off code-based
scheduling toward the REST API. That leaves deploy-time registration without a documented option: a
Helm init container has neither an HTTP client nor a token, and the function core itself uses appears
to be internal.

Please document either that `add_unique_scheduled_task_to_queue` is supported public API, or the
intended way to register project schedules outside a running API.

hanstrompert:
Closes #1873.

## Problem

The guides describe one way to create a schedule: `POST /api/schedules`. That is the wrong tool at
deploy time, where there is no reachable API and no token — a Helm init container that has just run
the migrations, for instance. Core's own `load-initial-schedule` does not use the endpoint either;
it calls `add_unique_scheduled_task_to_queue`, which is undocumented.

The result is that every project copies core's loop. nsi-orchestrator's `schedules.py` is the same
twelve lines as `cli/scheduler.py`: resolve workflow by name, enrich with `workflow_id`, queue.

## Change

`load_schedules(schedules, *, recreate=False)` in `schedules/service.py` takes the schedule dicts
consumers already write and returns the `workflow_name` of each one it skipped as unknown. Core's
CLI prints those and continues; a project deploy can raise on them, since a missing workflow means
the task was never registered or its migration has not run.

`load_initial_schedule` now calls it, with core's five schedules hoisted to `INITIAL_SCHEDULES`.
Behaviour is unchanged apart from the placeholder `"workflow_id": ""` entries becoming unnecessary.

Schedules registered this way are ordinary API-managed schedules — linker-table row, `source: API`,
editable through the API and the UI. Existing schedules are left alone unless `recreate=True`, so a
UI edit survives the next deploy.

## Docs

- New "Registering your own schedules from code" section in `guides/tasks.md`.
- The schedule API section listed six triggers; three of them (`CalendarIntervalTrigger`,
  `AndTrigger`, `OrTrigger`) return a 422, because `APSchedulerJobCreate.trigger` is
  `Literal["interval", "cron", "date"]`. Noted there rather than only in the new section.
- `load_schedules` rendered in `reference-docs/cli.md`.

## Tests

`load_schedules` unit tests in `test/unit_tests/schedules/test_service.py`; the
`load-initial-schedule` CLI tests are parametrized over the three found/missing cases.
120 tests pass in `test/unit_tests/cli/`; ruff, ruff format and mypy clean.


hanstrompert:
Both addressed in c63449fe.

**Issue 1 — valid.** `if name` filtered the skipped list on truthiness, so an empty `workflow_name`
registered nothing but was dropped from the result. A deploy raising on a non-empty list would have
reported success while silently omitting the job — the exact failure the return value exists to
prevent. The filter is now `is not None`, the lookup is skipped for a falsy name rather than querying
the DB for it, and the reported value is coerced to `str` so the return type holds for `None`.
Regression test added for `""` and `None`; it fails against the previous commit.

**Issue 2 — applied, with a correction.** The new test code is now annotated. For the record, the
repository does not currently enforce this on tests: the mypy pre-commit hook excludes `test/.*`, and
the pre-existing tests in `test_scheduler.py` (`test_show_schedule_empty`, `test_force_task_*`) are
unannotated, so the new code matched its neighbours rather than violating a gate. Annotations are
cheap and `test/unit_tests/schemas/test_schedules.py` already uses them, so I added them to
everything this PR touches. I left the surrounding unannotated tests alone as unrelated churn.

138 tests pass; ruff, ruff format and mypy clean.

orchestrator/core/cli/scheduler.py
```diff
@@ -27,16 +26,47 @@
...
     workflow_scheduler_queue,
 )
-from orchestrator.core.schemas.schedules import APSchedulerJobCreate
-from orchestrator.core.services.workflows import get_workflow_by_name
 from orchestrator.core.settings import app_settings
 from orchestrator.core.utils.redis_client import create_redis_client
 
 app: typer.Typer = typer.Typer()
 
+INITIAL_SCHEDULES: list[dict] = [
```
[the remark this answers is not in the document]
hanstrompert: Applied in 6843745d — `list[dict[str, Any]]`, matching the `load_schedules` parameter type. That
consistency is the real argument for it: the constant and the function consuming it now read the same.

The stated justification does not hold, though. There is no strict annotation requirement being
violated here:

- `disallow_any_generics` is not set in `[tool.mypy]`, so bare `dict` is accepted by design
- mypy passed on this file before the change
- bare `list[dict]` is the prevailing style in core — `exception_handlers.py:29`,
  `validate_products.py:140`, `utils/fixed_inputs.py:30`, and throughout `cli/generator/`

CLAUDE.md's "Type annotations required everywhere (mypy strict)" refers to that mypy config, and the
config does not ask for parameterized generics. This is the second finding citing a requirement
stricter than the repository enforces.

Not taking the `TypedDict` suggestion: `load_schedules` accepts plain dicts by design so consumers can
write literals without importing a type. Pinning core's constant to a stricter shape than the API it
feeds would be a mismatch, not an improvement.

orchestrator/core/schedules/service.py
```diff
@@ -136,6 +138,38 @@ def add_unique_scheduled_task_to_queue(payload: APSchedulerJobCreate, *, recreat
...
+        registered. Raise on a non-empty list to fail a deploy on an unmigrated workflow.
+    """
+
+    def load(schedule: dict[str, Any]) -> str | None:
+        """Queue one schedule, returning its workflow name when the workflow is unknown."""
+        workflow_name = schedule["workflow_name"]
+        workflow = get_workflow_by_name(workflow_name) if workflow_name else None
+        if not workflow:
+            logger.warning("Skipping schedule for unknown workflow", workflow_name=workflow_name)
+            return str(workflow_name)
```
mrijk: Why the conversion with str ?

orchestrator/core/schedules/service.py
```diff
@@ -136,6 +138,38 @@ def add_unique_scheduled_task_to_queue(payload: APSchedulerJobCreate, *, recreat
...
+        payload = APSchedulerJobCreate(**{**schedule, "workflow_id": workflow.workflow_id})
+        logger.info("Loading schedule", payload=payload)
+        add_unique_scheduled_task_to_queue(payload, recreate=recreate)
+        return None
+
+    return [name for name in map(load, schedules) if name is not None]
```
mrijk: I think map (and filter for that matter) are somewhat discouraged in Python in favour of list comprehension. Or you can probably use more-itertools.filter_map

test/unit_tests/cli/test_scheduler.py
```diff
@@ -96,66 +100,32 @@ def test_force_task_no_args_kwargs_uses_defaults():
...
-def test_load_initial_schedule_all_missing():
+@pytest.mark.parametrize(
+    "known_workflows, expected_added",
+    [
+        pytest.param(INITIAL_SCHEDULE_NAMES, len(INITIAL_SCHEDULES), id="all-found"),
+        pytest.param(INITIAL_SCHEDULE_NAMES[:3], 3, id="some-found"),
+        pytest.param([], 0, id="none-found"),
+    ],
+)
+def test_load_initial_schedule(known_workflows: list[str], expected_added: int) -> None:
```
mrijk: I think in general we don't bother about typing in unit tests

test/unit_tests/schedules/test_service.py
```diff
@@ -0,0 +1,54 @@
...
+
+from unittest.mock import Mock
+from uuid import NAMESPACE_DNS, uuid5
+
+import pytest
+
+from orchestrator.core.schedules.service import load_schedules
+
+
+@pytest.mark.parametrize("mock_add_schedule", [["my_task"]], indirect=True)
```
tjeerddie: this `mock_add_schedule` parametrize is confusing, does it add the value to the mock_add_schedule fixture?
It also seems unnecessary when it doesn't actually add more then one entry

orchestrator/core/schedules/service.py
```diff
@@ -136,6 +138,38 @@ def add_unique_scheduled_task_to_queue(payload: APSchedulerJobCreate, *, recreat
...
+    """
+
+    def load(schedule: dict[str, Any]) -> str | None:
+        """Queue one schedule, returning its workflow name when the workflow is unknown."""
+        workflow_name = schedule["workflow_name"]
+        workflow = get_workflow_by_name(workflow_name) if workflow_name else None
+        if not workflow:
+            logger.warning("Skipping schedule for unknown workflow", workflow_name=workflow_name)
+            return str(workflow_name)
+        payload = APSchedulerJobCreate(**{**schedule, "workflow_id": workflow.workflow_id})
```
mrijk: trying to figure out what happens here, but probably:

```
schedule_with_id = schedule | {"workflow_id": workflow.workflow_id}
APSchedulerJobCreate(**schedule_with_id)
```

tjeerddie (approved):

mrijk (approved):

orchestrator/core/schedules/service.py
mrijk: Of nog eenvoudiger:

APSchedulerJobCreate(**schedule_with_id, "workflow_id": workflow.workflow_id})

test/unit_tests/cli/test_scheduler.py
hanstrompert: This is what you get when you let AI talk to each other, Greptile thought that typing tests was mandatory, Claude disagreed but wanted to accommodate Greptile and added the typing after all.

orchestrator/core/schedules/service.py
hanstrompert: Removed. It only existed to satisfy the return type after Greptile's P1, and it was papering over a
real distinction: `workflow_name` is typed `str`, so an empty or `None` value is a malformed schedule,
not an unknown workflow. It now raises `ValueError` instead of joining the skipped list. 9fe5f9c1

orchestrator/core/schedules/service.py
hanstrompert: Agreed, now a comprehension:

```python
return [name for schedule in schedules if (name := load(schedule)) is not None]
```

Skipped `filter_map` — it filters on truthiness, which is the bug this branch just fixed. 9fe5f9c1

orchestrator/core/schedules/service.py
hanstrompert: Applied, `schedule | {"workflow_id": workflow.workflow_id}`. 9fe5f9c1

test/unit_tests/cli/test_scheduler.py
hanstrompert: Reverted — the new tests are unannotated like the rest of the file. 9fe5f9c1

test/unit_tests/schedules/test_service.py
hanstrompert: Correct on both counts: it fed the fixture through `indirect=True`, and two of the three tests carried
a single value. Replaced with a factory fixture the test calls:

```python
def test_load_schedules_returns_unknown_workflow_names(mock_schedule_queue):
    mock_add = mock_schedule_queue("known_task")
```

9fe5f9c1

orchestrator/core/schedules/service.py
```diff
@@ -136,6 +138,46 @@ def add_unique_scheduled_task_to_queue(payload: APSchedulerJobCreate, *, recreat
...
+
+    def load(schedule: dict[str, Any]) -> str | None:
+        """Queue one schedule, returning its workflow name when the workflow is unknown."""
+        workflow_name = schedule["workflow_name"]
+        if not workflow_name:
+            raise ValueError(f"Schedule {schedule.get('name', schedule)!r} has no workflow_name")
+        workflow = get_workflow_by_name(workflow_name)
+        if not workflow:
+            logger.warning("Skipping schedule for unknown workflow", workflow_name=workflow_name)
+            return workflow_name
```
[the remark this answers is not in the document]
hanstrompert: Valid, fixed in 0c1cf3b9. Reproduced first: `workflow_name: 123` returned `[123]` from a `list[str]`.

The previous guard only caught falsy values, so a truthy non-string was looked up, missed, and
reported as an unknown workflow — sending the caller to check migrations for a name that was never
valid. Guard is now `not isinstance(workflow_name, str) or not workflow_name`, and the message
includes the value. Test parametrized over `""`, `None`, `123` and `["task_a"]`.

docs/guides/tasks.md
```diff
@@ -295,6 +300,105 @@ The scheduler automatically loads any schedules that are imported before the sch
...
+validated as it is loaded, so a typo in `trigger_kwargs` raises a `ValidationError` rather than
+producing a job that silently never fires:
+
+| Field | Required | Description |
+|---|---|---|
+| `workflow_name` | yes | Name of the task to run, as registered in `workflows/__init__.py` and created by a migration. |
+| `trigger` | yes | One of `"interval"`, `"cron"` or `"date"`. |
+| `trigger_kwargs` | no (default `{}`) | Keyword arguments for that trigger, passed straight to APScheduler. |
+| `name` | no | Human readable name, shown in `show-schedule` and the UI. Defaults to the workflow description. |
+| `user_inputs` | no (default `[{}]`) | Input form data for the task, one `dict` per form page. |
```
Mark90, line 337: Can this be inspected from the object itself? 

```
::: orchestrator.core.schemas.schedules.APSchedulerJobCreate
    options:
        ...
```

Not sure if that works, but that should roughly be the syntax

docs/guides/tasks.md
```diff
@@ -295,6 +300,105 @@ The scheduler automatically loads any schedules that are imported before the sch
...
+
+Core's CLI prints the skipped names and carries on; a project deploy is usually better off failing on
+them, since a missing workflow means the task was never registered in `workflows/__init__.py` or its
+migration has not run.
+
+A schedule that is malformed rather than unmigrated — no `workflow_name`, or `trigger_kwargs` the
+trigger rejects — raises instead of being reported as skipped, so it cannot be mistaken for a
+workflow that is merely missing. Uncaught, that exits the command non-zero with a traceback, which
+fails the init container running it. Do not wrap the command in a retry loop that treats any non-zero
+exit as "not ready yet": a malformed schedule will never become valid, and the retry hides it.
```
Mark90, line 370: > Do not wrap the command in a retry loop that treats any non-zero exit as "not ready yet": a malformed schedule will never become valid, and the retry hides it.

I wouldn't know why anyone would consider it sensible to put that in a retry loop. It's a bit like "Also do not stick your finger in a blender as that hurts"

docs/guides/tasks.md
```diff
@@ -295,6 +300,105 @@ The scheduler automatically loads any schedules that are imported before the sch
...
+
+!!! Warning
+    Like `load-initial-schedule`, this is only idempotent once a scheduler has run. The deduplication
+    reads the `workflows_apscheduler_jobs` linker table, which the scheduler writes while draining the
+    queue. Against a database whose scheduler has never run, each invocation queues the job again.
+
+Schedules registered this way are ordinary API-managed schedules: they show as `source: API` in
+`show-schedule` and stay editable through the API and the UI. That also makes the code a starting
+point rather than a source of truth — an existing schedule for the workflow is left alone unless
+`recreate=True`, so a later edit through the UI survives the next deploy.
```
Mark90, line 400: And, maybe good to add, a schedule existing in DB is not deleted if the code definition is deleted

orchestrator/core/schedules/service.py
```diff
@@ -136,6 +138,49 @@ def add_unique_scheduled_task_to_queue(payload: APSchedulerJobCreate, *, recreat
...
+            for example ``trigger_kwargs`` the trigger rejects.
+    """
+
+    def load(schedule: dict[str, Any]) -> str | None:
+        """Queue one schedule, returning its workflow name when the workflow is unknown."""
+        workflow_name = schedule["workflow_name"]
+        if not isinstance(workflow_name, str) or not workflow_name:
+            raise ValueError(
+                f"Schedule {schedule.get('name', schedule)!r} has no valid workflow_name: {workflow_name!r}"
+            )
```
Mark90, line 171: I would validate/raise this before creating anything

orchestrator/core/schedules/service.py
```diff
@@ -136,6 +138,49 @@ def add_unique_scheduled_task_to_queue(payload: APSchedulerJobCreate, *, recreat
...
+        workflow = get_workflow_by_name(workflow_name)
+        if not workflow:
+            logger.warning("Skipping schedule for unknown workflow", workflow_name=workflow_name)
+            return workflow_name
+        payload = APSchedulerJobCreate(**(schedule | {"workflow_id": workflow.workflow_id}))
+        logger.info("Loading schedule", payload=payload)
+        add_unique_scheduled_task_to_queue(payload, recreate=recreate)
```
Mark90, line 178: What happens when the given `schedules: dict` payload contains multiple schedules for the same workflow? I suppose with recreate=False it only registers the first schedule (unless it already is in the DB), whereas with recreate=True it only registers the last

It's worth considering if we want `load_schedules` to evaluate that the given payload as a whole makes sense. So maybe it should detect duplicate workflow schedules.

Mark90 (approved):

docs/guides/tasks.md
hanstrompert: Tried it against our mkdocs config. The syntax works, but the output is the class name, its bases, and
a collapsing source dump of the class body — no field table. mkdocstrings reads docstrings, not
pydantic `Field` metadata, so `description=`, defaults and required-ness only appear as part of the
pasted source. `AppSettings` in settings-overview.md renders the same way.

Extracting them needs the `griffe-pydantic` extension, which is not in the docs group or mkdocs.yml.
Worth its own issue — it would improve every rendered model — but out of scope here.

Keeping the table, which also lets it say `workflow_id` is omitted; the model has it as required,
because for the API it is.

docs/guides/tasks.md
hanstrompert: Fair. Cut to one clause: "...fails the init container running it rather than resolving on a retry." 2efd464e

docs/guides/tasks.md
hanstrompert: Added:

> Removing an entry from the list does not remove its schedule: `load_schedules` only adds, and nothing
> reconciles the database against the declaration. Delete the schedule through the API or the UI, or
> leave the entry in place until you do.

2efd464e

orchestrator/core/schedules/service.py
hanstrompert: Done in 2efd464e. Loading is now three ordered phases — validate every workflow name, reject
duplicates, build every payload — before anything is queued, so a bad entry at index 3 no longer
leaves 0-2 on the queue. Test covers a malformed entry after a valid one.

Still only atomic for malformed input: unknown workflows queue the rest by design, and a Redis
failure mid-loop can still partially apply.

orchestrator/core/schedules/service.py
hanstrompert: Your analysis holds, with a wrinkle: the dedup reads the linker table the scheduler writes while
draining the queue, so within one call it is timing-dependent. On a fresh database `recreate=False`
queues both rather than only the first.

Now raises in 2efd464e:

    Multiple schedules declared for workflow(s): [...]. load_schedules registers at most one
    schedule per workflow; use the API for more.

Splitting across two `load_schedules` calls does not get you two schedules either — the constraint is
in `add_unique_scheduled_task_to_queue`, which finds the existing linker row and either skips or
recreates. The REST API can attach several schedules to one workflow, as it calls the non-unique
variant; hence the pointer in the message.

Issue #1777, opened 2026-07-10
[Bug]: Investigate and fix pstat.update() without reference update


Mark90:
`ProcessStat.update` creates a new ProcessStat instance with the updated variables and leaves the old instance unaffected:

```py
>>> 
>>> stat.update(state={'foo': 'bar'})
ProcessStat(process_id=UUID('54d9ec43-c6bc-40fe-bb4b-a1f3e107716d'), workflow='wf', state={'foo': 'bar'}, log=[], current_user='SYSTEM', user_model=None)
>>> stat.update(current_user='john')
ProcessStat(process_id=UUID('54d9ec43-c6bc-40fe-bb4b-a1f3e107716d'), workflow='wf', state={}, log=[], current_user='john', user_model=None)
>>> stat
ProcessStat(process_id=UUID('54d9ec43-c6bc-40fe-bb4b-a1f3e107716d'), workflow='wf', state={}, log=[], current_user='SYSTEM', user_model=None)
>>> 
>>> stat = stat.update(current_user='john', state={'foo': 'bar'})
>>> stat
ProcessStat(process_id=UUID('54d9ec43-c6bc-40fe-bb4b-a1f3e107716d'), workflow='wf', state={'foo': 'bar'}, log=[], current_user='john', user_model=None)
>>> 
```

So code that current calls `update()` without storing the reference to the updated object will not persist the changes.

We have quite a few cases where this is missing:

<img width="656" height="103" alt="Image" src="https://github.com/user-attachments/assets/0a0c055f-9344-475b-a54b-32172f9c8461" />

The decision is to make `.update()` the values in the current instance.
The reason is that there is little usage of this method, and the current implementation copies the entire `ProcessStat` object even for the smallest of changes.

## Tasks

- [ ] Identify all current wrong usages of `.update()` in the core. Estimate the current impact of the bug, and what would change after the fix
- [ ] Deprecate `ProcessStat.update()` with message to update the value directly on the instance
- [ ] Change all existing wrong usages

tjeerddie:
used in:
- abort_process (services/processes.py)
  - Impact of the bug is that the abort process was always done by `SYSTEM` because the user was never overwritten.
- thread_start_process and thread_resume_process (services/executors/threadpool.py)
  - the `pstat.update()` does nothing and the merge of the dict only accidentally works because the `StateMerger.merge` updates the state as a side-effect.

PR #1918, opened 2026-09-17
deprecate ProcessStat.update() and directly update the class instance

tjeerddie:
- fixes the abort process always being done by SYSTEM user because it never updated the user.
  - this is because load_process() constructs the pstat with `SYSTEM_USER` as current_user.
- change the state merge with pstat.update in thread_process_start and resume to use a new merge_state function instead of StateMerger.merge that updated the state as a side effect.
  - the pstat.update() never did anything here.
  - changed `deepmerge` to a dev package to check that the new `merge_state` gives the exact same result as the `StateMerger` from `deepmerge` package.
-  update tests.

related: https://github.com/workfloworchestrator/orchestrator-core/issues/1777


pyproject.toml
```diff
@@ -40,7 +40,6 @@ dependencies = [
     "anyio>=3.7.0",
     "apscheduler>=3.11.0",
     "click==8.*",
-    "deepmerge==3.0.1",
```
Mark90, line 43: Should be mentioned in the PR title/description

orchestrator/core/workflow.py
```diff
@@ -660,6 +661,7 @@ class ProcessStat:
     current_user: str
     user_model: OIDCUserModel | None = None
 
+    @deprecated(reason=("Assign the field(s) directly instead, e.g. `pstat.current_user = user`. This method will be removed in a future version."))
```
Mark90, line 664: You can link this comment to https://github.com/workfloworchestrator/orchestrator-core/issues/1921

orchestrator/core/services/processes.py
```diff
@@ -83,7 +82,18 @@
...
+def merge_state(base: State, nxt: State) -> State:
+    """Recursively merge `nxt` into `base`.
+
+    Dict values under a shared key are merged recursively; any other type mismatch or
+    scalar/list value in `nxt` overrides the value in `base`.
+    """
+    return base | {
+        k: merge_state(base[k], v) if isinstance(base.get(k), dict) and isinstance(v, dict) else v
+        for k, v in nxt.items()
+    }
```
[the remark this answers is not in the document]
Mark90: Agree on this, could you spec some unittests against the old StateMerger and see how this replacement function compares to it?

kvklink (approved):

PR #1928, opened 2026-09-21
Guard websocket disconnect() against an already-closed peer

hanstrompert:
Follow-up to #1904, which guarded `MemoryWebsocketManager.remove_ws` against closing an
already-closed socket. The same race remains on `disconnect()`, which the endpoints call directly to
reject a client, bypassing `remove_ws`: `ws.py:68` (invalid authentication), `ws.py:75`,
`processes.py:570` and `settings.py:174` (failed authorization).

If the client is gone by the time the rejection is sent, starlette raises
`RuntimeError: Cannot call "send" once a close message has been sent.` and uvicorn logs an
`Exception in ASGI application` traceback. Verified against 5.4.0:

```
remove_ws(DISCONNECTED):   OK, registry = {}
disconnect(DISCONNECTED):  RAISED RuntimeError: Cannot call "send" once a close message has been sent.
```

### Changes

- `close_websocket()` in `orchestrator/core/websocket/close.py` holds the guard once: returns early
  when `application_state` is DISCONNECTED, suppresses `RuntimeError`/`WebSocketDisconnect` around
  both the reason write and the close.
- Both managers' `disconnect()` route through it. The broadcast manager was unguarded entirely —
  #1904 only touched the memory backend, so Redis-backend deployments got none of that fix.
- The reason write moves inside the guard; previously a `send_text` failure propagated before the
  close was attempted, leaving the socket unclosed.
- `remove_ws()` drops its now-redundant inline guard.

### Tests

Nine parametrized cases across both managers: state guard, close raising mid-race, reason-send
raising before the close, and connected-list cleanup when the close is suppressed. All nine fail on
`main` and pass here.

`1601 passed, 21 skipped` in `test/unit_tests`; ruff and mypy clean.

### Also fixed

`BroadcastWebsocketManager.disconnect_all()` iterated `self.connected` while `disconnect()` removed
from it, so the iterator skipped every other socket and left it registered. It now iterates a
snapshot. Two parametrized cases (2 and 3 clients) assert every socket is closed and the list drains;
both fail without the change. Raised by review; folded in here rather than split out.


test/unit_tests/websocket/test_websocket_manager.py
```diff
@@ -204,6 +204,70 @@ async def test_memory_disconnect(memory_mgr, reason, expect_send: bool):
...
+    """The reason write races the peer too, and fails before the close is even attempted."""
+    mgr = request.getfixturevalue(mgr_name)
+    ws = _make_mock_ws()
+    ws.client, ws.headers = ("127.0.0.1", 1234), {}
+    ws.send_text.side_effect = WebSocketDisconnect(1006)
+    await mgr.disconnect(ws, reason={"error": "denied"})
+
+
+@pytest.mark.asyncio
+async def test_broadcast_disconnect_unregisters_when_close_raises(broadcast_mgr):
```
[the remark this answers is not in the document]
hanstrompert: The tests are not type annotated.

Mark90 (approved):
Lgtm!

orchestrator/core/websocket/managers/broadcast_websocket_manager.py
```diff
@@ -60,13 +61,12 @@ async def connect(self, websocket: WebSocket, channel: str) -> None:
...
     ) -> None:
-        if reason:
-            await websocket.send_text(json_dumps(reason))
-        await websocket.close(code)
+        await close_websocket(websocket, code=code, reason=reason)
         self.remove_ws_from_connected_list(websocket)
 
     async def disconnect_all(self) -> None:
-        for websocket in self.connected:
+        # Snapshot: disconnect() removes from self.connected as we iterate.
```
tjeerddie: i don't mind the comment but for what is the `Snapshot:`

orchestrator/core/websocket/managers/memory_websocket_manager.py
```diff
@@ -86,13 +83,8 @@ async def broadcast_data(self, channels: list[str], data: dict) -> None:
...
 
     async def remove_ws(self, websocket: WebSocket, channel: str) -> None:
-        # Guard on application_state, which is what starlette's send() checks. It goes DISCONNECTED
-        # while client_state stays CONNECTED when a send fails rather than the client disconnecting,
-        # and closing then raises. Suppress as well: the close itself races with the peer going away,
-        # and the registry must be cleaned up either way.
-        if websocket.application_state != WebSocketState.DISCONNECTED:
-            with suppress(RuntimeError, WebSocketDisconnect):
-                await self.disconnect(websocket)
+        # close_websocket never raises, so the registry is cleaned up whether or not the close lands.
```
tjeerddie: unnecessary comment?

tjeerddie (approved):

orchestrator/core/websocket/managers/broadcast_websocket_manager.py
hanstrompert: Removed, along with the comment. disconnect_all() now pops each socket off self.connected before awaiting disconnect(), so the loop no longer walks a list that disconnect() mutates underneath it and there is nothing left to explain. The subsequent remove_ws_from_connected_list() becomes a no-op. Clients now close newest-first; nothing depends on the order. 0c52081c

orchestrator/core/websocket/managers/memory_websocket_manager.py
hanstrompert: Agreed, removed. The comment restated close_websocket()'s contract at the call site, which is where the two would drift apart. Renamed it to close_websocket_safely() and moved "Never raises" into the docstring summary line instead. 286daece

